PR #1, opened 2017-10-03
first pass helper functions

konradjk:


resources.py
```diff
@@ -14,6 +14,45 @@ def public_genomes_vds_path(split=False, version=CURRENT_RELEASE):
     return 'gs://gnomad-public/release/{0}/vds/genomes/gnomad.genomes.r{0}.sites{1}.vds'.format(version, ".split" if split else "")
 
 
+def get_gnomad_data(data_type, hardcalls=None, split=False, hail_version=CURRENT_HAIL_VERSION):
```
lfrancioli: Should we not have these be `get_gnomad_vds` and return a VDS ?
And then add options to keep release samples only, annotate metadata, etc.

Seems like this would add more functionality.

resources.py
```diff
@@ -14,6 +14,45 @@ def public_genomes_vds_path(split=False, version=CURRENT_RELEASE):
...
+            return hardcalls_exomes_vds_path(split, hardcalls == 'adj', hail_version)
+    elif data_type == 'genomes':
+        if not hardcalls:
+            return raw_genomes_vds_path(hail_version)
+        else:
+            return hardcalls_genomes_vds_path(split, hardcalls == 'adj', hail_version)
+    return None
+
+
+def get_gnomad_meta(data_type):
```
lfrancioli: Same comment. This should return a KeyTable I think.

resources.py
konradjk: Agreed, but then, thoughts on keeping this as get_gnomad_data_path() and then having another wrapper to get VDS with others? Sometimes you just want a path

lfrancioli: I feel like getting them directly from resources.py would be enough if just a path is needed. But open for discussion :)

konradjk:
Ok check that out, should be a reasonable first start

PR #3, opened 2017-10-05
* Added release_samples option to get_gnomad_data

lfrancioli:
* Added population and release standard annotations to get_gnomad_meta

### DIFF
--- a/resources.py
+++ b/resources.py
@@ -21,7 +21,8 @@ def public_genomes_vds_path(split=False, version=CURRENT_RELEASE):
 
 
 def get_gnomad_data(hc, data_type, hardcalls=None, split=False, hail_version=CURRENT_HAIL_VERSION,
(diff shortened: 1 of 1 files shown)


resources.py
```diff
@@ -81,6 +86,8 @@ def get_gnomad_meta(hc, data_type, version=None):
         hc
         .import_table(get_gnomad_meta_path(data_type, version), impute=True)
         .key_by("sample" if data_type == "exomes" else "Sample")
+        .annotate(['release = ' + ('drop_status == "keep"' if data_type == "exomes" else 'keep'), #This is version dependent -- will need fixing when new metadata arrives
```
konradjk: Let's put a tag here like: `# unify_sample_qc: this is version dependent...`

konradjk: Also string format not `+`

PR #4, opened 2017-11-02
Refactoring / bug corrections of get_numbered_annotations

lfrancioli:


utils.py
```diff
@@ -352,40 +349,58 @@ def print_attributes(vds, path=None):
...
+    logger.info("Found the following fields:")
+    for k, v in annotations.iteritems():
+        if k is not None:
+            logger.info("{}-based annotations: {}".format(k, ",".join([fields[0] for fields in v])))
+        else:
+            logger.info("Annotations with no number: {}".format(",".join([fields[0] for fields in v])))
+
+    return annotations
+
+def group_annotations_by_attribute(schema, grouping_key , root, recursive = False):
```
konradjk: Want to make `root='va'` by default? Feels like that's going to be the case most of the time.

utils.py
```diff
@@ -352,40 +349,58 @@ def print_attributes(vds, path=None):
...
+    """
+    Groups annotations in a dictionnary by the given attribute key.
+    All annotations that do not have a Number attribute are returned under the key `None`
+
+    :param TStruct schema: Input schema
+    :param str root: Root path to get annotations
+    :param bool recursive: Whether to go recursively to look for Numbered annotations in TStruct fields
```
konradjk: Missing `grouping_key` in docstring. Also, `s/Numbered//` in this one

utils.py
```diff
@@ -352,40 +349,58 @@ def print_attributes(vds, path=None):
...
-                    dot_annotations.append(field)
+            if grouping_key in field.attributes:
+                annotations[field.attributes[grouping_key]].append((path, field))
+        elif recursive and isinstance(field.typ, TStruct):
+            f_annotations = group_annotations_by_attribute(schema, grouping_key, path, recursive)
+            for k,v in f_annotations.iteritems():
+                annotations[k].extend(v)
         else:
-            annotations.append(field)
+            annotations[None].append((path, field))
```
konradjk: Is this gonna work in this case? Previously, we had a `dot_annotations` and an `annotations` to catch things that were arrays of length `.` vs non-arrays.

konradjk: Ohhhhh, I see. So `.` will still be accessed like `annotations['.']` but then `annotations[None]` is the array versions. I think that's fine, but could be documented a bit more.

utils.py
```diff
@@ -352,40 +349,58 @@ def print_attributes(vds, path=None):
...
-            if 'Number' in field.attributes:
-                number = field.attributes['Number']
-                if number == "A":
-                    a_annotations.append(field)
-                elif number == "G":
-                    g_annotations.append(field)
-                else:
-                    dot_annotations.append(field)
+            if grouping_key in field.attributes:
+                annotations[field.attributes[grouping_key]].append((path, field))
```
konradjk: I'm not wild about the tuple of 2 as a value. It's harder to use and you have to remember which is which. Is this unavoidable? (It may be, but just thinking it through)

lfrancioli: The issue here is that the `Field` `name` doesn't have the path. How about using a namedtuple then?

utils.py
lfrancioli: OK! Yeah, I was looking for a key that cannot possibly a String

utils.py
konradjk: Yeah that works

lfrancioli:
@konradjk: Mind taking another quick look. I've taken care of your review comments but ended up using a different strategy regarding handling of the default "dimensions" of the fields in `get_numbered_attributes`. 

I also went ahead and documented a bunch of stuff ;)

utils.py
```diff
@@ -370,18 +463,21 @@ def get_numbered_annotations(schema , root='va', recursive = False):
...
+def group_annotations_by_attribute(schema, grouping_key, root='va', recursive = False, default_func = None):
     """
     Groups annotations in a dictionnary by the given attribute key.
     All annotations that do not have a Number attribute are returned under the key `None`
 
     :param TStruct schema: Input schema
     :param str root: Root path to get annotations
-    :param bool recursive: Whether to go recursively to look for Numbered annotations in TStruct fields
+    :param bool recursive: Whether to go recursively to look for annotations in TStruct fields
+    :param function(Field) default_func: A function that returns the grouping key as a function of the Field. This function is applied to get the grouping key when the grouping key is not foud in the Field attributes.
```
konradjk: s/foud/found/

PR #7, opened 2017-11-21
Added functions to unify VDS schemas

lfrancioli:
* Some updates to `flatten_struct`: recursive option + OrderedDict
* Added public PC loadings VDS to resources

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -7,8 +7,9 @@ import os
 
 from resources import *
 from hail import *
+from hail.expr import Field
 from slack_utils import *
-from collections import defaultdict, namedtuple
+from collections import defaultdict, namedtuple, OrderedDict
 from pprint import pprint, pformat
 
 logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
@@ -160,9 +161,10 @@ def filter_star(vds, a_based=None, r_based=None, g_based=None, additional_annota
     return vds.filter_alleles('v.altAlleles[aIndex - 1].alt == "*"', annotation=annotation, keep=False)
 
 
-def flatten_struct(struct, root='va', leaf_only=True):
+def flatten_struct(struct, root='va', leaf_only=True, recursive=True):
     """
-    Given a TStruct and its root path, creates a dict of each path -> Field by flattening the TStruct tree.
+    Given a TStruct and its root path, creates an OrderedDict of each path -> Field by flattening the TStruct tree.
+    The order of the fields is the same as the input `struct` fields.
     The following TStruct at root 'va', for example
     Struct{
      rsid: String,
@@ -190,16 +192,17 @@ def flatten_struct(struct, root='va', leaf_only=True):
     :param TStruct struct: The struct to flatten
     :param str root: The root path of the struct to flatten (added at the beginning of all dict keys)
     :param bool leaf_only: When set to true, only leaf nodes in the tree are output in the output
+    :param bool recursive: When set to false, returns a dict of field_name: field_type
     :return: Dictionary of path : Field
-    :rtype: dict of str:Field
+    :rtype: OrderedDict of str:Field
     """
-    result = {}
+    result = OrderedDict()
     for f in struct.fields:
-        path = '{}.{}'.format(root, f.name)
-        if isinstance(f.typ, TStruct):
-            result.update(flatten_struct(f.typ, path, leaf_only))
+        path = '{}.{}'.format(root, f.name) if root else f.name
+        if isinstance(f.typ, TStruct) and recursive:
             if not leaf_only:
                 result[path] = f
+            result.update(flatten_struct(f.typ, path, leaf_only))
         else:
             result[path] = f
     return result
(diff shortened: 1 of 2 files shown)


utils.py
```diff
@@ -926,4 +929,103 @@ def quote_field_name(f):
...
+                            attributes[k] = v
+
+                if isinstance(f.typ, TStruct) and f_overlap:
+                    fields[name] = Field(name, merge_TStructs([f.typ] + [f2.typ for f2 in f_overlap]))
+                else:
+                    fields[name] = f
+
+                fields[name].attributes = attributes
+
+    return TStruct.from_fields(fields.values())
```
konradjk: Is the `OrderedDict` just to avoid having to check the name of everything in fields? Sounds good if so, just curious.

utils.py
```diff
@@ -926,4 +929,103 @@ def quote_field_name(f):
...
 
+def merge_TStructs(s):
+    """
+
+    Merges multiple TStructs together and outputs a new TStruct with the union of the fields.
+    Notes:
+    - If a leaf field is found in multiple input TStructs, its type has to be the same
+    - Attributes are merged. If the same attribute is found in different TStructs it should have the same value.
+
+    :param list s: List of Structs to merge
```
konradjk: `:param list of TStruct s: List of Structs to merge`

utils.py
```diff
@@ -926,4 +929,103 @@ def quote_field_name(f):
...
+        for f in struct.fields:
+            path = '{}.{}'.format(root, f.name)
+            if not path in old_schema_fields.keys():
+                field_expr.append('{}: NA:{}'.format(f.name, f.typ))
+            elif isinstance(f.typ, TStruct):
+                field_expr.append('{}: {}'.format(f.name, get_schema_expr(f.typ, path, old_schema_fields)))
+            else:
+                field_expr.append('{0}: {1}'.format(f.name, path))
+
+        print '{{{}}}'.format(",".join(field_expr))
```
konradjk: Meant to leave this in as a print? `logger.debug`

utils.py
```diff
@@ -926,4 +929,103 @@ def quote_field_name(f):
...
+                fields[name].attributes = attributes
+
+    return TStruct.from_fields(fields.values())
+
+
+def replace_vds_variant_schema(vds, new_schema):
+    """
+
+    Replaces the input VDS va with the new schema. Values for all fields present in the old variant schema
+    that have the same type are kept. All other fields are filled with `NA`
```
konradjk: What if they're not the same type?

utils.py
```diff
@@ -926,4 +929,103 @@ def quote_field_name(f):
...
+                field_expr.append('{}: {}'.format(f.name, get_schema_expr(f.typ, path, old_schema_fields)))
+            else:
+                field_expr.append('{0}: {1}'.format(f.name, path))
+
+        print '{{{}}}'.format(",".join(field_expr))
+        return '{{{}}}'.format(",".join(field_expr))
+
+    for path, field in flatten_struct(new_schema, root='va').iteritems():
+        if field.attributes:
+            vds = vds.set_va_attributes(path, field.attributes)
```
konradjk: I thought rewriting a schema overwrites the attributes. If that's the case, shouldn't this come after the `annotate_variants_expr`? I could be wrong

utils.py
```diff
@@ -926,4 +929,103 @@ def quote_field_name(f):
...
+        field_expr = []
+
+        for f in struct.fields:
+            path = '{}.{}'.format(root, f.name)
+            if not path in old_schema_fields.keys():
+                field_expr.append('{}: NA:{}'.format(f.name, f.typ))
+            elif isinstance(f.typ, TStruct):
+                field_expr.append('{}: {}'.format(f.name, get_schema_expr(f.typ, path, old_schema_fields)))
+            else:
+                field_expr.append('{0}: {1}'.format(f.name, path))
```
konradjk: Why the switch to `{0}` and `{1}`?

utils.py
```diff
@@ -926,4 +929,103 @@ def quote_field_name(f):
...
+    Replaces the input VDS va with the new schema. Values for all fields present in the old variant schema
+    that have the same type are kept. All other fields are filled with `NA`
+
+    :param VariantDataset vds: input VDS
+    :param TStruct new_schema: new schema
+    :return: VDS with new schema
+    :rtype: VariantDataset
+    """
+
+    def get_schema_expr(struct, root, old_schema_fields):
```
konradjk: Is this worth a brief docstring even if it's not used too much? I'm good either way, just a suggestion. Especially since I have this feeling that this function might get edited for hail 0.2

utils.py
lfrancioli: Yeah, it's for O(1) lookups -- I could use a list too as it really wouldn't make a big difference.

utils.py
```diff
@@ -926,4 +929,103 @@ def quote_field_name(f):
     return '`{}`'.format(f) if re.search('^\d|\.', f) else f
 
 
+def merge_TStructs(s):
```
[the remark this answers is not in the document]
lfrancioli: Agreed. I also think that these functions will eventually go to contrib. Not sure if I should write tests now or when it goes there...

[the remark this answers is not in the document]
konradjk: You know how this works - if you don't write them now, it'll be harder to write them later 😄 - but I leave it up to you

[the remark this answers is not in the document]
lfrancioli: I know...just making excuses I guess ;)
I'll try and do that Monday..

konradjk:
Ok, given the interface changes that are coming, and some of this is going to have to change anyway, I'm just gonna merge this in.

PR #14, opened 2018-02-11
Functions for QC scripts

konradjk:
Stacks onto #13

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -6,8 +6,9 @@ import gzip
 import os
 
 from resources import *
-from hail import *
+from hail2 import *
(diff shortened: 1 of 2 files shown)


utils.py
```diff
@@ -897,6 +1012,349 @@ def quote_field_name(f):
 
     return '`{}`'.format(f) if re.search('^\d|\.', f) else f
 
+# Bootleg:
```
lfrancioli: well, I am guessing below will go :)

konradjk: Yep, already gone in `hail-0.2`

PR #22, opened 2018-03-12
pc_project first pass

konradjk:
So I don't love this at the moment, since it requires the original VDS (just to calculate the allele frequencies). We could hack our own PCA so that it adds AF into the loadings table, but then you have to use our PCA function in order to use this. Thoughts?

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -457,32 +457,24 @@ def filter_annotations_regex(annotation_fields, ignore_list):
     return [x for x in annotation_fields if not ann_in(x.name, ignore_list)]
 
 
-def pc_project(mt, pc_mt, pca_loadings_root='va.pca_loadings'):
+def pc_project(mt: hl.MatrixTable, pc_mt: hl.MatrixTable, pc_loadings: hl.Table):
     """
     Projects samples in `mt` on PCs computed in `pc_mt`
-    :param mt: MT containing the samples to project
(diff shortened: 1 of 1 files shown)


utils.py
```diff
@@ -457,32 +457,24 @@ def ann_in(name, lst):
...
+def pc_project(mt: hl.MatrixTable, pc_loadings: hl.Table):
     """
     Projects samples in `mt` on PCs computed in `pc_mt`
-    :param mt: MT containing the samples to project
-    :param pc_mt: MT containing the PC loadings for the variants
-    :param pca_loadings_root: Annotation root for the loadings. Can be either an Array[Double] or a Struct{ PC1: Double, PC2: Double, ...}
-    :return: MT with
-
-    pc_mt = pc_mt.annotate_variants_expr('va.pca.calldata = gs.callStats(g => v)')
+    :param MatrixTable mt: MT containing the samples to project (assumes data in `loadings` and `pca_af`)
```
lfrancioli: Is that paren meant for next line?

Also, I usually pass these as `str` params to get the little extra flexibility. But I suppose both ways are ok.

konradjk: Yes, it is. Will fix. And yeah, I thought about passing as strings, but it's nuking annotations in `mt` so its sorta dangerous to let them be called anything. But I guess that's the user's problem.

PR #29, opened 2018-03-21
New input meta resources

gtiao:
Have a look and see if this arrangement of files makes sense


resources/basics.py
```diff
@@ -337,5 +337,26 @@ def methylation_sites_mt_path(hail_version=CURRENT_HAIL_VERSION):
 context_mt_path = 'gs://gnomad-resources/constraint/context_processed.mt'
 
 
+# Sample QC files
```
konradjk: Looks like this got merged back in. You can remove these, as they're already in resources.sample_qc

PR #30, opened 2018-03-22
Changes corresponding to sample qc

konradjk:


resources/sample_qc.py
```diff
@@ -1,12 +1,20 @@
 
 
 # Sample QC files
-def qc_mt_path(data_type: str):
-    return 'gs://gnomad/sample_qc/mt/gnomad.{}.high_callrate_common_biallelic_snps.mt'.format(data_type)
+def qc_mt_path(data_type: str, pruned: bool = False):
+    # can be joint, pruned only for joint
+    return 'gs://gnomad/sample_qc/mt/gnomad.{}.high_callrate_common_biallelic_snps{}.mt'.format(data_type,
```
gtiao: Let's put in `{0}` and `{1}` to make the assignments absolutely clear

resources/sample_qc.py
```diff
@@ -1,12 +1,20 @@
...
+                                                                                                '.pruned' if pruned else '')
 
 
-def qc_ht_path(data_type: str):
-    return 'gs://gnomad/sample_qc/ht/gnomad.{}.high_callrate_common_biallelic_snps.ht'.format(data_type)
+def qc_ht_path(data_type: str, part: str):
+    # part can be hard_filters, platforms, or pop_platform
+    return 'gs://gnomad/sample_qc/ht/gnomad.{}.{}.ht'.format(data_type, part)
```
gtiao: See comment above

resources/basics.py
```diff
@@ -269,6 +269,14 @@ def metadata_exomes_tsv_path(version=CURRENT_EXOME_META):
     return 'gs://gnomad/metadata/exomes/gnomad.exomes.metadata.{0}.tsv.bgz'.format(version)
 
 
+def metadata_genomes_ht_path(version=CURRENT_GENOME_META):
```
gtiao: Given that we now have streamlined metadata available, e.g., `gs://gnomad/sample_qc/input_meta/gnomad.genomes.streamlined_metadata.2018-03-21.txt.bgz`, is there still a need to have the old metadata floating around? Can we remove these?

resources/sample_qc.py
konradjk: Really? I don't love those when its not necessary, but I'm happy to be voted down. TBH my favorite is:
```
return 'gs://gnomad/sample_qc/mt/gnomad.{data_type}.high_callrate_common_biallelic_snps{part}.mt'.format(data_type=data_type, part=part)
```
or the new python3 formatting:
```
return f'gs://gnomad/sample_qc/mt/gnomad.{data_type}.high_callrate_common_biallelic_snps{part}.mt'
```

gtiao: I don't particularly mind which format we pick (the other two options look good to me), but I like having the assignments clearer than just scanning across the line to count which thing is what

resources/basics.py
gtiao: That's fine, I guess I just meant the data files that are currently there (2017-06-02). Some of them are already hidden in the `sample_qc/input_meta/source` directory so it's weird that they are also "out there" to be seen.

konradjk: Yeah, I mean, that's fine. I'd probably keep the ones that are in there - no harm in having a few, and then the CURRENT_META_VERSION points to the newer ones.

PR #38, opened 2018-05-04
Necessary fixes for annotation pipelines

konradjk:
Note: should wait until https://github.com/hail-is/hail/pull/3505 goes in so that we can use `hl.min_rep` as a function inside split_multi_dynamic


konradjk:
As far as I know, it's only the `annotations/` scripts, which I did update

PR #41, opened 2018-05-18
Changes for variant QC / RF

lfrancioli:
* Moved RF to utils
* Added RF run hash path
* Updated rf_parh logic

### DIFF
--- a/utils/rf.py
+++ b/utils/rf.py
@@ -10,7 +10,7 @@ import pandas as pd
 
 def run_rf_test(
         mt: hl.MatrixTable,
-        output: str = '/Users/laurent/tmp'
+        output: str = '/tmp'
 ) -> Tuple[pyspark.ml.PipelineModel, hl.MatrixTable]:
     """
     Runs a dummy test RF on a given MT:
@@ -48,7 +48,7 @@ def run_rf_test(
     quantiles = get_columns_quantiles(ht, features_to_impute, [0.5])
     quantiles = {k: v[0] for k, v in quantiles.items()}
 
-    print(quantiles)
+    logger.info('Features median:\n{}'.format(f'{k}: {v}\n' for k,v in quantiles.items()))
     ht = ht.annotate(
         **{f: hl.or_else(ht[f], quantiles[f]) for f in features_to_impute}
     )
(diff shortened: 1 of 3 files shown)


utils/rf.py
```diff
@@ -0,0 +1,368 @@
...
+from pyspark.ml.classification import *
+from pyspark.ml import *
+from pyspark.sql import Row
+import hail as hl
+import pandas as pd
+
+
+def run_rf_test(
+        mt: hl.MatrixTable,
+        output: str = '/Users/laurent/tmp'
```
konradjk: Maybe just default to `/tmp`?

utils/rf.py
```diff
@@ -0,0 +1,368 @@
...
+    ht = ht.annotate(
+        **{f: hl.or_else(ht[f], quantiles[f]) for f in features_to_impute}
+    )
+    ht = ht.annotate_globals(medians= quantiles)
+
+    f3_after_imputation = f3stats(ht)
+    logger.info('Feature3 defined values after imputation: {}'.format(f3_after_imputation.n))
+    logger.info('Feature3 median: {}'.format(f3_after_imputation.med))
+
+    ht = ht.select('locus', 'alleles', 'label', 'feature1', 'feature2', 'feature3')
```
konradjk: I believe you can't `select` keys anymore, so remove locus and alleles

utils/rf.py
```diff
@@ -0,0 +1,368 @@
...
+
+    f3_before_imputation = f3stats(ht)
+    logger.info('Feature3 defined values before imputation: {}'.format(f3_before_imputation.n))
+    logger.info('Feature3 median: {}'.format(f3_before_imputation.med))
+
+    features_to_impute = ['feature3']
+    quantiles = get_columns_quantiles(ht, features_to_impute, [0.5])
+    quantiles = {k: v[0] for k, v in quantiles.items()}
+
+    print(quantiles)
```
konradjk: pprint?

utils/rf.py
```diff
@@ -0,0 +1,368 @@
...
+
+    ht = ht.annotate(
+        **{
+            probability_col_name: {label: rf_ht[ht['idx']]["probability"][i] for i, label in enumerate(get_labels(rf_model))},
+            prediction_col_name: rf_ht[ht['idx']]["predictedLabel"]
+        }
+    )
+
+    if not has_idx:
+        ht = ht.drop('idx')
```
konradjk: Can't drop a key. I think you now would need to `ht = ht.key_by().drop('idx')`

lfrancioli: Decided to always index rather than rely on previously annotated `idx` column if any. Seems more robust and flow makes more sense now I think.

PR #45, opened 2018-05-24
Support for minrep in split_multi_dynamic

konradjk:


utils/generic.py
```diff
@@ -69,16 +69,17 @@ def split_multi_dynamic(t: Union[hl.MatrixTable, hl.Table], keep_star: bool = Fa
         t = t.annotate(a_index=hl.range(1, hl.len(t.alleles)), was_split=hl.len(t.alleles) > 2)
         t = t.explode('a_index')
 
+        new_locus_alleles = hl.min_rep(t.locus, hl.array([t.alleles[0], t.alleles[t.a_index]]))
         if 'alleles' in t.key:
```
lfrancioli: This won't work properly for a HT keyed by `locus` only.

lfrancioli:
We could require the HT to be keyed by locus, alleles. Otherwise, flow needs a little

PR #48, opened 2018-05-29
Enhancement to generic for sample QC

lfrancioli:
* Moved `get_duplicated_samples` and `infer_families` from `gnomad_functions.py`
* Moved `assign_population_pcs` from `gnomad_hail`. 
* Adapted `pc_project` to flow naturally with output from `pc_project`
* Corrrect bug in `pc_project` (variant count was done before filtering)
* Adapted `assign_population_pcs` interface to make it more generic and flow naturally with output from `pc_relate` / `pc_project`
* Added `pair_plot` to plotting


utils/generic.py
```diff
@@ -145,24 +148,31 @@ def split_multi_dynamic(t: Union[hl.MatrixTable, hl.Table], keep_star: bool = Fa
...
     """
-    n_variants = mt.count_rows()
-
-    mt = mt.annotate_rows(**pc_loadings[mt.locus, mt.alleles])
     mt = mt.filter_rows(hl.is_defined(mt[loading_location]) & hl.is_defined(mt[af_location]) &
                         (mt[af_location] > 0) & (mt[af_location] < 1))
 
+    mt = mt.persist()
+
+    n_variants = mt.count_rows()
```
konradjk: Are you sure this is the correct order? I can understand this after filtering for `hl.is_defined(mt[loading_location]) & hl.is_defined(mt[af_location])`, but if your new dataset is monomorphic at any of the sites, that site should probably be counted, no? This count should be the same as the number of variants used for PCA, and should not be specific to the projected dataset, I'd think? I could be wrong.

lfrancioli: I think you're right. so in fact, before should have been pc_loadings.count_rows() and not mt.count_rows(). But thinking about it, it makes sense indeed. So now I'm less thrilled about not passing the loadings table any more, I'll re-work that part.

PR #52, opened 2018-05-31
Phase by transmission

lfrancioli:


utils/generic.py
```diff
@@ -308,3 +308,184 @@ def annotation_type_in_vcf_info(t: Any) -> bool:
...
+            :param int n_alleles: number of alleles at the site
+            :return:
+            """
+            return hl.cond(
+                gt.is_het(),
+                hl.array([
+                    allele_to_vector(gt, 0, n_alleles),
+                    allele_to_vector(gt, 1, n_alleles)
+                ]),
+                hl.array([allele_to_vector(gt, 0, n_alleles)])
```
konradjk: Make sure this does what you want in the case of missingness (I think so, but just checking).

utils/generic.py
```diff
@@ -308,3 +308,184 @@ def annotation_type_in_vcf_info(t: Any) -> bool:
...
+            )
+
+        def allele_to_vector(gt: hl.expr.CallExpression, allele_index: int, n_alleles: int) -> hl.expr.ArrayExpression:
```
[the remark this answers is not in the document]
konradjk: Wait, I guess it kind of is. But then I'm confused as to the role of `allele_index`

utils/generic.py
```diff
@@ -308,3 +308,184 @@ def annotation_type_in_vcf_info(t: Any) -> bool:
...
+            Given a genotype and which allele was transmitted to the offspring, returns the parent phased genotype.
+
+            :param CallExpression gt: Parent genotype
+            :param int transmitted_allele_index: index of transmitted allele (0 or 1)
+            :return: Phased parent genotype
+            :rtype: CallExpression
+            """
+            return hl.call(
+                gt[transmitted_allele_index],
+                hl.cond(
```
konradjk: I think you can probably `gt[hl.int(transmitted_allele_index == 0)]`

utils/generic.py
```diff
@@ -308,3 +308,184 @@ def annotation_type_in_vcf_info(t: Any) -> bool:
...
+            proband_call,
+            father_call,
+            mother_call,
+            hl.len(tm.alleles)
+        )
+    )
+
+    tm = tm.select_entries(
+        proband_entry=hl.struct(
+            **tm.proband_entry,
```
konradjk: Not really sure if there's a point in making `proband_call` etc general, if you're assuming a `proband_entry` field exists here anyway (at that point, why not just take a single TrioMatrix, with those assumptions asserted).

utils/generic.py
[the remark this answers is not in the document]
konradjk: Ok after a bunch more reading, I understand a bit better, but might be good to document why you're doing this (here, or a short comment at the `combinations` line)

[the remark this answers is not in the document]
lfrancioli: Yeah, so I could also use something like
```
hl.call(gt[0]).one_hot_alleles(alleles)
hl.call(gt[1]).one_hot_alleles(alleles)
```

Maybe clearer? Otherwise, I'll rename this `one_hot_allele` to make it clearer

utils/generic.py
lfrancioli: good call!

utils/generic.py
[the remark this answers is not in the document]
konradjk: Ooo yeah, if it's equivalent, I think that is better

utils/generic.py
lfrancioli: Fair point. The goal was more to be able to choose which genotype field you'd want to use. But I'm now leaning towards having `trio_matrix: hl.MatrixTable` and `genotype_field: str` as args to specify this.

konradjk:
Your call. Doesn't really seem necessary, but could be useful for convenience.

konradjk:
Thoughts on some tests?

lfrancioli:
Would be good, no time at all. If I remove that outer function though might be rather easy...so maybe I'll do that!

PR #55, opened 2018-06-05
Fix label

gtiao:


konradjk:
Fix looks good, I think the paths are zombie code though

PR #65, opened 2018-06-12
Frequency filtering function

konradjk:


utils/gnomad_functions.py
```diff
@@ -133,6 +133,77 @@ def read_list_data(input_file: str) -> List[str]:
     return output
 
 
+def filter_by_frequency(t: Union[hl.MatrixTable, hl.Table], frequency: float = None, allele_count: int = None,
+                        population: str = None, subpop: str = None, downsampling: int = None,
```
lfrancioli: should direction default really be `equal` ?

konradjk: what would you prefer?

PR #71, opened 2018-07-23
Add rank fix

lfrancioli:


lfrancioli:
@gtiao : Because `key_by` always sorts in ascending order, the score has to be sorted in ascending order. Not sure where you create the VQSR rank, but double-check that it does the right thing there.

lfrancioli:
Changed the implementation -- can anyone  take another look?

utils/gnomad_functions.py
```diff
@@ -359,54 +360,39 @@ def pretty_print_runs(runs: Dict, label_col: str = 'rf_label', prediction_col_na
             res_pd = res_pd.pivot(index=label_col, columns=prediction_col_name, values='n')
             logger.info("Testing results:\n{}".format(pformat(res_pd)))
 
-            
-def add_full_rankings(ht: hl.Table, score_field: hl.expr.NumericExpression) -> hl.Table:
+
+def add_rank(ht: hl.Table,
+             score_expr: hl.expr.NumericExpression,
+             subrank_expr: Dict[str, hl.expr.BooleanExpression] = None) -> hl.Table:
```
konradjk: Make `Optional[Dict...]`

konradjk:
Actually I'm gonna make a whitespace change. Don't merge just yet, but otherwise it's ready I think

konradjk:
Ok, take a quick look. Revert that last commit if you hate it, but if not, go ahead and merge

PR #83, opened 2018-11-30
pop color and names

lfrancioli:


utils/plotting.py
```diff
@@ -129,32 +129,32 @@ def new_show(t, n=10, width=140, truncate=40, types=True):
...
              'sas_non_consang': 'South Asian (F < 0.05)',
              'consanguineous': 'South Asian (F > 0.05)',
              'exac': 'ExAC',
-             'bg': 'Bulgaria (Eastern Europe)',
-             'de': 'Germany',
-             'ee': 'Estonia',
-             'es': 'Spain',
-             'fi': 'Finland',
-             'gb': 'UK',
+             'bgr': 'Bulgarian',
```
konradjk: We had semi-intentionally had the parenthetical here (since we're not necessarily convinced this is a strictly-Bulgarian signature, but might be picking up any Eastern European)

lfrancioli: So want it back?

lfrancioli:
Sorry for the mess with the plot stuff -- I  have removed that part.

utils/plotting.py
konradjk: Yes I think so

PR #88, opened 2019-02-07
Fallback to earlier gnomAD release if a resource file is not found fo…

lfrancioli:
…r desired version

- Fixed public PCA loadings path and type

### DIFF
--- a/resources/basics.py
+++ b/resources/basics.py
@@ -1,5 +1,6 @@
 import hail as hl
 from typing import *
+from utils.generic import check_hail_file
 
 CURRENT_HAIL_VERSION = "0.2"
 CURRENT_RELEASE = "2.1.1"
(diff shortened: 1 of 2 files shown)


resources/basics.py
```diff
@@ -54,18 +54,46 @@
...
+def get_release_file(file_path: str, version: str = CURRENT_RELEASE) -> str:
+    """
+    Tries to get the desired file from the corresponding release version on the google cloud.
+    If the file is not found for the desired release version, falls back on previous versions.
+
+    :param  str file_path: Desired file path, with {0} as placeholder(s) for the version number
+    :param str version: Desired file version
+    :return: Path for closest version of the file available
+    """
+    if hl.hadoop_exists(file_path.format(version)):
```
konradjk: Should we be checking for the _SUCCESS file?

resources/basics.py
```diff
@@ -54,18 +54,46 @@
...
+        return file_path.format(version)
+    else:
+        for v in range(RELEASES.index(version)-1, -1, -1):
```
konradjk: `for v in range(RELEASES.index(version) - 1, -1, -1)`

resources/basics.py
```diff
@@ -54,18 +54,46 @@
...
+            if hl.hadoop_exists(file_path.format(RELEASES[v])):
+                print("WARN: Resource {} could not be found for gnomAD release version {}.\n Loading gnomAD release version {} of the file. ({})".format(
```
konradjk: Might look nicer with an f-string, but that's up to you

resources/basics.py
konradjk: I wrote this function the other day to do just this, perhaps this could be useful here:

```
def check_hail_file(fname):
    return hl.hadoop_exists(f'{fname}/_SUCCESS')
```

resources/basics.py
lfrancioli: Because of the nested `format` statements I feel like `format` is a little more readable in this case...

konradjk: You could still nest them:
```
print("WARN: Resource {file_path.format(version)} could not be found for gnomAD release version {version}.\n Loading gnomAD release version {RELEASES[v]} of the file. ({file_path.format(RELEASES[v])})")
```
but if you disagree, that's fine.

lfrancioli:
Yeah, in this instance I somehow prefer `format` ... but it's personal obviously :)

PR #89, opened 2019-02-21
Added joint_plot

lfrancioli:
Should go into hail.plot at some point I guess, but not time for this now.


konradjk:
You could put it in hl.experimental.plot, where the bar is a bit lower

lfrancioli:
PR'ed to hail

konradjk:
👋

PR #90, opened 2019-03-14
simplify filter low conf function

konradjk:


utils/generic.py
```diff
@@ -127,24 +128,38 @@ def filter_low_conf_regions(mt: hl.MatrixTable, filter_lcr: bool = True, filter_
     :return: MT with low confidence regions removed
     :rtype: MatrixTable
     """
-    from gnomad_hail.resources import lcr_intervals_path, decoy_intervals_path, segdup_intervals_path
+    from gnomad_hail.resources import lcr_intervals_path, decoy_intervals_path, segdup_intervals_path, high_coverage_intervals_path
 
+    criteria = [True]
```
lfrancioli: not that it _really_ matters, but I'd rather have this be `[]` and then not trigger any filtering action if it stays empty

utils/generic.py
```diff
@@ -127,24 +128,38 @@ def filter_low_conf_regions(mt: hl.MatrixTable, filter_lcr: bool = True, filter_
...
         for region in high_conf_regions:
             region = hl.import_locus_intervals(region)
-            mt = mt.filter_rows(hl.is_defined(region[mt.locus]), keep=True)
+            criteria.append(hl.is_defined(region[mt.locus]))
+
+    if filter_exome_low_coverage_regions:
+        high_cov = hl.import_locus_intervals(high_coverage_intervals_path)
+        criteria.append(hl.is_missing(high_cov[mt.locus]))
+
+    import operator
```
lfrancioli: Move imports to top of function

utils/generic.py
```diff
@@ -127,24 +128,38 @@ def filter_low_conf_regions(mt: hl.MatrixTable, filter_lcr: bool = True, filter_
     :return: MT with low confidence regions removed
     :rtype: MatrixTable
```
lfrancioli: Update docstring

PR #93, opened 2019-03-21
adding select primitives function

konradjk:


konradjk:
All the time! This will remove structs (which you can `.flatten()` of course if you want, but sometimes you don't need to, or you can run this after a round of flatten to remove any more structs) and, more importantly, arrays (like the `ht.freq` array) - this will just remove those to cut down on file size on export.

gtiao:
Okeydoke, then this looks good to me

PR #94, opened 2019-04-03
Initial commit

ch-kr:
Initial commit of liftover code. Sorry it's a bit rough!


resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+from gnomad_hail import *
+import logging
+
+
+logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
+logger = logging.getLogger("gnomAD_liftover")
+logger.setLevel(logging.INFO)
+
+
+def checkpoint(ht: hl.Table, dtype: str) -> hl.Table:
```
konradjk: This function now exists in hail! `ht = ht.checkpoint(path)` does exactly this

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+    return mismatch
+
+
+def main(args):
+
+    hl.init()
+    
+    # get reference genomes and load liftover chain file
+    rg37 = hl.get_reference('GRCh37')
+    rg38 = rg38 = hl.get_reference('GRCh38')
```
konradjk: Double `rg38 =`

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+        ht = hl.read_table('gs://seqr-datasets/methods_dev/test_data/kc/gnomAD_genome_sites.ht')
+
+    return ht
+
+
+def lift_ht(ht: hl.Table, dtype: str, rg: hl.genetics.ReferenceGenome) -> hl.Table:
+    """
+    Lifts gnomAD release table from b37 to b38
+    :param Table ht: gnomAD release Table
+    :param str dtype: Data type (exome or genome)
```
konradjk: Probably worth spelling out `data_type` to be consistent with the rest of the codebase.

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    Creates a "checkpoint" by writing Table out and reading back in
+    :param Table ht: Table to be evaluated and saved
+    :param str dtype: Data type (exomes or genomes)
+    :return: Table after writing out and reading back in
+    :rtype: Table
+    """
+
+    if dtype == 'exome':
+        ht.write('gs://seqr-datasets/methods_dev/test_data/kc/gnomAD_exome_sites.ht', overwrite=True)
```
konradjk: Should probably move these paths to the top of the script. You could do something like:
```
intermediate_liftover_path = 'gs://seqr-datasets/methods_dev/test_data/kc/gnomAD_{}_sites.ht'
```
And then when you go to use it:
```
intermediate_liftover_path.format(data_type)
```

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+    ht = ht.annotate(new_locus = hl.liftover(ht.locus, rg, include_strand=True),
+                     old_locus=ht.locus
+                    )
+    ht = ht.key_by(locus = ht.new_locus.result)
+    ht = ht.annotate(no_target = hl.is_defined(ht.locus))
+    ht = checkpoint(ht, dtype)
+    return ht
+
+
+def reverse_complement(allele: str):
```
konradjk: This should just move somewhere into `utils`

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+    # get reference genomes and load liftover chain file
+    #rg37 = hl.get_reference('GRCh37')
+    #rg38 = rg38 = hl.get_reference('GRCh38')
+    #rg37.add_liftover('gs://hail-common/references/grch37_to_grch38.over.chain.gz', rg38)
+
+    # annotate release file with liftover coordinates
+    ht = ht.annotate(new_locus = hl.liftover(ht.locus, rg, include_strand=True),
+                     old_locus=ht.locus
+                    )
+    ht = ht.key_by(locus = ht.new_locus.result)
```
konradjk: Probably want to keep `alleles` in the key

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+    ht = ht.annotate(no_target = hl.is_defined(ht.locus))
```
konradjk: This doesn't seem like it's used anywhere down the line?

resources/gnomad_liftover.py
konradjk: Also, probably keep the name as `flip_base` (or something like `complement`) as there's no actual reversing going on here.

resources/gnomad_liftover.py
konradjk: And I guess on that note, maybe it's worth doing the reverse complementing here? But maybe that comes after we assess the results.

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+    Checks for mismatches between reference allele and allele in reference fasta
+    :param Table ht: Table to be checked
+    :return: Dictionary containing mismatch type ('total' or 'minus' as key) and umber of mismatches as value
+    :rtype: dict
+    """
+
+    mismatch = {}
+    ht = ht.filter(ht.reference_mismatch)
+    mismatch['total'] = ht.count()
+    mismatch['minus'] =  ht.filter(ht.new_locus.is_negative_strand).count()
```
konradjk: If you're finding this takes a while, it might be easier to put these queries (and more - you could also simultaneously count the ones that are not mismatched by removing the filter above) into one big aggregate. Let me know if that makes sense or I can suggest further.

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+    # add fasta sequence to rg38
+    rg38.add_sequence(
+            'gs://hail-common/references/Homo_sapiens_assembly38.fasta.gz', 
+            'gs://hail-common/references/Homo_sapiens_assembly38.fasta.fai'
+            )
+    
+    logger.info('Reading in release ht')
+    if args.type == 'exome':
+        dtype = 'exome'
+        ht = hl.read_table(public_exomes_ht_path(split=True, version=CURRENT_RELEASE))
```
konradjk: Fine to keep this here if you want to keep it modular for versions but this will do the same as:
```
ht = get_gnomad_public_data(dtype)
```
Note that dtype would need to be plural in this scheme - you may want to switch to that to be consistent with the rest of our scripts, but that's up to you.

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,160 @@
...
+    mismatch = check_mismatch(ht)
+    logger.info('{} reference mismatches in SNPs'.format(mismatch['total']))
+    logger.info('{} mismatches on minus strand'.format(mismatch['minus']))
+
+
+if __name__ == '__main__':
+
+    # Create argument parser
+    parser = argparse.ArgumentParser(description='This script lifts gnomad ht to b38')
+    parser.add_argument('-t', '--type', help='Data type', choices=['exome', 'genome'], required=True)
```
konradjk: As above, it's up to you if you want to keep it this way, but take a look at how we do the `--exomes` or `--genomes` flags in some of the other scripts (e.g. `generate_frequency_data.py`).

resources/gnomad_liftover.py
ch-kr: that's awesome, I had no idea! will remove this function

konradjk: It's brand new (as of 0.2.11 I think). I also added a hidden option not to rerun stuff above if the file already exists `ht = ht.checkpoint(path, _read_if_exists=True)` (note that if you change the code it won't regenerate the file, but it's good for fast iteration on downstream things)

resources/gnomad_liftover.py
ch-kr: you're right, I think I added that when I was testing to see how many variants on chr21 failed liftover because there was no liftover locus. will remove

resources/gnomad_liftover.py
ch-kr: I assumed it was best not to do too many things after liftover and before creating the checkpoint but can flip the alleles here - what do you think?

resources/gnomad_liftover.py
```diff
@@ -128,33 +98,39 @@ def main(args):
...
+    ht = get_gnomad_public_data(data_type, split=True, version=CURRENT_RELEASE)
+    
     #logger.info('Filtering to chr21 for testing')
     #ht = ht.filter(ht.locus.contig == '21')
 
     logger.info('Lifting ht to b38')
-    ht = lift_ht(ht, dtype, rg38)
+    ht = lift_ht(ht, data_type, rg38)
 
+    ht = hl.read_table(f'gs://seqr-datasets/methods_dev/test_data/kc/gnomAD_{data_type}_sites.ht')
```
konradjk: Probably worth putting in a gnomad-resources directory

resources/gnomad_liftover.py
```diff
@@ -0,0 +1,136 @@
...
+    ht = ht.filter(hl.is_snp(ht.alleles[0], ht.alleles[1]))
+    snp_count = ht.count()
+    logger.info('{} SNPs in table'.format(snp_count))
+
+    # check if reference allele matches what is in reference fasta
+    # for snps on negative strand, make sure reverse complement of ref allele matches what is in reference fasta
+    ht = ht.annotate(
+        reference_mismatch = hl.cond(
+                                ht.new_locus.is_negative_strand,
+                                (generic.flip_base(ht.alleles[0]) != hl.get_sequence(ht.locus.contig, ht.locus.position, reference_genome=rg)),
```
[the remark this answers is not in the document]
ch-kr: here locus is already the lifted over locus;  new_locus is a struct: 
```'new_locus': struct {
        result: locus<GRCh38>,
        is_negative_strand: bool
    }```
would it be clearer if I used new_locus.result.contig instead of locus?

File: resources/basics.py [hunk 13]
```diff
@@ -496,6 +496,39 @@ def qc_meta_path(data_type: str):
...
+    :return: Path to chosen Table
+    :rtype: str
+    """
+    if version not in RELEASES:
+        return DataException("Select version as one of: {}".format(','.join(RELEASES)))
+
+    if data_type == 'exomes':
+        return f'gs://gnomad/liftover/release/2.1.1/ht/exomes/gnomad.exomes.r{version}.liftover.ht'
+    elif data_type == 'genomes':
+        return f'gs://gnomad/liftover/release/2.1.1/ht/genomes/gnomad.genomes.r{version}.liftover.ht'
```
konradjk: You might consider `f'gs://gnomad/liftover/release/2.1.1/ht/{data_type}/gnomad.{data_type}.r{version}.liftover.ht'` in lieu of the if blocks. Also, once we're satisfied with these, they should go in gnomad-public, but that can obviously wait.

resources/basics.py
```diff
@@ -496,6 +496,39 @@ def qc_meta_path(data_type: str):
         return 'gs://gnomad/sample_qc/input_meta/gnomad.genomes.streamlined_metadata.2018-10-11.txt.bgz'
 
 
+# liftover data
+def get_gnomad_liftover_data_path(data_type, split=True, version=CURRENT_RELEASE) -> str:
+    """
+    Function to get paths to liftover gnomAD Tables
+    :param str data_type: One of `exomes` or `genomes`
+    :param bool split: Whether the dataset should be split
```
konradjk: `split` isn't used (and I don't think it needs to be - pretty much all tables are assumed to be split).

resources/basics.py
konradjk: Also you'll want to put `version` in here, even though only 2.1.1 will be supported for now (or remove `version`)

resources/basics.py
```diff
@@ -496,6 +496,39 @@ def qc_meta_path(data_type: str):
...
+    elif data_type == 'genomes':
+        return f'gs://gnomad/liftover/release/2.1.1/ht/genomes/gnomad.genomes.r{version}.liftover.ht'
+    return DataException("Select data_type as one of 'genomes' or 'exomes'")
+
+
+def get_gnomad_liftover_data(data_type, split=True, version=CURRENT_RELEASE) -> hl.Table:
+    """
+    Wrapper function to get liftover gnomAD data as Table
+    :param str data_type: One of `exomes` or `genomes`
+    :param bool split: Whether the dataset should be split
```
konradjk: same as above

resources/liftover.py
```diff
@@ -0,0 +1,192 @@
...
+logging.basicConfig(format="%(asctime)s (%(name)s %(lineno)s): %(message)s", datefmt='%m/%d/%Y %I:%M:%S %p')
+logger = logging.getLogger("liftover")
+logger.setLevel(logging.INFO)
+
+
+def get_checkpoint_path(gnomad: bool, data_type: str, table_path: str) -> str:
+    """
+    Creates a checkpoint path for Table
+    :param bool gnomad: Whether data is gnomAD data
+    :param str data_type: Data type (exomes or genomes for gnomAD; None otherwise)
```
konradjk: `not used otherwise`

resources/liftover.py
```diff
@@ -0,0 +1,192 @@
...
+    :param ReferenceGenome rg: Reference genome
+    :return: Table with liftover annotations
+    :rtype: Table
+    """
+
+    # annotate release file with liftover coordinates
+    ht = ht.annotate(new_locus=hl.liftover(ht.locus, rg, include_strand=True),
+                     old_locus=ht.locus
+                    )
+    ht = ht.key_by(locus=ht.new_locus.result, alleles = ht.alleles)
```
konradjk: `s/ = /=/`

resources/liftover.py
```diff
@@ -0,0 +1,192 @@
...
+            ) 
+
+        logger.info('Lifting ht to b{}'.format(args.build))
+        ht = lift_ht(ht, gnomad, data_type, table_path, rg37)
+
+        logger.info('Checking SNPs for reference mismatches')
+        ht = annotate_snp_mismatch(ht, data_type, rg37)
+
+    mismatch = check_mismatch(ht)
+    logger.info('{} reference mismatches in SNPs'.format(mismatch['total_mismatch']))
```
konradjk: Would be worth printing the full count and `'negative_strand'` count here as well

resources/liftover.py
ch-kr: oh oops, missed this one, thanks!

resources/basics.py
```diff
@@ -496,6 +496,36 @@ def qc_meta_path(data_type: str):
...
+    :return: Path to chosen Table
+    :rtype: str
+    """
+    if version not in RELEASES:
+        return DataException("Select version as one of: {}".format(','.join(RELEASES)))
+
+    if data_type != 'exomes' and data_type != 'genomes': 
+        return DataException("Select data_type as one of 'genomes' or 'exomes'")
+
+    return f'gs://gnomad/liftover/release/{version}/ht/{data_type}/gnomad.{data_type}.r{version}.liftover.ht'
```
konradjk: Isn't this in `gnomad-public` now?

resources/basics.py
```diff
@@ -496,6 +496,36 @@ def qc_meta_path(data_type: str):
...
+
+
+def get_gnomad_liftover_data(data_type, version=CURRENT_RELEASE) -> hl.Table:
```
konradjk: I don't think this function is needed as is, since it's just reading a table from another function. I wonder if this should be folded into `get_gnomad_public_data` (in order to actually do the join), but then I'm a little worried there will be lots and lots of shuffles happening.

utils/liftover.py
```diff
@@ -0,0 +1,194 @@
...
+    :param str table_path: Path to input Table (if data is not gnomAD data)
+    :param ReferenceGenome rg: Reference genome
+    :return: Table with liftover annotations
+    :rtype: Table
+    """
+
+    # annotate release file with liftover coordinates
+    ht = ht.annotate(new_locus=hl.liftover(ht.locus, rg, include_strand=True),
+                     old_locus=ht.locus
+                    )
```
konradjk: Do you not want to filter out loci that didn't liftover? `hl.is_defined(ht.new_locus)`

utils/liftover.py
```diff
@@ -0,0 +1,194 @@
...
+        logger.info('Filtering to chr21 for testing')
+        if args.build == 38:
+            ht = ht.filter(ht.locus.contig == '21')
+        else:
+            ht = ht.filter(ht.locus.contig == 'chr21')
+
+    logger.info('Adding chain file and loading fasta sequence for destination build')
+    if args.build == 38:
+
+        rg37.add_liftover('gs://hail-common/references/grch37_to_grch38.over.chain.gz', rg38)
```
konradjk: Not necessary, but I'd probably reframe these as "source" and "target" (getting the right one each at line 101), and then you can avoid duplicating code in the if/else block.

resources/basics.py
ch-kr: I can remove this function for now

utils/liftover.py
ch-kr: Good point - I think I'll count the number that didn't liftover and then remove

utils/liftover.py
```diff
@@ -97,9 +101,24 @@ def main(args):
...
+        source = hl.get_reference('GRCh37')
+        target = hl.get_reference('GRCh38')
+        chain = 'gs://hail-common/references/grch37_to_grch38.over.chain.gz'
+        target.add_sequence(
+            'gs://hail-common/references/Homo_sapiens_assembly38.fasta.gz', 
+            'gs://hail-common/references/Homo_sapiens_assembly38.fasta.fai'
+            )
+    else:
+        source = hl.get_reference('GRCh38')
+        target - hl.get_reference('GRCh37')
```
konradjk: `=`

utils/liftover.py
```diff
@@ -130,33 +149,11 @@ def main(args):
...
-        rg37.add_liftover('gs://hail-common/references/grch37_to_grch38.over.chain.gz', rg38)
-        rg38.add_sequence(
-            'gs://hail-common/references/Homo_sapiens_assembly38.fasta.gz', 
-            'gs://hail-common/references/Homo_sapiens_assembly38.fasta.fai'
-            )
-
-        logger.info('Lifting ht to b{}'.format(args.build))
-        ht = lift_ht(ht, gnomad, data_type, table_path, rg38)
+    logger.info('Lifting ht to b{}'.format(args.build))
+    ht = lift_ht(ht, gnomad, data_type, table_path, rg38)
```
konradjk: `target`

utils/liftover.py
```diff
@@ -130,33 +149,11 @@ def main(args):
...
-            'gs://hail-common/references/human_g1k_v37.fasta.fai'
-            ) 
-
-        logger.info('Lifting ht to b{}'.format(args.build))
-        ht = lift_ht(ht, gnomad, data_type, table_path, rg37)
-
-        logger.info('Checking SNPs for reference mismatches')
-        ht = annotate_snp_mismatch(ht, data_type, rg37)
+    logger.info('Checking SNPs for reference mismatches')
+    ht = annotate_snp_mismatch(ht, data_type, rg38)
```
konradjk: here too

PR #95, opened 2019-04-03
Make `adj` more flexible, reduce required coverage when haploid

lfrancioli:


konradjk:
Looks good. One thing, this should probably be an optional flag so that we have documentation that this is not how we did it in v2

lfrancioli:
I added an extra argument for it, so it's clearer / more flexible. I've set it to 10 as default to reflect gnomAD v2 value.

PR #96, opened 2019-04-10
Added from/to phred and FS computation from SB annotation

lfrancioli:
Whoever is going to review this should double-check the logic against GATK (URL in comment)

### DIFF
--- a/utils/generic.py
+++ b/utils/generic.py
@@ -1176,14 +1176,24 @@ def from_phred(phred_score_expr: hl.expr.NumericExpression) -> hl.expr.Float64Ex
     return 10 ** -(phred_score_expr / 10)
 
 
-def fs_from_sb(sb: Union[hl.expr.ArrayNumericExpression, hl.expr.ArrayExpression]) -> hl.expr.Int64Expression:
+def fs_from_sb(
+        sb: Union[hl.expr.ArrayNumericExpression, hl.expr.ArrayExpression],
+        normalize: bool = True,
+        min_cell_count: int = 200,
+        min_count: int = 4,
+        min_p_value: float = 1e-320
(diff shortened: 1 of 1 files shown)


utils/generic.py
```diff
@@ -1152,3 +1152,87 @@ def inbreeding_coeff(gt_counts: hl.expr.DictExpression) -> hl.expr.Float32Expres
...
+        sb, sb_sum
+    )
+
+    # FET
+    fs_expr = to_phred(
+        hl.max(
+            hl.fisher_exact_test(
+                fs_expr[0], fs_expr[1], fs_expr[2], fs_expr[3]
+            ).p_value,
+            MIN_PVALUE # Minimal p-val
```
konradjk: i... don't think you need this comment :-D

lfrancioli:
I like using chi-squared instead of norm. Have added this now!

utils/generic.py
```diff
@@ -1176,14 +1176,24 @@ def from_phred(phred_score_expr: hl.expr.NumericExpression) -> hl.expr.Float64Ex
...
-    - If sum(counts) > 400, they are normalized
-    - If sum(counts) < 4, returns missing
-    - Any p-value below 1e-320 is truncated to that value
+
+    Using default values will have the same behavior as the GATK implementation, that is:
+    - If sum(counts) > 2*`min_cell_count` (default to GATK value of 200), they are normalized
+    - If sum(counts) < `min_count` (default to GATK value of 4), returns missing
+    - Any p-value < `min_p_value` (default to GATK value of 1e-320) is truncated to that value
+
+    In addition to the default GATK behavior, setting `normalize` to `True` will perform a chi-squared test
```
konradjk: I think you mean `False` here?

PR #101, opened 2019-05-07
Sklearnot

lfrancioli:


utils/generic.py
```diff
@@ -871,14 +870,14 @@ def assign_population_pcs(
         pop_pca_scores: Union[hl.Table, pd.DataFrame],
         pc_cols: Union[hl.expr.ArrayExpression, List[str]],
         known_col: str = 'known_pop',
-        fit: RandomForestClassifier = None,
+        fit: Any = None,
```
gtiao: Change docstrings on line 902 to reflect the change here (or make note that it should be RandomForestClassifier but you aren't able to enforce it in the typehints)

lfrancioli: So I really think we want to keep the docstring type correct; then at least it's somewhere and when we generate doc it'll be correct. I've put a comment as to why the typehint is different next to the argument type/return type.

PR #104, opened 2019-05-29
Some VEP resources

konradjk:
Adding vep_or_lookup_vep, some paths to resources, and a minor fix to process_csq in light of OS

### DIFF
--- a/resources/basics.py
+++ b/resources/basics.py
@@ -434,7 +434,7 @@ def cpg_sites_ht_path():
     return 'gs://gnomad-public/resources/methylation/cpg.ht'
 
 
-def methylation_sites_mt_path(hail_version=CURRENT_HAIL_VERSION, ref='GRCh37'):
+def methylation_sites_ht_path(hail_version=CURRENT_HAIL_VERSION, ref='GRCh37'):
     if ref not in REFERENCES:
         return DataException("Select reference as one of: {}".format(','.join(REFERENCES)))
     if ref == 'GRCh37':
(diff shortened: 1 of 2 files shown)


utils/generic.py
```diff
@@ -232,6 +232,41 @@ def filter_low_conf_regions(mt: Union[hl.MatrixTable, hl.Table], filter_lcr: boo
...
+    :param Table ht: Input Table
+    :param Table reference_vep_ht: A reference database with VEP annotations (must be in top-level `vep`)
+    :return: VEPped Table
+    :rtype: Table
+    """
+    from gnomad_hail.resources.basics import vep_config_path, context_ht_path
+
+    VEP_REFERENCES = {
+        'GRCh37': context_ht_path(),
+        'GRCh38': '',
```
konradjk: fully aware this is missing - i'll probably VEP something up today and put it in here

resources/basics.py
```diff
@@ -469,10 +469,25 @@ def methylation_sites_mt_path(hail_version=CURRENT_HAIL_VERSION, ref='GRCh37'):
...
+        return 'gs://hail-common/vep/vep/vep95-GRCh38-loftee-gcloud.json'
+
+vep_config = vep_config_path()
+
+
+def context_ht_path(reference='GRCh37'):
+    if reference not in ('GRCh37', ):
+        return DataException("Reference must be GRCh37")
+    if reference == 'GRCh37':
+        return 'gs://gnomad-public/papers/2019-flagship-lof/v1.0/context/Homo_sapiens_assembly19.fasta.snps_only.vep_20181129.ht'
```
knguyen142: REFERENCE can be a a dict of dicts for multiple values. I think in terms of configs it's better to put all config related things in one place so it can easily be changed.

utils/generic.py
```diff
@@ -232,6 +232,46 @@ def filter_low_conf_regions(mt: Union[hl.MatrixTable, hl.Table], filter_lcr: boo
     return mt
 
 
+def vep_or_lookup_vep(ht, reference_vep_ht=None, reference=None, vep_config=None):
```
knguyen142: Should you pass through `block_size`, `name` and `csq`? https://hail.is/docs/0.2/methods/genetics.html?highlight=vep#hail.methods.vep

knguyen142:
Thanks @konradjk! Can I ask what the memoized vep cache will have and roughly how big?
Also, have you done any benchmarks for time comparisons? It's my understanding that in certain cases when the table to be vep'd has an overlap with many partitions in a large table for a join, whole partitions in the large table would be read in. I imagine for smaller tables it will be very fast. I'm curious of the performance since the Hail team seems to think Vep has gotten faster https://discuss.hail.is/t/vep-memoization-caching/920.

We're doing memoization at a higher level (in the joins and our own computations on the vep which would include vep annotations), but when we have new data to vep if this is a significant bump it'd be very useful.

resources/basics.py
konradjk: I generally agree - the reason we didn't do this before is we wanted the flexibility of functions to specify things like hail version. But I think that ship has sailed now, so unless anyone has any objections, I'll fix it to do so.

utils/generic.py
konradjk: Given that we've never used block_size or name, and I expressly do not want to allow `csq` until we think through how we could have a second VEP file with the string schema, I'm tempted not to for now.

resources/basics.py
knguyen142: Hmm, how would the caller specify the hail version? Also, I'm not too familiar with this package but I assume it's okay to just pin it to hail v02 since we'd never import hail v01 and v02, right?

konradjk: Yeah, a bunch of these functions had `hail_version=...` as arguments, but I think the utility of these is long gone

knguyen142:
@konradjk we can try it out and see how it performs in our benchmark. I'm curious since some info from our end suggests joining with big datasets is more costly than running vep, but this might help us figure that out. Who usually does the review for this repo so we can merge?

konradjk:
Since you've given a review, I'm happy to force a merge.

PR #109, opened 2019-08-14
Liftover

ch-kr:
I added some features and edits discussed in code review a few months ago:

1.  The code can now liftover a mt, not just a ht
2.  This infers the build (and assumes a destination build based on the inferred build)
3.  It splits the input mt/ht if it hasn't already been split

### DIFF
--- a/utils/liftover.py
+++ b/utils/liftover.py
@@ -1,6 +1,6 @@
 import argparse
 from gnomad_hail.resources import *
-from gnomad_hail.utils.generic import flip_base 
+from gnomad_hail.utils.generic import flip_base, get_reference_genome 
 import logging
 from os.path import dirname, basename
 import sys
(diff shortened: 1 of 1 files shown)


utils/liftover.py
```diff
@@ -11,74 +11,122 @@
...
+    t = t.filter(keep_expr) if isinstance(t, hl.Table) else t.filter_rows(keep_expr)
+    
+    row_key_expr = {
+        'locus': t.new_locus.result,
+        'alleles': t.alleles
+    }
+    t = t.key_by(**row_key_expr) if isinstance(t, hl.Table) else t.key_rows_by(**row_key_expr)
+
+    logger.info('Writing out lifted over data')
+    t = t.checkpoint(get_checkpoint_path(gnomad, data_type, path, isinstance(t, hl.Table)), overwrite=True)
```
konradjk: as a general stylistic thing, you may want to parameterize `overwrite` and have that something you pass into the function (so that someone running it down the line doesn't accidentally overwrite a release file)

PR #113, opened 2019-09-20
Removed hardcoded AD indices from get_adj_expr

ch-kr:
I noticed that the adj expression expects the input data to have only biallelics, but we annotate with adj before splitting multiallelics (https://github.com/macarthur-lab/gnomad_qc/blob/master/sample_qc/generate_hardcalls.py)


konradjk:
Oops, I realized this should be `gt_expr` not `mt.GT` (since that's a parameter rather than a hard-coding of the field)

ch-kr:
OH you're right! Will edit

PR #118, opened 2019-10-07
Create LICENSE

gtiao:


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 MacArthur Lab
```
konradjk: Let's name folks. Me, you, Laurent, and Daniel I think

PR #121, opened 2019-10-29
Sparse mt utils

lfrancioli:


utils/sparse_mt.py
```diff
@@ -0,0 +1,432 @@
...
+                    len(ref.y_contigs),
+                    ",".join(ref.y_contigs)
+                )
+            )
+        chr_y = ref.y_contigs[0]
+
+    def get_contig_size(contig: str) -> int:
+        contig = hl.utils.range_table(ref.contig_length(contig), n_partitions=int(ref.contig_length(contig) / 500_000))
+        contig = contig.annotate(
+            locus=hl.locus(contig=contig, pos=contig.idx + 1)
```
ch-kr: maybe it's a good idea to add `reference_genome=ref`? otherwise this assumes the default reference genome matches whichever build the mt is in

utils/sparse_mt.py
```diff
@@ -0,0 +1,432 @@
...
+        return contig_size
+
+    def get_chr_dp_ann(chr: str) -> hl.Table:
+        contig_size = get_contig_size(chr)
+        chr_mt = hl.filter_intervals(mt, [hl.parse_locus_interval(chr)])
+        return chr_mt.select_cols(**{
+            f'{chr}_mean_dp': hl.agg.sum(hl.cond(chr_mt.LGT.is_hom_ref(), chr_mt.DP * (chr_mt.END - chr_mt.locus.position), chr_mt.DP)) / contig_size
+        }).cols()
+
+    normalizaion_chrom_dp = get_chr_dp_ann(normalization_contig)
```
ch-kr: normalizaion -> normalization

utils/generic.py
```diff
@@ -1409,3 +1444,140 @@ def get_csq_from_struct(element: hl.expr.StructExpression, feature_type: str) ->
...
+
+    # Make sure the HT is keyed by locus, alleles
+    info_ht = info_ht.key_by('locus', 'alleles')
+
+    # Convert int64 fields to int32 (int64 isn't supported by VCF)
+    for f, ft in info_ht.info.dtype.items():
+        if ft == hl.dtype('int64'):
+            logger.warning(f"Coercing field info.{f} from int64 to int32 for VCF output. Value will be capped at int32 max value.")
+            info_ht = info_ht.annotate(
+                info=info_ht.info.annotate(**{f: hl.int32(hl.min(2**31 - 1, info_ht.info[f]))})
```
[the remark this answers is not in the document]
lfrancioli: well, this unfortunately means that you loose the information about any number greater than 2,147,483,647 ... It could potentially be bad for some info. For the current info fields, it's totally fine but an example of a situation where it could be bad would be it you were to store a global genomic position (so, max would be 3.2bln) in an info field. 
I'm open to suggestions for another behavior!

utils/generic.py
```diff
@@ -1409,3 +1444,140 @@ def get_csq_from_struct(element: hl.expr.StructExpression, feature_type: str) ->
...
+
+    # Create an MT with no cols so that we acn export to VCF
+    info_mt = info_ht.to_matrix_table_row_major(columns=['s'], entry_field_name='s')
+    return info_mt.filter_cols(False)
+
+
+def sort_intervals(intervals: List[hl.Interval]):
+    """
+    Sorts an array of intervals by:
+    start contig, then start position, then end contig, then end position
```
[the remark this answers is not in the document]
lfrancioli: Typically not, but they  can...

utils/sparse_mt.py
```diff
@@ -0,0 +1,432 @@
+from .generic import *
+
+INFO_SUM_AGG_FIELDS= ['QUALapprox']
+INFO_INT32_SUM_AGG_FIELDS = ['VarDP']
+INFO_MEDIAN_AGG_FIELDS = ['ReadPosRankSum', 'MQRankSum']
+INFO_ARRAY_SUM_AGG_FIELDS = ['SB', 'RAW_MQandDP']
+
+def compute_last_ref_block_end(mt: hl.MatrixTable) -> hl.Table:
```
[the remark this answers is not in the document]
lfrancioli: this is a tough one indeed, I'll try adding some comments

utils/sparse_mt.py
lfrancioli: good catch!

utils/sparse_mt.py
```diff
@@ -18,23 +18,34 @@ def compute_last_ref_block_end(mt: hl.MatrixTable) -> hl.Table:
...
     ht = mt._localize_entries('__entries', '__cols')
+
+    # Compute the position by using hl.scan._prev_nonnull.
+    # This was inspired by hl.experimental.densify
+    # _prev_non_null is an aggregator that keeps the previous record in memory
+    # and updates it with the given value at the row if it's not null (missing)
+    # The following code computes the following annotation for each row:
+    # 1. Keep a scan of the entries using _prev_nonnull, keeping the start (ht.locus) and end (entry.END) of each ref block  (1.1)
+    # 2. For the current row locus, record the start of the block that starts the furthest away,
+    #    that is the minimum position in the current scan for any block that overlaps the current locus (2.1)
```
ch-kr: thank you for all of these -- they're so helpful!!

utils/generic.py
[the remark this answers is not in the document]
ch-kr: ohh I see, thanks! it seems really silly that VCFs don't support int32 if this cap could impact some of the info values. the warning seems reasonable to me though.

utils/sparse_mt.py
lfrancioli: Of course! And yes, this is complex code using undocumented (may I say secret?) hail aggregator, so additional comments were definitely warranted! :)

PR #123, opened 2019-11-14
New structure for resources

lfrancioli:


resources/resource_utils.py
```diff
@@ -0,0 +1,60 @@
...
+        The `path`/`source_path` attributes of the versioned resource are those
+        of default version of the resource (first in the dict keys).
+
+        In addition, all versions of the resource are stored in the `versions` attribute.
+
+        :param dict of str -> _Resource versions: A dict of version name -> resource. The default version should be first in the dict keys.
+        :return: A Resource with versions
+        :rtype: _Resource
+        """
+        latest = versions[list(versions)[0]]
```
nawatts: Using a `default_version` keyword argument would be a clearer way to do this. Dicts don't connote an ordering, so when looking at code where a versioned resource is instantiated, it's not obvious that the order of keys is significant. That would require reading this docstring. Also, it's only in Python 3.7+ that the order of the dictionary keys is guaranteed to be preserved.

resources/resource_utils.py
```diff
@@ -0,0 +1,60 @@
+from typing import Optional, Dict
+import hail as hl
+
+
+# Resource classes
+class _Resource:
```
nawatts: Should this be public? The base class could be useful for typing.

resources/resource_utils.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Generic abstract resource class.
+    """
+    def __init__(self, path: str, source_path: Optional[str] = None, **kwargs):
+        if type(self) is _Resource:
+            raise TypeError("Can't instantiate abstract class _Resource")
+
+        self.source_path = source_path
+        self.path = path
```
nawatts: Some documentation explaining the difference between `source_path` and `path` would be nice. So far as I can tell, `source_path` is only used when the resource is imported from a non-Hail format? But the attribute names don't make that clear, `source_path` could easily be interpreted as the path from which a Table is read.

resources/resource_utils.py
```diff
@@ -0,0 +1,60 @@
...
+        self.__dict__.update(kwargs)
```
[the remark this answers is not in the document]
nawatts: Since `source_path` is already optional and unused in some resources and the arbitrary keyword args seem to only be used for importing data (`dbsnp.vcf_header_path`), maybe those could be combined into one optional dict argument `import_sources`. That would isolate use of arbitrary attributes to only reference_data_import.py, which already treats each individual resource differently.

resources/grch38/reference_data_import.py
```diff
@@ -0,0 +1,81 @@
...
+            min_partitions=100
+        ).rows().write(clinvar.path, overwrite=args.overwrite)
+
+    if args.na12878:
+        hl.import_vcf(
+            na12878_giab.source_path, force_bgz=True, min_partitions=100
+        ).write(na12878_giab.path, overwrite=args.overwrite)
+
+    if args.dbsnp:
+        hl.import_vcf(
```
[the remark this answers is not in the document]
lfrancioli: yup

resources/resource_utils.py
lfrancioli: yeah, why not? Should we use a different name then? AbstractResource or BaseResource ?

resources/resource_utils.py
lfrancioli: I'm open, but I think I like keeping source_path -- there are really quite a few resources that will have this. Totally agree on the documentation and open to suggestions for other names!

resources/resource_utils.py
nawatts: I was thinking Resource without the underscore, since the underscore implies that it is not to be used outside this module.

resources/resource_utils.py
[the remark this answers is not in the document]
lfrancioli: So, the big advantage with classes is the ability to have TableResource and TableMatrixResource that each can load their corresponding resource automatically. I think it's so much nicer this way.

Now, I agree that using kwargs may not be the best option here. Maybe just an optional attributes Dict would be better? We need it for the occasional additional file that should be bundled but it'll be a minor use case.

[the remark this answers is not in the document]
lfrancioli: As for using a dict for `import_sources`, if others find this better I'm fine with it but I don't really see the advantage over having a clear optional `source_path` attribute.

resources/resource_utils.py
lfrancioli: Yup totally agree!

resources/grch38/reference_data.py
```diff
@@ -0,0 +1,50 @@
...
+)
+
+na12878_giab = MatrixTableResource(
+    path='gs://gnomad-public/truth-sets/hail-0.2/HG001_GRCh38_GIAB_highconf_CG-IllFB-IllGATKHC-Ion-10X-SOLID_CHROM1-X_v.3.3.2_highconf_PGandRTGphasetransfer.mt',
+    source_path='gs://gnomad-public/truth-sets/source/HG001_GRCh38_GIAB_highconf_CG-IllFB-IllGATKHC-Ion-10X-SOLID_CHROM1-X_v.3.3.2_highconf_PGandRTGphasetransfer.vcf.gz'
+)
+
+# Versioned resources: versions should be listed from most recent to oldest
+clinvar = TableResource.versioned({
+    '20181028': TableResource(
```
nawatts: Should this match the date in the HT/VCF file names?

lfrancioli: duh! thanks!

lfrancioli:
Thanks @nawatts for all the comments! Just pushed another version trying to address them. One thing I was debating is if I should add the `versions` to all of the resources in a versioned resource. If we do, I could also add a pointer to the default_resource. 
All in all, I think it's overkill (but happy to do) but thought I'd throw this in.

lfrancioli:
All right, couldn't help myself -- I did try something slightly different for the versioned resource. I'm overthinking this, so hopefully we can converge on something nice and easy and not overthought :)

resources/resource_utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+# Resource classes
+class BaseResource:
+    """
+    Generic abstract resource class.
+    """
+
+
+    def __init__(self, path: str, source_path: Optional[str] = None, attributes: Dict[str, Any] = None):
```
nawatts: Is `attributes` expected to be mostly used for resources that are imported from multiple files? Could it be named something more specific, like `other_sources`?

resources/resource_utils.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        if type(self) is BaseResource:
+            raise TypeError("Can't instantiate abstract class BaseResource")
+
+        self.source_path = source_path
+        self.path = path
+        self.attributes = attributes
+
+    def __repr__(self):
+        attr_str = [fr'path={self.path}']
```
mike-w-wilson: For my own knowledge why use the raw string?

resources/resource_utils.py
```diff
@@ -0,0 +1,101 @@
...
+        super().__init__(
+            path=versions[default_version].path,
+            source_path=versions[default_version].source_path,
+            attributes=versions[default_version].attributes
+        )
+        self.default_version = default_version
+        self.versions = versions
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}(default_version={self.default_version}, default_resource={self.versions[self.default_version]}, versions={list(self.versions.keys())})'
```
mike-w-wilson: Why not use r strings here?

resources/grch38/reference_data.py
```diff
@@ -0,0 +1,59 @@
...
+    versions={
+        'b151': TableResource(
+            path='gs://gnomad-public/resources/grch38/dbsnp_b151_grch38_all_20180418.ht',
+            source_path='gs://gnomad-public/resources/grch38/dbsnp_b151_grch38_all_20180418.vcf.bgz',
+            attributes={'vcf_header_path': 'gs://gnomad-public/resources/grch38/dbsnp_b151_grch38_all_20180418.vcf.header'}
+        )
+    }
+)
+
+hapmap = TableResource(
```
mike-w-wilson: Should  `hap_map`,`kgp_omni`, and `mills` be moved above the versioned resources? Or if not and they are versioned, why dont they have a version?

resources/resource_utils.py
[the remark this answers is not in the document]
lfrancioli: sorry -- this is a remnant of intermediate stuff I was doing, thanks for catching this.

lfrancioli: sorry -- this is a remnant of intermediate stuff I was doing, thanks for catching this.

resources/grch38/reference_data.py
lfrancioli: they could, but does the order really matter?

lfrancioli:
OK, so I'm back in Boston and after thinking about this I suggest the following:
1. Let's find a better name for `source_path` -- it's clearly not working :) `import_path`? Or maybe as @nawatts suggested have an `import_sources` dict for all import params? (I've warmed up to the idea ;) )

2. I think the best way forward is if @mike-w-wilson goes ahead and takes it from there. In particular, moving the v2 and ukbb resources to the new framework would be great to make sure that it can handle nicely what we need. Also, would be good to think a bit about the `get_gnomad_mt` type of functions (which typically take a bunch of args). Do we want to keep them or do we want to subclass and include all these args in their `.mt()` and `.ht()` functions.
A simple example of this for a v3 resources would be:

```
class QCMT(MatrixTableResource):

    def mt(self, remove_hard_filtered_samples: bool = True) -> hl.MatrixTable:
        qc_mt = hl.read_matrix_table(self.path)
        if remove_hard_filtered_samples:
            hard_filtered_ht = hl.read_table(hard_filtered_samples_ht_path)
            qc_mt = qc_mt.filter_cols(hl.is_missing(hard_filtered_ht[qc_mt.col_key]))
        return qc_mt

qc_mt = QCMT(path='gs://gnomad/sample_qc/mt/genomes_v3/gnomad_v3_qc_mt_v2_sites_dense.mt')
```

On the one hand I feel like subclassing may be the cleanest way to do this. On the other, maybe it's overkill?

mike-w-wilson:
Sorry this took a while @lfrancioli and @nawatts , I was pulled off for other work. I only pushed a small commit for now to make sure we are happy with the dictionary approach.

lfrancioli:
I'm good with it!

resources/grch38/reference_data.py
```diff
@@ -0,0 +1,74 @@
+
```
nawatts: This module needs to import the resource classes that it uses.

resources/grch38/reference_data_import.py
```diff
@@ -0,0 +1,82 @@
+import argparse
+from .reference_data import *
```
nawatts: This module needs to import Hail.

mike-w-wilson:
@lfrancioli I've finally pushed the changes I've made to resources. I've added all gnomAD files that I believed to be relevant to the gnomad.py modules and have removed the basics, sample_qc, and variant_qc modules for now since they are v2 specific. For reference data, I've mimicked Laurent's GRCh38 approach in the GRCh37 directory. But there are a lot more reference data files for 37 right now. Some of them do not yet have Tables and are just intervals while others are already MatrixTables. For now, reference_data_import.py is using either TableResource or MatrixTableResource from resource_utils.py depending on what is currently written in gnomad-public bucket. For the intervals/bed files, I have them write to Tables in the import script but have not yet run it so most interval.ht do not exist. I wanted to confirm this looked ok before doing any compute. I also ran pylint and black on all the files so I apologize as the GRCh38 reference data files diffs may look pretty gnarly. I only addressed @nawatts  import feedback in those files.

resources/grch37/gnomad.py
```diff
@@ -0,0 +1,345 @@
...
+
+    if data_type == "exomes":
+        current_fam = CURRENT_EXOME_FAM
+    else:
+        current_fam = CURRENT_GENOME_FAM
+
+    return PedigreeResource(path=_fam_path(data_type, current_fam, true_trios))
+
+
+def vep_config_path():
```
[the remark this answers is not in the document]
mike-w-wilson: I agree, it felt out of place putting it here but I wasn't sure where else it could live or if it was even necessary. It looks to only be used in https://github.com/macarthur-lab/gnomad_hail/blob/03b9d7e6ff7e57828f880bc74145f540c8c485b5/utils/generic.py#L271 so perhaps it should be moved into that module?

resources/grch37/reference_data.py
```diff
@@ -0,0 +1,181 @@
...
+    path="gs://gnomad-public/truth-sets/hail-0.2/1000G_omni2.5.b37.mt",
+    import_sources={
+        "source_path": "gs://gnomad-public/truth-sets/source/1000G_omni2.5.b37.vcf.bgz",
+    },
+)
+
+mills = MatrixTableResource(
+    path="gs://gnomad-public/truth-sets/hail-0.2/Mills_and_1000G_gold_standard.indels.b37.mt",
+    import_sources={
+        "source_path": "gs://genomics-public-data/resources/broad/hg38/v0/Mills_and_1000G_gold_standard.indels.hg38.vcf.gz",
```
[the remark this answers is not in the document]
mike-w-wilson: 🤦‍♂ , found the original source and will update.

resources/grch37/gnomad.py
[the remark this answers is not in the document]
lfrancioli: sounds good to me!

mike-w-wilson:
@nawatts @lfrancioli , I had a few typos in reference data paths and have updated utils to use the new resources after Nick pointed out that modules outside of resources would be impacted. Would either of you mind checking this out again? Thank you!

utils/generic.py
```diff
@@ -278,7 +278,7 @@ def vep_or_lookup_vep(ht, reference_vep_ht=None, reference=None, vep_config=None
     :return: VEPped Table
     :rtype: Table
     """
-    from gnomad_hail.resources.basics import vep_config_path, context_ht_path
+    from gnomad_qc.resources.basics import vep_config_path, context_ht_path
```
lfrancioli: we need this to work without importing `gnomad_qc` resources. the VEP configs should defintely stay in the `gnomad_hail` repo.

utils/gnomad_functions.py
```diff
@@ -372,7 +372,7 @@ def get_rf_runs(data_type: str) -> Dict:
     :rtype: dict
     """
     
-    from gnomad_hail.resources.variant_qc import rf_run_hash_path
+    from gnomad_qc.resources.variant_qc import rf_run_hash_path
```
lfrancioli: Can you add a `#FIXME` here, as code should be adapted to have no reference to `gnomad_qc`

utils/variant_qc_plots.py
```diff
@@ -1,5 +1,5 @@
 from gnomad_hail import *
-from gnomad_hail.resources.variant_qc import *
+from gnomad_qc.resources.variant_qc import *
```
lfrancioli: Same here, please add a `#FIXME` so we know this is something that needs to be tackled.

nawatts:
Went through and resolved a few remaining import errors in utils. Pylint also reports false positives on imports from `bokeh.palettes` and `pyspark.sql.functions` since those modules dynamically generate their exports, so I added comments to ignore messages about them.

mike-w-wilson:
@lfrancioli, back to you!

lfrancioli:
From github, it looks like there are a few conflicts w.r.t. master. Not sure what they are, but a merge will be needed. I suspect this can be done on merging, but have you had a look?

Otherwise, I'm happy with this!

PR #124, opened 2019-11-19
V3 variant qc utils

lfrancioli:


utils/variant_qc.py
```diff
@@ -0,0 +1,346 @@
...
+            proband_is_female: Optional[hl.expr.BooleanExpression]
+    ) -> hl.expr.BooleanExpression:
+        """
+        Helper method to get whether a given genotype combination is a DNM at a given locus with a given proband sex.
+        """
+        if proband_is_female is None:
+            logger.warning("Since no proband sex expression was given to generate_fam_stats_expr, only DNMs in autosomes will be counted.")
+            return hl.or_missing(
+                locus.in_autosome(),
+                trio_mt.proband_entry.GT.is_het() & trio_mt.father_entry.GT.is_hom_ref() & trio_mt.mother_entry.GT.is_hom_ref()
```
jkgoodrich: Can this be `proband_gt.is_het() & father_gt.is_hom_ref() & mother_gt.is_hom_ref()` like below, or is there a reason that it needs to use the trio_mt?

utils/variant_qc.py
```diff
@@ -0,0 +1,346 @@
...
+
+    :param MatrixTable trio_mt: A trio stndard trio MT (with the format as produced by hail.methods.trio_matrix
+    :param dict of str -> BooleanExpression transmitted_strata: Strata for the transmission counts
+    :param dict of str -> BooleanExpression de_novo_strata: Strata for the de novo counts
+    :param BooleanExpression proband_is_female_expr: An optional expression giving the sex the proband. If not given, DNMs are only computed for autosomes.
+    :return: An expression with the counts
+    :rtype: StructExpression
+    """
+
+    # Create map for transmitted, untransmitted and DNM
```
jkgoodrich: Are there any other cases where having these values would be useful? If so maybe move outside this function

utils/variant_qc.py
```diff
@@ -0,0 +1,346 @@
...
+                proband_gt.is_het() & father_gt.is_hom_ref() & mother_gt.is_hom_ref(),
+                hl.or_missing(
+                    ~proband_is_female,
+                    proband_gt.is_het() & father_gt.is_hom_ref()
+                )
+            )
+        )
+
+    # Create transmission counters
+    fam_stats_expr = {
```
jkgoodrich: Not needed, but Is there anyway to merge this and the one below it since you are just doing the same thing but taking the first or second entry with a small change to the name

utils/variant_qc.py
```diff
@@ -0,0 +1,346 @@
...
+                )[1]
+            )
+        ) for name, expr in transmitted_strata.items()
+    })
+
+    # Create de novo counters
+    fam_stats_expr.update({
+        f'n_de_novo_{name}': hl.agg.filter(
+            _get_composite_filter_expr(
+                _is_dnm(
```
jkgoodrich: I think this should be proband, father, mother

utils/variant_qc.py
```diff
@@ -0,0 +1,346 @@
...
+            [x[0] for x in indexed_bins],
+            [x[1] for x in indexed_bins]
+        )
+
+    ht = ht.annotate(
+        **{f'_filter_{rid}': rexpr for rid, rexpr in rank_expr.items()},
+        _score=score_expr
+    )
+
+    logger.info(f'Adding rank using approximate_quantiles binned into {n_bins}, using k={10*n_bins}')
```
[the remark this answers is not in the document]
lfrancioli: So I'm not quite sure how `k` and `n_bins` play together to be honest.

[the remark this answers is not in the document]
lfrancioli: So I want to refrain setting them using a single parameter. The 1000 bins I've obtained from testing on the gnomAD data.

utils/variant_qc.py
lfrancioli: it _should_ be `proband_gt.is_het() & father_gt.is_hom_ref() & mother_gt.is_hom_ref()` ! Thanks for catching this -- typical refactoring fatigue :)

utils/variant_qc.py
lfrancioli: I'm not sure.. I thought about this for a while and I think maybe it's best to keep it there for now and take it out if needed? But happy to take it out too if you want.

utils/variant_qc.py
lfrancioli: good point, this was silly!

utils/variant_qc.py
jkgoodrich: Sounds good. If there is nothing else that will use it then it makes sense to keep here

utils/variant_qc.py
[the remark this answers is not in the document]
jkgoodrich: Makes sense, I just didn't know if the logger should just write out what `k` was passed in rather than `10*n_bins`. Doesn't really make a difference for the code at all, just logs what k was actually passed int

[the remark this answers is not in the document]
lfrancioli: ohhhhh duh! Sorry, I should read my own code -- you're absolutely right!

lfrancioli:
Back to you @jkgoodrich ! Thanks a lot for the great review; all the issues are fixed except for that hail 0.2.28 problem, which is obviously on their side.

PR #127, opened 2019-12-06
Automated ploidy cutoffs for karyotype assignment in sample_qc

ch-kr:
I decided to have the code check for X and Y for each case, as that made the most sense to me, but I'm open to any feedback!

### DIFF
--- a/utils/sample_qc.py
+++ b/utils/sample_qc.py
@@ -398,7 +398,6 @@ def get_ploidy_cutoffs(
         )
     )
 
-    # Set cutoffs -- keep cutoffs permissive to account for loss of sex chromosomes
     cutoffs = (
                 [
                     sex_stats["xy"].x.mean + (normal_ploidy_cutoff * sex_stats["xy"].x.stdev),
--- a/utils/sample_qc.py
+++ b/utils/sample_qc.py
@@ -365,6 +365,14 @@ def infer_sex(
 
 def get_sex_expr(
     ht: hl.Table,
+    chr_x_ploidy: hl.expr.NumericExpression,
+    chr_y_ploidy: hl.expr.NumericExpression,
+    f_stat: hl.expr.NumericExpression,
+    f_stat_XX_cutoff: float = -0.2,
+    f_stat_XY_cutoff: float = 0.2,
+    lower_SD_cutoff: int = 5,
+    upper_SD_cutoff: int = 6,
+    Y_lower_bound: int = 2,
 ) -> hl.expr.StructExpression:
     """
     Creates a struct with the following annotations:
@@ -372,45 +380,63 @@ def get_sex_expr(
     - sex (str), which can be either 'male', 'female', 'ambiguous_sex', or 'sex_aneuploidy'
     - is_female (missing for karyotypes that aren't either 'XX' or 'XY')
 
-    :param Table ht: Table with samples, their imputed sex, and their sex chromosome ploidies
+    :param Table ht: Table with samples, their f-stats, and their sex chromosome ploidies
+    :param NumericExpression chr_x_ploidy: chrom X ploidy (or relative ploidy)
+    :param NumericExpression  chr_y_ploidy: chrom Y ploidy (or relative ploidy)
+    :param NumericExpression f_stat: chrom X f-stat
+    :param float f_stat_XX_cutoff: Min f-stat for XX samples
+    :param flat f_stat_XY_cutoff: Max f-stat for XY samples
+    :param int lower_SD_cutoff: Number of standard deviations to use when determining upper limit of sex chromosome ploidy cutoffs
+    :param int upper_SD_cutoff: Permissive number of standard deviations to use when determining lower ploidy limit and aneuploidy cutoffs
+    :param int Y_lower_bound: Number of standard deviations to use as the lower bound for Y, specifically to correct for somatic loss of Y
     :rtype: StructExpression
     """
-    def get_ploidy_cutoffs(ht: hl.Table, sex: str) -> dict:
+    def get_ploidy_cutoffs(ht: hl.Table, sex: str, f_stat_cutoff: float, lower_SD_cutoff: int, upper_SD_cutoff: int, Y_lower_bound: int) -> dict:
         """
         Gets chromosome X and Y ploidy cutoffs for XY and XX samples
 
-        :param Table ht: Table with samples, their imputed sex, and their sex chromosome ploidies
-        :param str sex: 'XY' or 'XX'
+        :param Table ht: Table with samples and their sex chromosome ploidies
+        :param str sex: Sex for which to check X/Y ploidies. Assumes input is either XX or XY.
+        :param float f_stat_cutoff: f-stat cutoff value. Assumes XX values will be above cutoff and XY values will be below cutoff.
+        :param int lower_SD_cutoff: Number of standard deviations to use when determining upper limit of sex chromosome ploidy cutoffs
+        :param int upper_SD_cutoff: Permissive number of standard deviations to use when determining lower ploidy limit and aneuploidy cutoffs
+        :param int Y_lower_bound: Number of standard deviations to use as the lower bound for Y, specifically to correct for somatic loss of Y
         :return: Dictionary of ploidy cutoffs; key: sex chrom, values: [lower cut, upper cut, outlier (aneuploidy) cut]
         :rtype: dict
         """
-        if sex == "XY":
-            ht = ht.filter((hl.is_defined(ht.is_female)) & (~ht.is_female))
+        if sex == "XX":
+            ht = ht.filter(ht.f_stat < f_stat_cutoff)
         else:
-            ht = ht.filter(ht.is_female)
+            ht = ht.filter(ht.f_stat > f_stat_cutoff)
 
         # Get mean/stdev for chrX/Y ploidies
         X_stats = ht.aggregate(hl.agg.stats(ht.chrX_ploidy))
         Y_stats = ht.aggregate(hl.agg.stats(ht.chrY_ploidy))
 
         # Set cutoffs -- keep cutoffs permissive to account for loss of sex chromosomes
-        # NOTE: keep lower bound of Y particularly permissive (0.05) to save males with somatic loss of Y 
         cutoffs = {}
-        cutoffs["X"] = [X_stats.mean - (5 * X_stats.stdev), X_stats.mean + (5 * X_stats.stdev), X_stats.mean + (6 * X_stats.stdev)]
-        cutoffs["Y"] = [min(0.05, Y_stats.mean - (6 * Y_stats.stdev)), Y_stats.mean + (5 * Y_stats.stdev), Y_stats.mean + (6 * Y_stats.stdev)]
+        cutoffs["X"] = [
+                        X_stats.mean - (upper_SD_cutoff * X_stats.stdev), X_stats.mean + (lower_SD_cutoff * X_stats.stdev), 
+                        X_stats.mean + (upper_SD_cutoff * X_stats.stdev)
+                        ]
+        cutoffs["Y"] = [
+                        abs(min((Y_lower_bound * Y_stats.stdev), Y_stats.mean - (upper_SD_cutoff * Y_stats.stdev))), 
+                        Y_stats.mean + (lower_SD_cutoff * Y_stats.stdev), 
+                        Y_stats.mean + (upper_SD_cutoff * Y_stats.stdev)
+                        ]
         return cutoffs
    
  
-    xx_cutoffs = get_ploidy_cutoffs(ht, "XX")
-    xy_cutoffs = get_ploidy_cutoffs(ht, "XY")
+    xx_cutoffs = get_ploidy_cutoffs(ht, "XX", f_stat_XX_cutoff, lower_SD_cutoff, upper_SD_cutoff, Y_lower_bound)
+    xy_cutoffs = get_ploidy_cutoffs(ht, "XY", f_stat_XY_cutoff, lower_SD_cutoff, upper_SD_cutoff, Y_lower_bound)
 
     xx_x_ploidy_cutoffs = xx_cutoffs["X"][:2]
-    xx_y_ploidy_cutoff = xx_cutoffs["Y"][2]
     xxx_x_ploidy_cutoff = xx_cutoffs["X"][2]
+    xx_y_ploidy_cutoff = xx_cutoffs["Y"][1]
     logger.info('XX cutoffs:')
     logger.info(f'X {xx_x_ploidy_cutoffs}')
-    logger.info(f'Y {xx_y_ploidy_cutoff}')
     logger.info(f'XXX {xxx_x_ploidy_cutoff}')
+    logger.info(f'Y {xx_y_ploidy_cutoff}')
 
     xy_x_ploidy_cutoff = xy_cutoffs["X"][1]
     xy_y_ploidy_cutoffs = xy_cutoffs["Y"][:2]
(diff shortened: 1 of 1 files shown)


utils/sample_qc.py
```diff
@@ -364,91 +364,109 @@ def infer_sex(
 
 
 def get_sex_expr(
-    chr_x_ploidy: hl.expr.NumericExpression,
```
lfrancioli: why remove these two expression params?

utils/sample_qc.py
```diff
@@ -364,91 +364,109 @@ def infer_sex(
...
+    def get_ploidy_cutoffs(ht: hl.Table, sex: str) -> dict:
+        """
+        Gets chromosome X and Y ploidy cutoffs for XY and XX samples
+
+        :param Table ht: Table with samples, their imputed sex, and their sex chromosome ploidies
+        :param str sex: 'XY' or 'XX'
+        :return: Dictionary of ploidy cutoffs; key: sex chrom, values: [lower cut, upper cut, outlier (aneuploidy) cut]
+        :rtype: dict
+        """
+        if sex == "XY":
```
lfrancioli: since the goal of this function is to create `is_female`, it's odd that it's using it -- how did you compute this in the first place? 

A reasonable place to start would be to use `f_stat` to bootstrap this. If so, I would revive the `f_stat` related args.

lfrancioli: Happy to use  something entirely different from `f-stat` btw if you have suggestions

utils/sample_qc.py
```diff
@@ -364,91 +364,109 @@ def infer_sex(
...
-    :param Tuple[float, float] xy_y_ploidy_cutoffs: Boundaries around the chom Y ploidy for males
-    :param float yy_y_ploidy_cutoff: Min chrom Y ploidy for YY
-    :param float xxx_x_ploidy_cutoff: Min chrom X ploidy for XXX
-    :param float f_stat_female_cutoff: Min f-stat for females
-    :param flat f_stat_male_cutoff: Max f-stat for males
-    :return: Struct expression with sex anotations
+    :param Table ht: Table with samples, their imputed sex, and their sex chromosome ploidies
     :rtype: StructExpression
     """
+    def get_ploidy_cutoffs(ht: hl.Table, sex: str) -> dict:
```
lfrancioli: Although this is clearly internal, I think having the number of SDs as args here could make sense

utils/sample_qc.py
```diff
@@ -364,91 +364,109 @@ def infer_sex(
...
+
+        # Get mean/stdev for chrX/Y ploidies
+        X_stats = ht.aggregate(hl.agg.stats(ht.chrX_ploidy))
+        Y_stats = ht.aggregate(hl.agg.stats(ht.chrY_ploidy))
+
+        # Set cutoffs -- keep cutoffs permissive to account for loss of sex chromosomes
+        # NOTE: keep lower bound of Y particularly permissive (0.05) to save males with somatic loss of Y 
+        cutoffs = {}
+        cutoffs["X"] = [X_stats.mean - (5 * X_stats.stdev), X_stats.mean + (5 * X_stats.stdev), X_stats.mean + (6 * X_stats.stdev)]
+        cutoffs["Y"] = [min(0.05, Y_stats.mean - (6 * Y_stats.stdev)), Y_stats.mean + (5 * Y_stats.stdev), Y_stats.mean + (6 * Y_stats.stdev)]
```
lfrancioli: I don't think we should have a `min(0.05...` here --  this makes some assumption about 0.05 being a reasonable cutoff. My preference would be something like 2*n_SDs (or more if necessary).

utils/sample_qc.py
lfrancioli: Also, we may actually want to promote these as args to the main function, so that if these are too lax/stringent in a future callset they'd be easy to adjust

utils/sample_qc.py
lfrancioli: I also think that we possibly  need to allow for somatic loss of X in females BTW

utils/sample_qc.py
```diff
@@ -364,91 +364,109 @@ def infer_sex(
...
-                (chr_x_ploidy > xy_x_ploidy_cutoffs[0]) &
-                (chr_x_ploidy < xy_x_ploidy_cutoffs[1]) &
-                (f_stat > f_stat_male_cutoff),
-                'X0')
-                .default('Ambiguous')
-        )
+                ((ht.chrX_ploidy >= xxx_x_ploidy_cutoff) &
+                (ht.chrY_ploidy < xx_y_ploidy_cutoff)),
+                'XXX') 
+                .default('Ambiguous') # NOTE: XXXY, XXXYY, XYYY, XXXX, XXXXY, XXXXX also exist
```
lfrancioli: We could rename this to `Other` if you think it fits better?

utils/sample_qc.py
```diff
@@ -364,91 +364,109 @@ def infer_sex(
...
+    logger.info('XY cutoffs:')
+    logger.info(f'X {xy_x_ploidy_cutoff}')
+    logger.info(f'Y {xy_y_ploidy_cutoffs}')
+    logger.info(f'YY {yy_y_ploidy_cutoff}')
 
     sex_expr = hl.struct(
         sex_karyotype=(
             hl.case()
                 .when(
-                (chr_x_ploidy > xx_x_ploidy_cutoffs[0]) &
```
lfrancioli: I wonder if this could be restructured as something like
````
hl.case().when(..., 'X').when(..., 'XX').when(..., 'XXX') +
hl.case().when(..., 'Y').when(..., 'YY')
```
instead of having each and every X, Y possibility spelled out.

utils/sample_qc.py
ch-kr: I changed the code to take in a ht (annotated with ploidy and with the output of hail's `impute_sex`) for the `get_ploidy_cutoffs` subfunction, so I removed the extra parameters

utils/sample_qc.py
ch-kr: The `is_female` is output by hail's `impute_sex`. I could change this to check based on f-stat and revive the parameters instead of passing in the full table

utils/sample_qc.py
ch-kr: great idea!

utils/sample_qc.py
ch-kr: I like ambiguous, since samples that fall just outside the cutoffs will fall into this default bucket too, but happy to change this to any other terms

utils/sample_qc.py
lfrancioli: ah yeah... hmm why?
I think that returning expressions is a better / more modular way of structuring code.

utils/sample_qc.py
lfrancioli: yes, I think we should go back to using expressions and then make sure that the cutoff selection is well documented

utils/sample_qc.py
lfrancioli: Ambiguous is fine by me -- this was just about the note you added

ch-kr:
thanks for the review, Laurent -- I made a few more changes based on your comments. let me know what you think!

utils/sample_qc.py
```diff
@@ -364,82 +364,139 @@ def infer_sex(
...
+        cutoffs["Y"] = [
+                        abs(min((Y_lower_bound * Y_stats.stdev), Y_stats.mean - (upper_SD_cutoff * Y_stats.stdev))), 
+                        Y_stats.mean + (lower_SD_cutoff * Y_stats.stdev), 
+                        Y_stats.mean + (upper_SD_cutoff * Y_stats.stdev)
+                        ]
+        return cutoffs
+   
+ 
+    xx_cutoffs = get_ploidy_cutoffs(ht, "XX", f_stat_XX_cutoff, lower_SD_cutoff, upper_SD_cutoff, Y_lower_bound)
+    xy_cutoffs = get_ploidy_cutoffs(ht, "XY", f_stat_XY_cutoff, lower_SD_cutoff, upper_SD_cutoff, Y_lower_bound)
```
lfrancioli: Add argue that here, it would make more sense  to get  the `X` and `Y` cutoffs for the different ploidies in this function rather than the cutoffs for males vs females. The arguments would be very much the same but then what you'd get would be an array where the index is the ploidy for that sex chrom and the value the (lower, upper) cutoffs.
So something like
```
x_cutoffs = get_ploidy_cutoffs(ht, "X", f_stat_XX_cutoff, f_stat_XY_cutoff, lower_SD_cutoff, upper_SD_cutoff)
```
which would return something like this for a perfect 20x cohort (so 10x depth per copy of any contig), with a cutoff set to an mean + SD value of 1,  2 and 3 for 1, 2 and 3 X copies resp. 
```
[(0, 0), (9, 11), (18, 22), (27, 33)]
```
Note that the first element could also be `NA` since I believe that no copy of X is a karyotype that's not compatible with life.

Similarly for Y, it would look like:
```
```
y_cutoffs = get_ploidy_cutoffs(ht, "Y", f_stat_XX_cutoff, f_stat_XY_cutoff, lower_SD_cutoff, upper_SD_cutoff, Y_lower_bound)
```
Which  would look like:
```
[(0, 1), (7, 11), ()]
```

I  think this is conceptually more pleasing than estimating cutoffs for both chroms based on male female and then have another layer of mapping the male/female to copy number cutoffs.

File: utils/sample_qc.py [hunk 8]
```diff
@@ -364,82 +364,139 @@ def infer_sex(
...
+
+    xx_x_ploidy_cutoffs = xx_cutoffs["X"][:2]
+    xxx_x_ploidy_cutoff = xx_cutoffs["X"][2]
```
lfrancioli: As far as  I can tell, this is never set. (only two values are set for `cutoffs['X']`in `get_ploidy_cutoffs`

utils/sample_qc.py
```diff
@@ -364,82 +364,139 @@ def infer_sex(
...
-                'X0')
-                .default('Ambiguous')
+                     (ht.chrY_ploidy >= yy_y_ploidy_cutoff),
+                    'YY')
+                .or_missing()
+            )
+    )
+
+    sex_expr = sex_expr.annotate(
+        sex_karyotype=(hl.case()
```
lfrancioli: I wonder if this expression could be simplified if the expressions above used `default('')` instead or `or_missing`. Seems like we could simply do `X_karyotype` +  `Y_karyotype`. Though  you'd need a special case to replace it with `Ambiguous` if it's `len` == 0. Also, this would return `X` rather than `X0`...which I wonder if it is a problem or not -- maybe something to double-check?

utils/sample_qc.py
```diff
@@ -364,82 +364,139 @@ def infer_sex(
...
-                (f_stat > f_stat_male_cutoff),
-                'X0')
-                .default('Ambiguous')
+                     (ht.chrY_ploidy >= yy_y_ploidy_cutoff),
+                    'YY')
+                .or_missing()
+            )
+    )
+
+    sex_expr = sex_expr.annotate(
```
lfrancioli: I think it's missing a `return` here --right?

utils/sample_qc.py
```diff
@@ -364,82 +364,139 @@ def infer_sex(
...
-    xy_y_ploidy_cutoffs: Tuple[float, float] = (0.15, 1.2),
-    yy_y_ploidy_cutoff: float = 1.3,
-    xxx_x_ploidy_cutoff:float = 2.5,
-    f_stat_female_cutoff: float = -0.2,
-    f_stat_male_cutoff: float = 0.2
+    f_stat_XX_cutoff: float = -0.2,
+    f_stat_XY_cutoff: float = 0.2,
+    lower_SD_cutoff: int = 5,
+    upper_SD_cutoff: int = 6,
+    Y_lower_bound: int = 2,
```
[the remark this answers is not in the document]
ch-kr: I misinterpreted your comment earlier for the Y lower bound being 2*SDs (which ended up being a reasonable value for this dataset). I've changed the Y_lower_bound to 7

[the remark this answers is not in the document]
ch-kr: Actually, scratch that. I changed the code per your suggestion below ("Add argue that here, it would make more sense to get the X and Y cutoffs for the different ploidies in this function rather than the cutoffs for males vs females") and redefined the ploidy cutoffs. The lower bound of Y coverage is now the upper bound of Y coverage in samples categorized roughly as "XX" by f-stat

utils/sample_qc.py
```diff
@@ -364,82 +364,139 @@ def infer_sex(
...
-    :return: Struct expression with sex anotations
+    :param NumericExpression f_stat: chrom X f-stat
+    :param float f_stat_XX_cutoff: Min f-stat for XX samples
+    :param flat f_stat_XY_cutoff: Max f-stat for XY samples
+    :param int lower_SD_cutoff: Number of standard deviations to use when determining upper limit of sex chromosome ploidy cutoffs
+    :param int upper_SD_cutoff: Permissive number of standard deviations to use when determining lower ploidy limit and aneuploidy cutoffs
+    :param int Y_lower_bound: Number of standard deviations to use as the lower bound for Y, specifically to correct for somatic loss of Y
     :rtype: StructExpression
     """
+    def get_ploidy_cutoffs(ht: hl.Table, sex: str, f_stat_cutoff: float, lower_SD_cutoff: int, upper_SD_cutoff: int, Y_lower_bound: int) -> dict:
```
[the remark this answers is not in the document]
ch-kr: I agree the field names might not exist in the ht, but I think I need to pass the ht in for this function; maybe it would be easiest to sit down and talk about this?

utils/sample_qc.py
ch-kr: I implemented something similar to this suggestion, where the cutoffs for X are:
```
[ (single X ploidy lower cutoff, single X ploidy upper cutoff), (double X ploidy lower cutoff, double X ploidy upper cutoff), triple X ploidy lower cutoff]
```
and the cutoffs for Y are:
```
[single Y ploidy lower cutoff, single Y ploidy upper cutoff, double Y ploidy lower cutoff]
```
This made a little more sense to me, but I'm happy to discuss and rework this

ch-kr:
thanks for the review -- I tried some new changes, let me know what you think! happy to also sit down and chat through this sometime instead in case that's easier

utils/sample_qc.py
```diff
@@ -364,98 +364,109 @@ def infer_sex(
...
-                .or_missing()
-        ),
-        is_female=(
+        sex_karyotype=(
             hl.case()
-                .when(sex_expr.sex_karyotype == 'XX', True)
-                .when(sex_expr.sex_karyotype == 'XY', False)
-                .or_missing()
-        )
+                .when((hl.len(sex_expr.X_karyotype) < 4) & (hl.len(sex_expr.Y_karyotype) < 3),
```
lfrancioli: Is the only condition being tested here that both `X_karyotype` and `Y_karyotype` are  not `Ambiguous`  ?
Maybe consider something like:
```
sex_karyotype=hl.cond(
    (sex_expr.X_karyotype == 'Ambiguous') |  (sex_expr.X_karyotype == 'Ambiguous'),
    'Ambiguous',
    sex_expr.X_karyotype + sex_expr.Y_karyotype 
)
```

utils/sample_qc.py
```diff
@@ -364,82 +364,139 @@ def infer_sex(
 
 
 def get_sex_expr(
+    ht: hl.Table,
```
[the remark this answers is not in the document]
ch-kr: that's amazing -- I didn't know you could do that! I'll switch to that instead of passing in the table

utils/sample_qc.py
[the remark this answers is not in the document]
ch-kr: I added a note about field name assumptions for the ht. As the code is now, `impute_sex_ploidy` should return `chrX_ploidy` and `chrY_ploidy`

utils/sample_qc.py
```diff
@@ -364,98 +364,109 @@ def infer_sex(
...
-    xx_x_ploidy_cutoffs: Tuple[float, float] = (1.4, 2.25),
-    xy_x_ploidy_cutoffs: Tuple[float, float] = (0.5, 1.4),
-    xx_y_ploidy_cutoff: float = 0.1,
-    xy_y_ploidy_cutoffs: Tuple[float, float] = (0.15, 1.2),
-    yy_y_ploidy_cutoff: float = 1.3,
-    xxx_x_ploidy_cutoff:float = 2.5,
-    f_stat_female_cutoff: float = -0.2,
-    f_stat_male_cutoff: float = 0.2
+    f_stat_cutoff: float = 0.5,
+    lower_SD_cutoff: int = 5,
```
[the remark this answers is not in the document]
ch-kr: Great idea, I'll do that now.

utils/sample_qc.py
```diff
@@ -364,98 +364,109 @@ def infer_sex(
...
+
+        # Set cutoffs -- keep cutoffs permissive to account for loss of sex chromosomes
+        cutoffs = {}
+        cutoffs["X"] = [
+                        (xy_x_stats.mean - (upper_SD_cutoff * xy_x_stats.stdev), xy_x_stats.mean + (lower_SD_cutoff * xy_x_stats.stdev)),
+                        (xx_x_stats.mean - (lower_SD_cutoff * xx_x_stats.stdev), xx_x_stats.mean + (lower_SD_cutoff * xx_x_stats.stdev)),
+                        xx_x_stats.mean + (upper_SD_cutoff * xx_x_stats.stdev)
+                        ]
+        cutoffs["Y"] = [
+                        abs(xx_y_stats.mean - (upper_SD_cutoff * xx_y_stats.stdev)),
```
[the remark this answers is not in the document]
ch-kr: I added this `abs` because i wanted to use the upper cutoff of Y ploidy in XX samples as the lower cutoff for a single copy of Y (any samples above this cutoff will get marked as at least having one copy of Y). Unfortunately, I realized when testing the code that Y ploidy distribution in the XX samples was so tight that this value was negative (and in this case, `max(0, ...)` wouldn't work, as every sample would get called as having at least one copy of Y). Happy to change this, let me know what you think!

utils/sample_qc.py
```diff
@@ -364,98 +364,109 @@ def infer_sex(
...
+
+        :param Table ht: Table with samples and their sex chromosome ploidies
+        :param str sex: Sex for which to check X/Y ploidies. Assumes input is either XX or XY.
+        :param float f_stat_cutoff: f-stat cutoff value. Assumes XX values will be below cutoff and XY values will be above cutoff.
+        :param int lower_SD_cutoff: Number of standard deviations to use when determining upper limit of sex chromosome ploidy cutoffs
+        :param int upper_SD_cutoff: Permissive number of standard deviations to use when determining lower ploidy limit and aneuploidy cutoffs
+        :return: Dictionary of ploidy cutoffs; key: sex chrom, values: [lower cut, upper cut, outlier (aneuploidy) cut]
+        :rtype: dict
+        """
+        xx_ht = ht.filter(ht.f_stat < f_stat_cutoff)
```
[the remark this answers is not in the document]
ch-kr: this is a great idea -- much cleaner!

ch-kr:
thanks for the suggestions -- the code looks much tidier now! I think we're getting close, let me know what you think about the lower cutoff for single copy of Y (https://github.com/macarthur-lab/gnomad_hail/pull/127#discussion_r366894932)

ch-kr:
I updated the cutoffs and unnested `get_ploidy_cutoffs`, as we discussed. Let me know if I've missed anything!

utils/sample_qc.py
```diff
@@ -363,99 +363,114 @@ def infer_sex(
...
+    ht: hl.Table, 
+    f_stat_cutoff: float, 
+    normal_ploidy_cutoff: int = 5, 
+    aneuploidy_cutoff: int = 6
+) -> tuple:
+    """
+    Gets chromosome X and Y ploidy cutoffs for XY and XX samples. Note this assumes the input hail Table has the fields f_stat, chrX_ploidy, and chrY_ploidy.
+    Returns a tuple of sex chromosome ploidy cutoffs: ([x_ploidy_cutoffs], [y_ploidy_cutoffs]).
+    x_ploidy_cutoffs: [upper cutoff for single X, (lower cutoff for double X, upper cutoff for double X), lower cutoff for triple X]
+    y_ploidy_cutoffs: [lower cutoff for single Y, upper cutoff for single Y, lower cutoff for double Y]
```
lfrancioli: what do you think of using `((lower cutoff for single Y, upper cutoff for single Y), lower cutoff for double Y)` for consistency' sake? so that similar to X each element of the returned value corresponds to a specific ploidy?
Otherwise, I think we should change `x_ploidy_cutoffs` to: `(upper cutoff for single X, (lower cutoff for double X, upper cutoff for double X), lower cutoff for triple X)`

utils/sample_qc.py
```diff
@@ -363,99 +363,114 @@ def infer_sex(
...
+            hl.cond(ht.f_stat < f_stat_cutoff, "xx", "xy"),
+            hl.struct(
+                    x=hl.agg.stats(ht.chrX_ploidy),
+                    y=hl.agg.stats(ht.chrY_ploidy)
+            )
+        )
+    )
+
+    cutoffs = (
+                [
```
lfrancioli: let's use tuples here ( rather than lists for the return type so that we can fully describe the return type since the type of the encapsulated values is mixed.

utils/sample_qc.py
```diff
@@ -363,99 +363,114 @@ def infer_sex(
     return (sex_ht)
 
 
+def get_ploidy_cutoffs(
+    ht: hl.Table, 
+    f_stat_cutoff: float, 
+    normal_ploidy_cutoff: int = 5, 
+    aneuploidy_cutoff: int = 6
+) -> tuple:
```
lfrancioli: `tuple` won't work, we need the typing class `Tuple`. If you  agree with my  proposals, then the full return type would be:
`Tuple[Tuple[float, Tuple[float, float], float], Tuple[Tuple[float, float], float]]`

utils/sample_qc.py
```diff
@@ -363,99 +363,114 @@ def infer_sex(
...
-    f_stat: hl.expr.NumericExpression,
-    xx_x_ploidy_cutoffs: Tuple[float, float] = (1.4, 2.25),
-    xy_x_ploidy_cutoffs: Tuple[float, float] = (0.5, 1.4),
-    xx_y_ploidy_cutoff: float = 0.1,
-    xy_y_ploidy_cutoffs: Tuple[float, float] = (0.15, 1.2),
-    yy_y_ploidy_cutoff: float = 1.3,
-    xxx_x_ploidy_cutoff:float = 2.5,
-    f_stat_female_cutoff: float = -0.2,
-    f_stat_male_cutoff: float = 0.2
+    x_ploidy_cutoffs: list,
```
lfrancioli: same as `tuple` up  there, these need to be typing classes (so `list`- > `List`). That said, if  you agree with my proposed changes, then it should be `Tuple[float, Tuple[float, float], float]` for `x_ploidy_cutoffs` and `Tuple[Tuple[float, float], float]` for `y_ploidy_cutoffs`

ch-kr:
Thanks for those suggestions -- I added the edits and an additional set of logging statements (just to send the XX/XY stats and X/Y cutoffs to stdout)

PR #138, opened 2020-01-23
Added rep_on_read function to generic.py

ch-kr:


utils/generic.py
```diff
@@ -1534,4 +1534,18 @@ def interval_length(interval: hl.Interval) -> int:
                 interval.end.position
         )
     else:
-        return interval.end.position - interval.start.position
+        return interval.end.position - interval.start.positio
```
nawatts: Looks like the "n" in position got accidentally deleted.

utils/generic.py
```diff
@@ -1534,4 +1534,18 @@ def interval_length(interval: hl.Interval) -> int:
...
+
+
+def rep_on_read(path: str, n_partitions: int) -> hl.MatrixTable:
+    """
+    Reparitions a MatrixTable on read
```
nawatts: "Repartitions" is missing a "t".

utils/generic.py
```diff
@@ -1535,3 +1535,17 @@ def interval_length(interval: hl.Interval) -> int:
         )
     else:
         return interval.end.position - interval.start.position
+
+
+def rep_on_read(path: str, n_partitions: int) -> hl.MatrixTable:
+    """
+    Repartitions a MatrixTable on read
```
lfrancioli: It should be made clear when to use this function as opposed to `read` and `repartition` sequentially.

PR #139, opened 2020-01-24
Add job to run Pylint on pushes to master and PRs

nawatts:
Following on from #133, this adds a check to automatically run [Pylint](https://www.pylint.org/) on pushes to master and PRs.

This runs pylint with warning messages disabled. Refactor and convention messages are also disabled by the .pylintrc configuration. Thus, only fatal or error messages (syntax errors, undefined variables, etc) will prevent a PR from passing this status check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,29 @@ on:
       - master
   pull_request:
 jobs:
+  pylint:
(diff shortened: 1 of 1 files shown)


nawatts:
This check will fail (likely until #123 is merged) since there are some problems in the master branch now.

PR #141, opened 2020-01-24
Add link to API reference in README

nawatts:
Refer people to the API reference pages instead of the code to browse what's available.

Also increases the depth of the table of contents on that page to make it more useful.

### DIFF
--- a/docs/generate_api_reference.py
+++ b/docs/generate_api_reference.py
@@ -91,7 +91,7 @@ def write_module_doc(module_name):
 PACKAGE_DOC_TEMPLATE = """{title}
 
 .. toctree::
-    :maxdepth: 1
+    :maxdepth: 2
 
     {module_links}
 """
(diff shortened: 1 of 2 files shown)


docs/generate_api_reference.py
```diff
@@ -91,7 +91,7 @@ def write_module_doc(module_name):
 PACKAGE_DOC_TEMPLATE = """{title}
 
 .. toctree::
-    :maxdepth: 1
+    :maxdepth: 2
```
gtiao: I'm not familiar with this -- why is maxdepth now 2?

nawatts: This controls the depth of the table of contents tree. Currently, with maxdepth = 1, the API reference page displays a list of links with two items:

* resources
* utils

Increasing maxdepth to 2 also lists out the modules under those, which makes it a lot easier to quickly scan through what's available.

* resources
   * grch37
   * grch38
   * resource_utils
* utils
   * annotations
   * constants
   * generic
   ...

gtiao: Sounds reasonable to me!

PR #143, opened 2020-01-28
Fixup docstring formatting

nawatts:
Fix docstring formatting. For example, without a space between the summary line and parameter list, Sphinx won't format the parameter list.

Also (annoyingly) Sphinx can't parse multi-word types in parameter descriptions. So `:param list of str foo:` has to be replaced with:
```
:param foo:
:type foo: list of str
```

Some time, I'll look into whether it's possibly to pull the parameter types from the type annotations instead of the docstring.

### DIFF
--- a/gnomad_hail/utils/generic.py
+++ b/gnomad_hail/utils/generic.py
@@ -91,8 +91,9 @@ def get_reference_genome(
     """
     Returns the reference genome associated with the input Locus expression
 
-    :param LocusExpression or IntervalExpression locus: Input locus
-    :param  bool add_sequence: If set, the fasta sequence is added to the reference genome
+    :param locus: Input locus
+    :type locus: LocusExpression or IntervalExpression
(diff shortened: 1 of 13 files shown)


gnomad_hail/utils/generic.py
```diff
@@ -238,12 +246,14 @@ def filter_low_conf_regions(mt: Union[hl.MatrixTable, hl.Table], filter_lcr: boo
     """
     Filters low-confidence regions
 
-    :param MatrixTable or Table mt: MatrixTable or Table to filter
+    :param mt: MatrixTable or Table to filter
+    :param mt: MatrixTable or Table
```
lfrancioli: should be `type`

nawatts: Fixed in bfa3b1a55893f55ae8c532f4c04890b68e2a005e

PR #145, opened 2020-01-28
Updated init script now that gnomad_hail lib is in a subdir of the repo

lfrancioli:


lfrancioli:
yeah, it's not uber-clean.It's just that the PWD of the notebook is set to `/home/hail`, so if we don't want to deal with updating `PYTHONPATH` this works

PR #148, opened 2020-01-29
Link Hail types to Hail documentation

nawatts:
This adds the [Intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) extension to automatically link references to Hail types in parameter and return type documentation to Hail's documentation.


docs/conf.py
```diff
@@ -65,5 +67,50 @@ def extract_summary(doc, document):
...
+# For example, Table is documented as hail.Table, not hail.table.Table.
+# Thus, intersphinx cannot link some of the Hail references generated by
+# sphinx_autodoc_typehints.
+#
+# To work around this, override sphinx_autodoc_typehints's get_annotation_module
+# function and map the qualified names to what Hail's documentation uses.
+original_get_annotation_module = sphinx_autodoc_typehints.get_annotation_module
+
+
+hail_module_map = {
```
nawatts: There might be a more general way of mapping these (like map anything starting with "hail.expr" to "hail.expr") but this covers everything that is currently referenced in gnomad_hail documentation.

PR #149, opened 2020-01-29
Generalize variant qc

jkgoodrich:
Hey Laurent, So I think I made the changes we discussed and tested the pipeline on the UKBB VQSR data. I have not yet updated you v3 gnomAD QC files to work with the changes. 

A couple things: for default_score_bin_agg - Some of these annotations are a bit different from v2 and famstats is no longer an annotation on the fam_ht, so do you think we should add generate_fam_stats and some of your other annotation functions into to an annotations file in gnomad_hail if the strata you used are standards that we will want? It currently has a densify step, so that would somehow need to be made generalizable. 

I also need to modify the plotting code, which I think you did for the older versions of these functions, but I need to look through them to see what else might need to be changed.


gnomad_hail/utils/variant_qc.py
```diff
@@ -263,80 +275,481 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
+        bin_expr_no_snv = {bin_id.rsplit("_", 1)[0] for bin_id in bin_ht.bin_stats}
+        bin_ht = bin_ht.annotate_globals(
+            bin_stats=hl.struct(
+                **{
+                    bin_id: hl.struct(
+                        **{snv: bin_ht.bin_stats[f"{bin_id}_{snv}"] for snv in ["snv", "indel"]}
+                    )
+                    for bin_id in bin_expr_no_snv
+                }
+            )
```
jkgoodrich: I had to add this because a later step in the pipeline gets screwed up because it uses bin_stats

gnomad_hail/utils/variant_qc.py
```diff
@@ -188,38 +196,43 @@ def _is_dnm(
...
-        The `rank_expr` defines which data the rank(s) should be computed on. E.g., to get an SNV rank and an Indel rank,
-        the following could be used:
-        rank_expr={
-        'snv_rank': hl.is_snp(ht.alleles[0], ht.alleles[1]),
-        'indels_rank': ~hl.is_snp(ht.alleles[0], ht.alleles[1])
+        The `bin_expr` defines which data the bin(s) should be computed on. E.g., to get an SNV quantile bin and an Indel
+        quantile bin, the following could be used:
+        bin_expr={
+       'snv_bin': hl.is_snp(ht.alleles[0], ht.alleles[1]),
+       'indels_bin': ~hl.is_snp(ht.alleles[0], ht.alleles[1])
```
nawatts: These two lines need to be indented the same as the others in this section.

gnomad_hail/utils/variant_qc.py
```diff
@@ -188,38 +196,43 @@ def _is_dnm(
...
-    :return: Table with the binned ranks
+    :param Table ht: Input Table
+    :param NumericExpression score_expr: Expression containing the score
+    :param dict of str -> BooleanExpression bin_expr: Quantile bin(s) to be computed (see notes)
+    :param bool stratify_snv_indel: Should all `bin_expr` items be stratified by snv / indels
+    :param int n_bins: Number of bins to bin the data into
+    :param int k: The `k` parameter of approx_quantiles
+    :param bool desc: Whether to bin the score in descending order
+    :return: Table with the quantile bins
+    :rtype: Table
```
nawatts: Parameter types and return types no longer need to be specified in docstrings.

gnomad_hail/utils/variant_qc.py
```diff
@@ -262,80 +275,481 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
+    - adj singletons
+    - adj biallelic singletons
+
+    :param Table ht: Input table
+    :param int n_bins: Number of bins to bin into
+    :param bool singleton: Should bins be stratified by singletons
+    :param bool biallelic: Should bins be stratified by bi-alleleic variants
+    :param bool adj: Should bins be stratified by adj filtering
+    :param dictionary of boolean expr keyed by str add_substrat: Any additional stratifications for adding bins
+    :return table with bin number for each variant
```
nawatts: `:return` needs to have a colon after it (`:return:`) to be interpreted correctly.

gnomad_hail/utils/variant_qc.py
```diff
@@ -262,80 +275,481 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
+        `n_vqsr_pos_train` - count of variants that were a VQSR positive train site per group
+        `n_vqsr_neg_train` - count of variants that were a VQSR negative train site per group
+
+    Requires that `bin_ht` be annotated with an `info` struct that includes QD, FS, and MQ in order to add an
+    annotation for `fail_hard_filters`
+
+    :param Table bin_ht: Input Table with a `bin_id` annotation
+    :param callable agg_func: Function that returns a dict of any additional aggregations to perform
+    :param str checkpoint_path: If provided an intermediate checkpoint table is created with all required annotations before shuffling.
+    :return Table grouped by rank(s) and with counts of QC metrics
```
nawatts: `:return` needs to have a colon after it (`:return:`) to be interpreted correctly.

gnomad_hail/utils/variant_qc.py
```diff
@@ -262,80 +275,481 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
+def default_score_bin_agg(
+        ht: hl.Table,
+        truth_ht: hl.Table,
+        fam_stats_ht: hl.Table
+) -> Dict[str, hl.expr.Aggregation]:
+    """
+    Default aggregation function to pass to `compute_aggregate_binned_data` to add aggregations for number of ClinVar
+    variants, number of truth variants (omni, mills, hapmap, and kgp_phase1), and family statistics.
+
+    Note that the following fields should be present:
```
nawatts: I don't really understand why, but Sphinx insists on a blank line between this line and "In ht:".

gnomad_hail/utils/variant_qc.py
```diff
@@ -262,80 +274,476 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
+        bin_expr=bin_expr,
+        n_bins=n_bins
+    )
+    ht = ht.join(bin_ht, how='left')
+
+    return ht
+
+
+def default_score_bin_agg(
+        ht: hl.Table,
```
lfrancioli: This should be of type GroupedTable

gnomad_hail/utils/variant_qc.py
```diff
@@ -262,80 +274,476 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
+        )
+    )
+
+
+def compute_binned_truth_sample_concordance(
+        ht: hl.Table,
+        binned_score_ht: hl.Table
+) -> hl.Table:
+    """
+    The input HT should contain two row fields:
```
lfrancioli: Can you add a little high-level comment on what this does?

gnomad_hail/utils/variant_qc.py
```diff
@@ -262,80 +274,476 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
+    return (
+        bin_ht.group_by(
+            bin_id=bin_ht.bin_id,
+            contig=bin_ht.locus.contig,
+            snv=hl.is_snp(bin_ht.alleles[0], bin_ht.alleles[1]),
+            bi_allelic=~bin_ht.was_split,
+            singleton=bin_ht.singleton,
+            release_adj=bin_ht.ac > 0,
+            bin=bin_ht.bin
+        )._set_buffer_size(20000)
```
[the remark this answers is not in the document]
jkgoodrich: I know we discussed this yesterday and it sounded great, but I am struggling to implement this since when you return a GroupedTable it doesn't link back to the annotations that were on the table prior to grouping so I get an error. should we just put it all in one function like before and then name it `default_compute_grouped_binned_ht` or `default_compute_aggregate_binned_data`?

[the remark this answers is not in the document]
lfrancioli: Let's take this to Slack -- might be easier

gnomad_hail/utils/variant_qc.py
```diff
@@ -541,64 +591,53 @@ def default_score_bin_agg(
         else grch38_resources.reference_data.get_truth_ht()
     )[ht.key]
     fam = fam_stats_ht[ht.key]
-
+    fam.describe()
```
lfrancioli: looks like a debugging stmt :)

PR #150, opened 2020-01-29
Add setup.py

nawatts:
This lays the groundwork for publishing on PyPI.

* https://packaging.python.org/tutorials/packaging-projects/#creating-setup-py
* https://packaging.python.org/guides/distributing-packages-using-setuptools/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 hail
 hdbscan
 ipywidgets
-sklearn
+scikit-learn
 slackclient==2.0.0
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,34 @@
+import setuptools
+
+
+with open("README.md", "r") as readme_file:
+    long_description = readme_file.read()
+
+with open("requirements.txt", "r") as requirements_file:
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 hail
 hdbscan
 ipywidgets
-sklearn
```
nawatts: The [sklearn project page](https://pypi.org/project/sklearn/) says "Use scikit-learn instead".

konradjk: We use `sklearn` in the code though - either we have to change that and make sure it works 1:1, or revert this for now and do that later. I leave it up to you

nawatts: Only the [distribution package](https://packaging.python.org/glossary/#term-distribution-package) has changed. The [import package](https://packaging.python.org/glossary/#term-import-package) is still `sklearn`.

https://scikit-learn.org/stable/getting_started.html

The only import from scikit-learn is `from sklearn.ensemble import RandomForestClassifier`, which works after pip installing `scikit-learn` instead of `sklearn`.

PR #153, opened 2020-02-03
Added get_thruth_ht to grch38 resources

lfrancioli:


lfrancioli:
@jkgoodrich I agree this is quite useful! Should be a quick one!

PR #154, opened 2020-02-03
Added remaining gnomad v3 public resources

lfrancioli:
- Relative -> absolute imports for reference data imports to ease running them
- Removed a few stray commas

### DIFF
--- a/gnomad_hail/resources/grch38/reference_data.py
+++ b/gnomad_hail/resources/grch38/reference_data.py
@@ -33,7 +33,7 @@ syndip = MatrixTableResource(
 )
 
 syndip_hc_intervals = TableResource(
-    path='gs://gnomad-public/truth-sets/source/HG001_GRCh38_GIAB_highconf_CG-IllFB-IllGATKHC-Ion-10X-SOLID_CHROM1-X_v.3.3.2_highconf_nosomaticdel_noCENorHET7_hc_regions.ht',
+    path='gs://gnomad-public/truth-sets/source/syndip_b38_hc_regions.ht',
     import_sources={
         "source_path": 'gs://gnomad-public/truth-sets/source/syndip.b38.bed'
     }
--- a/gnomad_hail/resources/grch38/reference_data.py
+++ b/gnomad_hail/resources/grch38/reference_data.py
@@ -11,14 +11,32 @@ purcell_5k = TableResource(
     path="gs://gnomad-public/resources/grch38/purcell5k.ht",
     import_sources={
         "source_path": "gs://gnomad-public/intervals/purcell5k.interval_list",
-    },
+    }
 )
 
 na12878_giab = MatrixTableResource(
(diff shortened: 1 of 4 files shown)


gnomad_hail/resources/grch38/reference_data_import.py
```diff
@@ -55,6 +57,19 @@ def main(args):
...
+            reference_genome='GRCh38',
+            skip_invalid_intervals=True
+        ).write(na12878_giab_hc_intervals.path, overwrite=args.overwrite)
+
+    if args.syndip:
+        hl.import_bed(
+            syndip_hc_intervals.import_sources['source_path'],
+            reference_genome='GRCh38',
+            skip_invalid_intervals=True
+        ).write(na12878_giab_hc_intervals.path, overwrite=args.overwrite)
```
mike-w-wilson: This be `syndip_hc_intervals.path`

gnomad_hail/resources/grch38/reference_data.py
```diff
@@ -11,14 +11,32 @@
     path="gs://gnomad-public/resources/grch38/purcell5k.ht",
     import_sources={
         "source_path": "gs://gnomad-public/intervals/purcell5k.interval_list",
-    },
+    }
```
nawatts: I think @mike-w-wilson was using [Black](https://black.readthedocs.io/en/stable/) and it put these commas here.

It would be nice if we agreed on a style guide (or better yet, all ran Black 😁) to avoid stylistic churn like this.

lfrancioli: We've never left stray commas like this anywhere else... I'm not necessarily opposed to it but it should be applied all over if we do

mike-w-wilson: Yup! I did run Black on this. Definitely all for agreeing and adhering to a style guide - not sure if this fits into our recent standardization meetings but maybe we could approach it there.

lfrancioli: Sounds like a good idea. For now, I'll merge this since anyways the entire lib will need to be formatted once we agree on a format.

nawatts: Agreed that consistency throughout is preferable. A style guide would be one way to achieve that, but usually ends up with lots of nitpick comments in PR reviews. Alternatively, a formatter applies the style guide automatically.

Also, trailing commas are useful to reduce noise in diffs.
https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas

nawatts: > I'll merge this since anyways the entire lib will need to be formatted once we agree on a format

👍 Certainly did not mean for that comment to be taken as blocking.

mattsolo1: Having used formatters in JavaScript (and Black a little bit), I can say this is a game-changer! It's disconcerting at first, but you'll progress to a point where you can't live without it 😂.

Funny thread from JS community on this subject: https://twitter.com/dan_abramov/status/1161998154138603525?lang=en

PR #155, opened 2020-02-04
Addition of LazyTableResource for resource that are imported at runtime

lfrancioli:
- A few minor non-breaking changes to how the `expected_file_extension` argument is used

### DIFF
--- a/gnomad_hail/resources/resource_utils.py
+++ b/gnomad_hail/resources/resource_utils.py
@@ -33,7 +33,12 @@ class BaseResource(ABC):
 
         if path is not None and expected_file_extensions and not [ext for ext in expected_file_extensions if path.endswith(ext)]:
             logger.warning(
-                f"Created the following {self.__class__.__name__} with a path that doesn't ends with {expected_file_extension}: {self}")
+                "Created the following {} with a path that doesn't ends with {}: {}".format(
+                    self.__class__.__name__,
+                    " or ".join(expected_file_extensions),
+                    self
+                )
+            )
 
     def __repr__(self):
         attr_str = [f'path={self.path}']
--- a/gnomad_hail/resources/resource_utils.py
+++ b/gnomad_hail/resources/resource_utils.py
@@ -1,8 +1,14 @@
 from typing import Optional, Dict, Any
+from typing_extensions import Protocol
 from gnomad_hail.utils.gnomad_functions import logger
 import hail as hl
 from hail.linalg import BlockMatrix
 from abc import ABC, abstractmethod
+
+
+# Protocol class for LazyTable Callable with kwargs
+class LazyTableImportFuncCallable(Protocol):
+    def __call__(self, path: str, **kwargs) -> hl.Table: ...
 
 
 # Resource classes
@@ -12,7 +18,7 @@ class BaseResource(ABC):
     """
 
     @abstractmethod
-    def __init__(self, path: str, import_sources: Optional[Dict[str, Any]] = None, expected_file_extension: str = ""):
+    def __init__(self, path: str, import_sources: Optional[Dict[str, Any]] = None, expected_file_extension: Optional[str] = None):
         """
         Creates a Resource
 
@@ -23,7 +29,7 @@ class BaseResource(ABC):
         self.path = path
         self.import_sources = import_sources
 
-        if expected_file_extension and not path.endswith(expected_file_extension):
+        if expected_file_extension is not None and not path.endswith(expected_file_extension):
             logger.warning(
                 f"Created the following {self.__class__.__name__} with a path that doesn't ends with {expected_file_extension}: {self}")
 
@@ -39,11 +45,11 @@ class TableResource(BaseResource):
     A Hail Table resource
     """
 
-    def __init__(self, path: str, import_sources: Optional[Dict[str, Any]] = None, ):
+    def __init__(self, path: str, import_sources: Optional[Dict[str, Any]] = None, expected_file_extension: Optional[str] = '.ht'):
         super().__init__(
             path=path,
             import_sources=import_sources,
-            expected_file_extension='.ht'
+            expected_file_extension=expected_file_extension
         )
 
     def ht(self) -> hl.Table:
@@ -53,6 +59,38 @@ class TableResource(BaseResource):
         :return: Hail Table resource
         """
         return hl.read_table(self.path)
+
+
+class LazyTableResource(TableResource):
+    """
+    A Hail Table resource that imports the table from source when .ht() is called for the first time.
+    """
+
+    _ht = None
+
+    def __init__(
+            self,
+            path: str,
+            import_func: LazyTableImportFuncCallable,
+            import_args: Optional[Dict[str, Any]] = None,
+    ):
+        super().__init__(
+            path=path,
+            import_sources=None,
+            expected_file_extension=None
+        )
+        self.import_func=import_func
+        self.import_args = import_args
+
+    def ht(self) -> hl.Table:
+        """
+        Imports and return the HT
+
+        :return: Hail Table resource
+        """
+        if self._ht is None:
+            self._ht = self.import_func(path=self.path, **self.import_args)
+        return self._ht
 
 
 class MatrixTableResource(BaseResource):
--- a/gnomad_hail/resources/resource_utils.py
+++ b/gnomad_hail/resources/resource_utils.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict, Any, Callable
+from typing import Optional, Dict, Any, Callable, List
 from gnomad_hail.utils.gnomad_functions import logger
 import hail as hl
 from hail.linalg import BlockMatrix
@@ -9,6 +9,11 @@ from abc import ABC, abstractmethod
 class BaseResource(ABC):
     """
     Generic abstract resource class.
+
+    :param path: The resource path
+    :param import_sources: Any sources that are required for the import and need to be kept track of (e.g. .vcf path for an imported VCF)
+    :param import_func: A function used to import the resource. `import_func` will be passed the `import_sources` dictionary as kwargs. `import_func` will be passed the `import_sources` dictionary as kwargs.
+    :param expected_file_extensions: A list of all expected file extensions. If path doesn't end with one of these, a warning if emitted.
     """
 
     @abstractmethod
@@ -16,19 +21,17 @@ class BaseResource(ABC):
             self,
             path: Optional[str] = None,
             import_sources: Optional[Dict[str, Any]] = None,
-            expected_file_extension: Optional[str] = None
+            import_func: Optional[Callable] = None,
+            expected_file_extensions: Optional[List[str]] = None
     ):
-        """
-        Creates a Resource
-
-        :param path: The resource path
-        :param import_sources: Additional attributes for the resource
-        """
+        if path is None and import_func is None:
+            raise ValueError(f"{self.__class__.__name__} requires at least one of path or import_func arguments.")
 
         self.path = path
         self.import_sources = import_sources
+        self.import_func = import_func
 
-        if path is not None and expected_file_extension is not None and not path.endswith(expected_file_extension):
+        if path is not None and expected_file_extensions and not [ext for ext in expected_file_extensions if path.endswith(ext)]:
             logger.warning(
                 f"Created the following {self.__class__.__name__} with a path that doesn't ends with {expected_file_extension}: {self}")
 
@@ -38,30 +41,35 @@ class BaseResource(ABC):
             attr_str.append(f'import_sources={self.import_sources}')
         return f'{self.__class__.__name__}({",".join(attr_str)})'
 
+    @abstractmethod
+    def import_resource(self):
+        """
+        Abstract method to import the resource using its import_func and writes it in its path.
+        """
+        pass
+
 
 class TableResource(BaseResource):
     """
     A Hail Table resource
+
+    :param path: The Table path (typically ending in .ht)
+    :param import_sources: Any sources that are required for the import and need to be kept track of and/or passed to the import_func (e.g. .vcf path for an imported VCF)
+    :param import_func: A function used to import the Table. `import_func` will be passed the `import_sources` dictionary as kwargs.
     """
 
     def __init__(
             self,
             path: Optional[str] = None,
             import_sources: Optional[Dict[str, Any]] = None,
-            import_func: Optional[Callable[..., hl.Table]] = None,
-            expected_file_extension: Optional[str] = '.ht'
+            import_func: Optional[Callable[..., hl.Table]] = None
     ):
-
-        if path is None and import_func is None:
-            raise ValueError("TableResource requires one of path or import_func arguments.")
-
         super().__init__(
             path=path,
             import_sources=import_sources,
-            expected_file_extension=expected_file_extension
+            import_func=import_func,
+            expected_file_extensions=['.ht']
         )
-
-        self.import_func =import_func
 
     def ht(self, force_import: bool = False) -> hl.Table:
         """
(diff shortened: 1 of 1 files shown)


lfrancioli:
I do not love `Lazy` here but neither Mike nor I could find anything better. @nawatts would be great if you want to take a look at this PR and/or suggest a better name!

lfrancioli:
The use-case for this is that we have a bunch of resources that are quite small that we typically do not want to necessarily import and write as HT as it is overkill. Yet it'd be great to include them in a streamlined interface.

nawatts:
Alternatively, we could merge this into the base TableResource: add the `import_func` parameter to TableResource and default it to `hl.read_table(self.path)`.

lfrancioli:
hmmm, maybe I'll drop the `Protocol` thing. Although it looks like the right way to do this seems like `typing_extensions` needs installation, which might be overkill

lfrancioli:
digesting that last suggestion...

lfrancioli:
> Alternatively, we could merge this into the base TableResource: add the `import_func` parameter to TableResource and default it to `hl.read_table(self.path)`.

Yes, I think I like this after all

nawatts:
```python
def ht():
    if not self._ht:
        if self.import_function:
            self._ht = self.import_function(self.import_sources)
        else:
            self._ht = hl.read_table(self.path)

    return self._ht
```

lfrancioli:
we need to think what to do with `path` though

lfrancioli:
Maybe we could even store the import functionalities for all resources in there then instead of in https://macarthur-lab.github.io/gnomad_hail/api_reference/resources/grch38/reference_data_import.html

lfrancioli:
and `ht()` would do read if resources exists, else import

lfrancioli:
OK @nawatts , @mike-w-wilson , I've folded the import logic into TableResource. Take a look and comment. When we agree I'll apply the same logic to MatrixTableResource.

A note is that after some consideration I chose not to `try` `read_table` and default to import otherwise because the Hail always returns a `HailError` when reading doesn't work and it sometimes contains useful information. I also thought of checking if the file exists, and importing if not, but this also causes troubles because it's hard to make it independent of the storage used.

lfrancioli:
I added a tentative `force_import` argument that could be used if a user wanted to re-import a resource.
I suppose another avenue would be an argument that specifies what to do in case of a read error.

nawatts:
> A note is that after some consideration I chose not to try read_table and default to import otherwise because the Hail always returns a HailError when reading doesn't work and it sometimes contains useful information.

👍If a consumer wants to import on failure, then they have the option of catching the error from from `ht` and calling it again with `force_import=True`.

> I also thought of checking if the file exists, and importing if not, but this also causes troubles because it's hard to make it independent of the storage used.

👍Checking if files exist at `self.path` doesn't provide any guarantee anyway. The table files could still be deleted between the check and the read or while the read is in progress. Also, expecting the read to work is more [idiomatic Python](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/).

> When we agree I'll apply the same logic to MatrixTableResource.

I think the changes to `__init__` to add `import_func` should be applied to the base resource class instead of duplicated in TableResource, MatrixTableResource, etc.

lfrancioli:
@nawatts @mike-w-wilson I've just pushed a suggestion for import_resource for all resource types. Would be great to have feedback. Once we settle on an implementation, I'll update the import scripts.

gnomad_hail/resources/resource_utils.py
```diff
@@ -9,114 +9,221 @@
...
 
         self.path = path
         self.import_sources = import_sources
+        self.import_func = import_func
 
-        if expected_file_extension and not path.endswith(expected_file_extension):
+        if path is not None and expected_file_extensions and not [ext for ext in expected_file_extensions if path.endswith(ext)]:
             logger.warning(
-                f"Created the following {self.__class__.__name__} with a path that doesn't ends with {expected_file_extension}: {self}")
+                "Created the following {} with a path that doesn't ends with {}: {}".format(
```
nawatts: "doesn't ends with" => "doesn't end with"

gnomad_hail/resources/resource_utils.py
```diff
@@ -9,114 +9,221 @@
...
 
         :return: Hail Table resource
         """
-        return hl.read_table(self.path)
+        if self.path is None or force_import:
+            return self.import_func(**self.import_sources)
+        else:
+            return hl.read_table(self.path)
+
+    def import_resource(self, overwrite: bool = False, **kwargs):
```
nawatts: If arbitrary kwargs can be passed to `Table.write`, is there any reason to separate `overwrite` from other kwargs? Same question for MatrixTableResource.

gnomad_hail/resources/resource_utils.py
```diff
@@ -9,114 +9,221 @@
...
 class PedigreeResource(BaseResource):
     """
     A pedigree resource
+
+    :param path: The Pedigree path (typically ending in .fam or .ped)
+    :param import_sources: Any sources that are required for the import and need to be kept track of and/or passed to the import_func (e.g. .vcf path for an imported VCF)
+    :param import_func: A function used to import the MatrixTable. `import_func` will be passed the `import_sources` dictionary as kwargs.
+    :param quant_pheno: If ``True``, phenotype is interpreted as quantitative.
+    :param delimiter: Field delimiter regex.
+    :param missing: The string used to denote missing values. For case-control, 0, -9, and non-numeric are also treated as missing.
```
nawatts: These new options don't really seem like they're in scope of this PR. @mike-w-wilson can you take a look at these?

gnomad_hail/resources/resource_utils.py
```diff
@@ -9,114 +9,221 @@
...
+
+    def ht(self) -> hl.Table:
         """
         Reads the pedigree into a family HT using hl.import_fam().
 
-        :param delimiter: Delimiter used in the ped file
         :return: Family table
         """
-        return hl.import_fam(self.path, delimiter=delimiter)
+        return hl.import_fam(self.path, quant_pheno=self.quant_pheno, delimiter=self.delimiter, missing=self.missing)
```
nawatts: The `ht` method on this class seems out of place. Other resource classes have _one_ method to get the Hail object they're associated with (`ht` for TableResource, `mt` for MatrixTableResource, `bm` for BlockMatrixResource). Why does PedigreeResource have another method in addition to `pedigree`? And it requires adding to the class two more attributes (`quant_pheno` and `missing`) that are ignored by the primary method `pedigree`.

gnomad_hail/resources/resource_utils.py
```diff
@@ -9,114 +9,221 @@
...
         """
         Reads the pedigree into an hl.Pedigree using hl.Pedigree.read().
 
         :param delimiter: Delimiter used in the ped file
         :return: pedigree
         """
-        return hl.Pedigree.read(self.path, delimiter=delimiter)
+        return hl.Pedigree.read(self.path, delimiter=self.delimiter)
+
+    def import_resource(self):
```
nawatts: Since the `import_resource` method on most resource classes' (Table, MatrixTable, and BlockMatrix) has an overwrite option and passes kwargs to the underlying Hail write function, I would expect this to do the same.

gnomad_hail/resources/resource_utils.py
```diff
@@ -127,23 +234,29 @@ def bm(self) -> BlockMatrix:
...
     Abstract class for a versioned resource
-    """
 
-    def __init__(self, default_version: str, versions: Dict[str, BaseResource]):
-        """
-        Creates a versioned resource.
-        The `path`/`source_path` attributes of the versioned resource are those
-        of the default version of the resource.
-        In addition, all versions of the resource are stored in the `versions` attribute.
+    The `path`/`source_path` attributes of the versioned resource are those of the default version of the resource.
```
nawatts: `source_path` should probably be replaced with `import_sources` here. `source_path` is one of the keys in `import_sources` for most resources, but it is not an attribute of resource objects. Same comment for docstrings on other versioned resource classes.

gnomad_hail/resources/resource_utils.py
```diff
@@ -127,23 +234,29 @@ def bm(self) -> BlockMatrix:
...
+    In addition, all versions of the resource are stored in the `versions` attribute.
 
-        :param default_version: The default version of this resource (needs to be in the `versions` dict)
-        :param versions: A dict of version name -> resource.
-        """
+    :param default_version: The default version of this resource (must to be in the `versions` dict)
```
nawatts: "must to be in" => "must be in"

gnomad_hail/resources/resource_utils.py
lfrancioli: yes, `overwrirte` is used all the time, the other args essentially never. I still thought that adding this pass-over functionality could potentially be helpful.

gnomad_hail/resources/resource_utils.py
lfrancioli: I think these were overlooked before -- they are IMHO helpful to have here as they contain key information regarding the resource that should be easily accessible.

gnomad_hail/resources/resource_utils.py
lfrancioli: Pedigree information is weird like this in that sometimes we use it in a python context (hence the `.pedigree()` function and sometimes in a hail annotation context that requires a `ht`. In my opinion bundling everything about the resource within the class is the option that makes the most sense.

gnomad_hail/resources/resource_utils.py
lfrancioli: `Pedigree.write` takes no arguments. It WILL overwrite (I'll add this as a comment and we _could_ add an `overwrite` arg with default `True` (which would be different from other resources) and throw a `NotImplementedError` if it is set to `False`. We could also add the `kwargs` to future-proof things although atm no argument is supported.

gnomad_hail/resources/resource_utils.py
lfrancioli: yes! That's definitely a remnant of a previous implementation iteration that wasn't caught earlier!

lfrancioli:
Thanks @nawatts for the comments! I'll let @mike-w-wilson weigh in on the `overwrite` and `Pedigree` discussion.

gnomad_hail/resources/resource_utils.py
nawatts: Since `import_resource` is a method on the base resource class, it should have the same signature for all resource classes. If you have a collection of resources, you should be able to iterate over them and call `import_resource` without worrying about the details of its implementation in each subclass. For PedigreeResource, I think `overwrite` should default False the same as the other classes and raise (or warn) if it's not set to True.

With the current implementation, if we move most of the code in the two reference data import scripts ([GRCh37](https://github.com/macarthur-lab/gnomad_hail/blob/923fec00302cb06c1b90508b20fe6926c3b25591/gnomad_hail/resources/grch37/reference_data_import.py) and [GRCh38](https://github.com/macarthur-lab/gnomad_hail/blob/923fec00302cb06c1b90508b20fe6926c3b25591/gnomad_hail/resources/grch38/reference_data_import.py)) into the resource definitions using `import_func`, we can't do this:
```python
all_resources = [
   purcell_5k_intervals,
   clinvar,
   na12878_giab,
]

for resource in all_resources:
  resource.import_resource(overwrite=True)
```
because PedigreeResources would have to be special cased.

gnomad_hail/resources/resource_utils.py
nawatts: I kind of want to move these to `import_sources` (though the "sources" name doesn't quite fit) since they're related to importing data from a non-Hail native file format.

gnomad_hail/resources/resource_utils.py
mike-w-wilson: I haven't worked with pedigrees much in hail but agree they seem a bit out of place since we do not define any of the other parameters for hail functions for other resources.

gnomad_hail/resources/resource_utils.py
mike-w-wilson: We sort of do this with importing the dbsnp header file? I know it is an actual file so a source but also an extra parameter in the hail function?

nawatts: Maybe we should rename `import_sources` to something more generic like `import_info`?

mike-w-wilson: That works for me

lfrancioli: we could rename `import_sources` to `import_info`, that said these aren't about importing but are pieces of information about the resource which only pertain to this specific resource type.

lfrancioli: actually, I was thinking of `quant_pheno` -- I agree that the rest _is_ about importing.

gnomad_hail/resources/resource_utils.py
mike-w-wilson: If we add a dumby `overwrite` to pedigree resources couldnt we do what you mention above @nawatts ? Since the pedigree will always overwrite?

gnomad_hail/resources/resource_utils.py
nawatts: They're arguments to **import**_fam. There's no requirement or assumption on what keys should be in `import_sources`/`import_info` for different resource types or even individual resources. I think the fact that `import_info` is per-individual resource is a better fit here since different individual pedigree resources are used differently (with `.ht()` or with `.pedigree()`).

lfrancioli: so here I think the weird bit is - as Nick pointed out - the fact that we store this information in plain text as there is no Hail-native format (hail's `Pedigree.write` just writes a text file). For this reason, `import` here as a bit of a different meaning in that it'd really be a function that converts some other format into...well a text format (fam file). 
Now when we use the resource, the fam file is read and unfortunately at that point additional args are needed. I truly believe that storing these with the resource will provide the best user experience and don't see these args as `import_info`.

gnomad_hail/resources/resource_utils.py
nawatts: Yes, we'd need an `overwrite` argument here (doesn't really matter if it's used or not) to avoid a `TypeError: import_resource() got an unexpected keyword argument 'overwrite'` in that example.

lfrancioli: > Since `import_resource` is a method on the base resource class, it should have the same signature in the base class and in all resource subclasses. If you have a collection of resources, you should be able to iterate over them and call `import_resource` without worrying about the details of its implementation in each subclass. For PedigreeResource, I think `overwrite` should default False the same as the other classes and raise (or warn) if it's not set to True.
> 
> With the current implementation, if we move most of the code in the two reference data import scripts ([GRCh37](https://github.com/macarthur-lab/gnomad_hail/blob/923fec00302cb06c1b90508b20fe6926c3b25591/gnomad_hail/resources/grch37/reference_data_import.py) and [GRCh38](https://github.com/macarthur-lab/gnomad_hail/blob/923fec00302cb06c1b90508b20fe6926c3b25591/gnomad_hail/resources/grch38/reference_data_import.py)) into the resource definitions using `import_func`, we can't do this:
> 
> ```python
> all_resources = [
>    purcell_5k_intervals,
>    clinvar,
>    na12878_giab,
> ]
> 
> for resource in all_resources:
>   resource.import_resource(overwrite=True)
> ```
> 
> because PedigreeResources would have to be special cased.

Haha, I've actually written that already :)

lfrancioli: I'm fine to add the dummy `overwrite` -- now, I think it needs to be set to `True` by default here

lfrancioli: For other resources, I feel like keeping the hail default of `overwrite=False` would have made sense but now we're facing a pickle. Maybe no default?

mike-w-wilson: With no default set, we'd be using hails default which is `False` right?

lfrancioli: No, we'd force the user to provide a value

lfrancioli: I guess a last option, which was suggested by @nawatts, is to only use `kwargs` and pass these to the corresponding hail write function.

mike-w-wilson:
Thanks for adding all the class docstrings. They are super clear.

PR #156, opened 2020-02-05
Added function to annotate relationship

ch-kr:


gnomad_hail/utils/sample_qc.py
```diff
@@ -659,6 +659,47 @@ def run_pca_with_relateds(
         return pca_evals, pca_scores, pca_loadings
 
 
+def annotate_relationship(
+    relatedness_ht: hl.Table,
+    first_degree_threshold: List[float] = [0.1767767, 0.4],
```
bw2: to continue the streak of only posting naming comments, it's slightly confusing that `first_degree_threshold` is a range and `second_degree_threshold` is a number. Maybe this could be two args like ..._upper_threshold, _lower_threshold.

nawatts: Also, this is definitely a nitpick, but since the threshold argument is a fixed size (2 elements), `Tuple[float, float]` might be a more appropriate type for it.

gnomad_hail/utils/sample_qc.py
```diff
@@ -659,6 +659,47 @@ def run_pca_with_relateds(
...
+            & (relatedness_ht.kin < first_degree_threshold[1])
+            & (relatedness_ht.ibd2 >= ibd2_parent_offspring_threshold),
+            "Full-sibling",
+        )
+        .when(
+            (relatedness_ht.kin > first_degree_threshold[0])
+            & (relatedness_ht.kin < first_degree_threshold[1])
+            & (relatedness_ht.ibd2 < ibd2_parent_offspring_threshold),
+            "Parent-child",
+        )
```
bw2: Is it worth handling the case where relatedness_ht.kin >= first_degree_threshold[1] ?

ch-kr: yes, definitely!

PR #158, opened 2020-02-06
New import resources

lfrancioli:


lfrancioli:
Looking for feedback, but I think this may be a nice way of keeping code minimal for the resources as it'll remove the need for adding resources to other import script individually.

lfrancioli:
I plan to also modify resouces to implement `import_func` as part of this PR.

lfrancioli:
I've now moved grch38 resources to the new framework proposal. I renamed `import_sources` to `import_args` (seems clearest but happy to consider alternatives, like `import_info`). 
Also, not too happy with having `NO_CHR_TO_CHR_CONTIG_RECODING` and `import_sites_vcf` in `resource_utils.py` -- seems like the right place based on the filename but odd since the rest of this is all about the resources classes.

gnomad_hail/resources/resource_utils.py
```diff
@@ -323,7 +323,7 @@ class VersionedPedigreeResource(BaseVersionedResource, PedigreeResource):
     """
     Versioned Pedigree resource
 
-    The attributes (path, import_sources and import_fun of the versioned resource are those of the default version of the resource.
+    The attributes (path, import_args and import_fun of the versioned resource are those of the default version of the resource.
```
nawatts: import_fun => import_func

Also, the other versioned resource classes still refer to source_path instead of import_args.

gnomad_hail/resources/import_resources.py
```diff
@@ -0,0 +1,82 @@
...
+    .. code-block:: python
+
+        import gnomad_hail.resources.grch37 as grch37
+        grch37_resources = get_module_importable_resources(grch37, prefix='grch37')
+    
+    :param module: Input module
+    :param prefix: 
+    :return: 
+    """
+    _prefix = f'{prefix}.' if prefix is not None else ''
```
nawatts: Nitpick, but just `if prefix` instead of `if prefix is not None` would be preferable here. Otherwise, passing an empty string for prefix would result in a prefix of ".".

gnomad_hail/resources/import_resources.py
```diff
@@ -0,0 +1,82 @@
...
+    resources = grch37_resources
+    resources.update(grch38_resources)
+    for arg in [arg for arg, arg_value in vars(args).items() if arg_value and arg != 'overwrite']:
+            resource_name, resource = resources[arg]
+            print(f"Importing {resource_name}...")
+            resource.import_resource(args.overwrite)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
nawatts: Rather than adding an optional argument for every possible resource, I think using one positional argument for the list of resources would be preferable. It's a more accurate representation of what the arguments represent: a collection of resource names. And it would simplify the code for both creating and processing the arguments.

Something like:
`parser.add_argument("resources", nargs="+", choices=resources.keys())`

gnomad_hail/resources/import_resources.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+        
+
+grch37_resources = get_module_importable_resources(grch37, 'grch37')
+grch38_resources = get_module_importable_resources(grch38, 'grch38')
+
+
+def main(args):
+    resources = grch37_resources
+    resources.update(grch38_resources)
```
nawatts: It would be preferable to create a copy of the GRCh37 resources dictionary and add GRCh38 resources to that copy. Though it's unlikely, if some code called this `main` function, the contents of `grch37_resources` wouldn't be accurate after that.

gnomad_hail/resources/import_resources.py
```diff
@@ -0,0 +1,83 @@
+from inspect import getmembers
+from typing import Dict, Tuple, Optional
+from gnomad_hail.resources.resource_utils import BaseResource, BaseVersionedResource
+import gnomad_hail.resources.grch37 as grch37
+import gnomad_hail.resources.grch38 as grch38
+import argparse
+
+
+# Generate a dictionary of resource available for import for a given genome build
+def get_module_importable_resources(module, prefix: Optional[str] = None) -> Dict[str, Tuple[str, BaseResource]]:
```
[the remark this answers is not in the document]
lfrancioli: I'm envisioning importing this function to import resources in other projects, e.g. gnomad_qc/v3 and applying it to import project-specific resources.

gnomad_hail/resources/grch37/reference_data.py
```diff
@@ -100,90 +149,114 @@
 
 lcr_intervals = TableResource(
     path="gs://gnomad-public/resources/intervals/LCR.GRCh37_compliant.interval_list.ht",
-    import_sources={
-        "source_path": "gs://gnomad-public/intervals/LCR.GRCh37_compliant.interval_list",
+    import_func=hl.import_locus_intervals,
+    import_args={
+        "path": "gs://gnomad-public/intervals/LCR.GRCh37_compliant.interval_list",
```
mike-w-wilson: since we are using `skip_invalid_loci` on vcfs, should we use `skip_invalid_intervals` on bed files/lnterval files?

gnomad_hail/resources/grch38/intervals.py
```diff
@@ -0,0 +1,22 @@
...
+
+telomeres_and_centromeres = TableResource(
+    path='gs://gnomad-public/resources/intervals/hg38.telomeresAndMergedCentromeres.ht',
+    import_func=hl.import_bed,
+    import_args={
+        "path": 'gs://gnomad-public/resources/grch38/hg38.telomeresAndMergedCentromeres.bed',
+        "reference_genome": 'GRCh38',
+        "skip_invalid_intervals": True
+    }
+)
```
mike-w-wilson: I think we should either move these into `reference_data.py` or move all the interval files out and into this module like `syndip_hc_intervals` I'm fine either way but would prefer to be consistent across the modules as to what goes where. If we chose two separate files, I think 37 should be adjusted as well to have a separate `intervals`module.

gnomad_hail/resources/import_resources.py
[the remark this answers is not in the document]
nawatts: Why would we use this function to import resources over `from gnomad_hail.resources import grch37`?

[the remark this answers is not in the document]
lfrancioli: This would be used to create import scripts for project-specific resources. An example is VQSR that is generated for each project as VCF using GATK and then needs importing in hail. But this file isn't public as it contains sites from unreleasable samples. Samples metadata is another example. So by having this generic functions, we can make simple project-specific import scripts that are only a few lines long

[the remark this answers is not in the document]
nawatts: Oh, ok. I think I misunderstood. I thought you were saying use this function to import resources from the GRCh37/38 resources modules here.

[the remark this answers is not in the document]
lfrancioli: Right, I can see how my comment was ambiguous :)

PR #160, opened 2020-02-10
Resource and reference path updates

mike-w-wilson:
This updates the file paths of our resources to a new structure. Both `path` and `source_path` now live under the resource in a specific build subdirectory. For example `gs://gnomad-public/resources/<build>/purcell_5k_intervals/purcell5k.ht`. Resources from the same source, such as `na12878` live in the same subdirectory such as `na12878_giab` and `na12878_giab_hc_intervals` both live `gs://gnomad-public/resources/grch38/na12878/`. 

This also moves intervals back into `reference_data.py` in grch38. I do not have strong feelings on this but both grch37 and grch38 should follow the same structure. Happy to move grch37 interval files out of `reference_data.py` into an intervals module if people feel strongly about it. Once these paths are merged, I will move all source files and recreate the resources using `import_resources.py`


gnomad_hail/resources/grch37/reference_data.py
```diff
@@ -8,54 +8,54 @@
...
+        "path": "gs://gnomad-public/resources/grch37/kgp/1000G_omni2.5.b37.vcf.bgz",
         "force_bgz": True,
         "min_partitions": 100,
         "reference_genome": "GRCh37",
     },
 )
 
 mills = TableResource(
-    path="gs://gnomad-public/truth-sets/hail-0.2/Mills_and_1000G_gold_standard.indels.b37.ht",
+    path="gs:/gnomad-public/resources/grch37/mills/Mills_and_1000G_gold_standard.indels.b37.ht",
```
nawatts: The paths here and in `import_args` are missing a `/` in `gs://`.

mike-w-wilson:
I had to update the date format used in grch37 dbsnp as argparse convert `-` to `_`. I've also removed the grch38 intervals module as I added them back to grch38s reference data module last commit.

nawatts:
> I had to update the date format used in grch37 dbsnp as argparse converts - to _ and the module cannot be found.

@lfrancioli here's one reason for https://github.com/macarthur-lab/gnomad_hail/pull/158#discussion_r375941661.

lfrancioli:
> > I had to update the date format used in grch37 dbsnp as argparse converts - to _ and the module cannot be found.
> 
> @lfrancioli here's one reason for [#158 (comment)](https://github.com/macarthur-lab/gnomad_hail/pull/158#discussion_r375941661).

Yup! Again, I'm not opposed to this and my understanding is that you were going to make a proposal so for the switch!

lfrancioli:
When all the resources have been regenerated in their new location, I'm totally fine for this to be merged.

gnomad_hail/resources/grch38/reference_data.py
```diff
@@ -13,6 +13,9 @@
 
 def _import_purcell_5k(path) -> hl.Table:
     p5k = hl.import_locus_intervals(path, reference_genome='GRCh37')
+    rg37 = hl.get_reference('GRCh37')
+    rg38 = hl.get_reference('GRCh38')
+    rg37.add_liftover('gs://hail-common/references/grch37_to_grch38.over.chain.gz', rg38)
```
nawatts: [Hail docs](https://hail.is/docs/0.2/genetics/hail.genetics.ReferenceGenome.html#hail.genetics.ReferenceGenome.add_liftover) say "This method can only be run once per reference genome." Will this break if we need to use liftover in another import? Maybe this should be extracted into a function like `ensure_liftover` that first checks if the chain file has been registered using `has_liftover`.

nawatts: Or is there a way to copy the reference genome objects, add the liftover to them, and use the copies in the import. That way this wouldn't modify the global reference genomes.

mike-w-wilson: From what I can tell reading the docs, we can't copy them.  I can make an `ensure_liftover` resource util unless someone knows how to copy the objects.

nawatts: It looks like even if a copy was possible, it would also be registered globally as well.

> Note that constructing a new reference genome, either by using the class constructor or by using ReferenceGenome.read() will add the reference genome to the list of known references; it is possible to access the reference genome using get_reference() anytime afterwards.
> 
> -- https://hail.is/docs/0.2/genetics/hail.genetics.ReferenceGenome.html

lfrancioli: Good catch! should we just do 
```
if not rg37.has_liftover:
    rg37.add_liftover('gs://hail-common/references/grch37_to_grch38.over.chain.gz', rg38)
```

I feel like adding a function for this is overkill..

nawatts: That function may already exist in liftover.py:
https://github.com/macarthur-lab/gnomad_hail/blob/ecf5e9cc9b297bfd8a24ca683da62b49e27d71bd/gnomad_hail/utils/liftover.py#L35-L65

cc @ch-kr

nawatts: Always using one function for liftover ensures we're consistent with the chain files. If every function that needs to liftover does `if not has_liftover, add_liftover` and the same chain file isn't used everywhere, then the results of liftover could depend on what order functions are called in.

lfrancioli: Damn, that function in `liftover.py` isn't safe either :(

lfrancioli: I think we need our own reference class wrapper if we don't want an error and we want to have a clean `ensure_liftover` mechanism.

lfrancioli: Otherwise either we let it error out or we assume there is only one possible chain file.

lfrancioli: Thinking about it, I feel like a simple reference singleton with defaults may well be useful

mike-w-wilson: I'm not sure I understand 

> simple reference singleton with defaults

Would this mean adding a new resource class and just returning a reference object with chain files added?

lfrancioli: well, we could have a default chain file that is stored, and when calling `.add_liftover` it would   make sure that they are the same file. 
All in all, we have quite a few places in the library that deal with reference genomes. I think we should address this more globally outside of this PR. But I think it's worth thinking about a good mechanism to handle them.

lfrancioli: So I'd vote to make and issue so that we can keep track of this but not alter the PR

mike-w-wilson: https://github.com/macarthur-lab/gnomad_hail/issues/162

PR #161, opened 2020-02-11
Updated infer_sex function to use sex chromosome ploidies and return …

ch-kr:
…table with sex karyotypes


gnomad_hail/utils/sample_qc.py
```diff
@@ -278,86 +278,55 @@ def assign_platform_from_pcs(
...
-        y_mt: hl.MatrixTable,
-        platform_ht: hl.Table,
-        male_min_f_stat: float,
-        female_max_f_stat: float,
-        min_male_y_sites_called: int = 500,
-        max_y_female_call_rate: float = 0.15,
-        min_y_male_call_rate: float = 0.8
+        mt: hl.MatrixTable,
+        ploidy_ht: hl.Table,
+        f_stat_cutoff: float,
```
lfrancioli: should we have a default here?

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,86 +278,55 @@ def assign_platform_from_pcs(
...
-    :param max_y_female_call_rate:
-    :param min_y_male_call_rate:
+    :param mt: Input MatrixTable
+    :param ploidy_ht: Table with imputed sex chromosome ploidies
+    :param f_stat_cutoff: f-stat to roughly divide 'XX' from 'XY' samples. Assumes XX samples are below cutoff and XY are above cutoff.
+    :param sites_ht: Optional Table to use. If present, filters input MatrixTable to sites in this Table prior to imputing sex.
+    :param aaf_expr: Optional. Name of field in input MatrixTable with alternate allele frequency.
+    :param gt_expr: Name of entry field storing the genotype. Default: 'GT'
+    :param float aaf_threshold: Minimum alternate allele frequency to be used in f-stat calculations.
+    :param float male_threshold: Threshold above which a sample will be called male.
```
lfrancioli: If keeping these arguments, I think it should be made clear that they are cutoffs on the F-stat  (and not ploidy)

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,86 +278,55 @@ def assign_platform_from_pcs(
...
-    :param x_mt:
-    :param y_mt:
-    :param platform_ht:
-    :param male_min_f_stat:
-    :param female_max_f_stat:
-    :param min_male_y_sites_called:
-    :param max_y_female_call_rate:
-    :param min_y_male_call_rate:
+    :param mt: Input MatrixTable
+    :param ploidy_ht: Table with imputed sex chromosome ploidies
```
lfrancioli: Should it be clear that this is a Table of samples?

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,86 +278,55 @@ def assign_platform_from_pcs(
...
-    mean_male_y_sites_called = sex_ht.aggregate(hl.agg.filter(~sex_ht.is_female, hl.agg.group_by(sex_ht.qc_platform, hl.agg.mean(sex_ht.n_y_sites_called))))
-    y_call_rate_stats = sex_ht.aggregate(
-        hl.agg.filter(
-            hl.is_defined(sex_ht.is_female),
-            hl.agg.group_by(hl.tuple([sex_ht.qc_platform, sex_ht.is_female]),
-                            hl.agg.stats(sex_ht.y_call_rate)
-                            )
-        )
-    )
+    logger.info("Filtering mt to chrX")
```
lfrancioli: Maybe we should get the list X-contigs before and print the actual list in the INFO ?

lfrancioli: Otherwise maybe use "x contig(s)" instead?

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,86 +278,55 @@ def assign_platform_from_pcs(
...
+        f_stat_cutoff: float,
+        sites_ht: Optional[hl.Table],
+        aaf_expr: Optional[str] = None,
+        gt_expr: str = 'GT',
+        aaf_threshold: float = 0.001,
+        male_threshold: float = 0.75,
+        female_threshold: float = 0.5,
 ) -> hl.Table:
     """
     Imputes sample sex based on X-chromosome heterozygosity and Y-callrate (if Y calls are present)
```
lfrancioli: Maybe worth specifying exactly what will be annotated?

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,86 +278,55 @@ def assign_platform_from_pcs(
...
+    :param mt: Input MatrixTable
+    :param ploidy_ht: Table with imputed sex chromosome ploidies
+    :param f_stat_cutoff: f-stat to roughly divide 'XX' from 'XY' samples. Assumes XX samples are below cutoff and XY are above cutoff.
+    :param sites_ht: Optional Table to use. If present, filters input MatrixTable to sites in this Table prior to imputing sex.
+    :param aaf_expr: Optional. Name of field in input MatrixTable with alternate allele frequency.
+    :param gt_expr: Name of entry field storing the genotype. Default: 'GT'
+    :param float aaf_threshold: Minimum alternate allele frequency to be used in f-stat calculations.
+    :param float male_threshold: Threshold above which a sample will be called male.
+    :param float female_threshold: Threshold below which a sample will be called female
     :return:
```
lfrancioli: Would be great to add a small blurb about what it returns so the docs are nice a complete.

ch-kr: oops!

ch-kr:
I actually like keeping the ploidy part separate. I imagine that you'd have to stop the pipeline after running impute ploidy to check the ploidy distributions and make sure the default values are separating the sexes correctly. What do you think?

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,85 +278,69 @@ def assign_platform_from_pcs(
...
-    sex_ht = sex_ht.annotate_globals(
-        impute_sex_params=hl.struct(
-            male_min_f_stat=male_min_f_stat,
-            female_max_f_stat=female_max_f_stat,
-            min_male_y_sites_called=min_male_y_sites_called,
-            max_y_female_call_rate=max_y_female_call_rate,
-            min_y_male_call_rate=min_y_male_call_rate
+    Imputes sample sex based on X-chromosome heterozygosity and sex chromosome ploidy. 
+    Returns Table with the following fields:
+    - s (str): Sample
```
lfrancioli: indent so it appears as a list in the docs.

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,85 +278,69 @@ def assign_platform_from_pcs(
...
+
+    logger.info("Calculating inbreeding coefficient on chrX")
+    sex_ht = hl.impute_sex(mt[gt_expr], aaf_threshold=aaf_threshold, male_threshold=f_stat_cutoff, female_threshold=f_stat_cutoff, aaf=aaf_expr)
+
+    logger.info("Annotating sex ht with sex chromosome ploidies")
+    sex_ht = sex_ht.annotate(**ploidy_ht[sex_ht.key])
+
+    logger.info("Inferring sex karyotypes")
+    x_ploidy_cutoffs, y_ploidy_cutoffs = get_ploidy_cutoffs(sex_ht, f_stat_cutoff)
+    return sex_ht.annotate(
```
lfrancioli: use `select` instead of `annotate`/`drop`  ?

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,85 +278,69 @@ def assign_platform_from_pcs(
...
+    logger.info(f"Filtering mt to biallelic SNPs in X contigs: {x_contigs}")
+    if 'was_split' in list(mt.row):
+        mt = mt.filter_rows((~mt.was_split) & hl.is_snp(mt.alleles[0], mt.alleles[1]))
+    else:
+        mt = mt.filter_rows((hl.len(mt.alleles) == 2) & hl.is_snp(mt.alleles[0], mt.alleles[1]))
+    mt = hl.filter_intervals(mt, [hl.parse_locus_interval(contig) for contig in x_contigs])
+
+    if sites_ht:
+        if aaf_expr == None:
+            logger.warn("sites_ht was provided, but aaf_expr is missing. Assuming name of field with alternate allele frequency is 'AF'.")
```
lfrancioli: The docstring doesn't mention that the sites HT is used for AF purpose; add to the docstring?
Also, I'm wondering about the benefit of having `sites_ht` as an argument at all; seems more intuitive for users to filter/annotate to the sites they want outside of the function. Now if bringing the ploidy computation in there may be more of a case to be made as you'd want  all sites for that.

ch-kr:
Good point. I've added the ploidy calculation to this function and thus also kept the sites_ht argument.

gnomad_hail/utils/sample_qc.py
```diff
@@ -278,86 +278,94 @@ def assign_platform_from_pcs(
...
-        min_male_y_sites_called: int = 500,
-        max_y_female_call_rate: float = 0.15,
-        min_y_male_call_rate: float = 0.8
+def default_annotate_sex(
+        mt: hl.MatrixTable,
+        is_sparse: bool = True,
+        excluded_intervals: Optional[hl.Table] = None,
+        included_intervals: Optional[hl.Table] = None,
+        normalization_contig: str = 'chr20',
+        chr_x: Optional[str] = None,
```
lfrancioli: Because it is a _default_ function, maybe remove `chr_x`, `chr_y` as args? They're super unlikely to be used and also there is a bit of inconsistency in using `x_contigs = get_reference_genome(mt.locus).x_contigs` for the 2nd part of the code if the user provided specific weird contigs..

ch-kr: good point! will remove these

PR #163, opened 2020-02-11
Moved famstats and sibling stats to annotations and added a default function from gnomad_v3_qc.

jkgoodrich:
Hey Laurent! So I took a first round at the things we discussed. The big things in this PR are:
-I moved the generate_fam_stats from variant_qc to annotations
-Added your v3 default function (split out into two functions to allow densify in between as discussed) 
-Renamed to trio_stats since that is really what it is doing
-Added sibling variant sharing counts to allow for identifying sibling singletons in a similar way we to the trio calculations
-Added parent and child allele counts to generate_trio_stats_expr


gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +425,503 @@ def annotate_freq(
...
+
+    return ht.filter(
+        ht.n_de_novos_raw + ht.n_transmitted_raw + ht.n_untransmitted_raw > 0
+    )
+
+
+def generate_sib_stats_expr(
+    sib_mt: hl.MatrixTable,
+    strata: Dict[str, hl.expr.BooleanExpression] = {"raw": True},
+    sib1_entry: str = 'sib1_entry',
```
lfrancioli: It's probably something that we need to address more widely, but passing MT + col names vs expr is something we should try and be consistent with. I now tend to like expr better because they can be nested and / or computed from other expr without needing an `annotate`. Anyhow, just a thought for something more global rather than immediate action needed.

gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +425,503 @@ def annotate_freq(
...
+    :param sib_ht: Input sibling pairs table
+    :param i_col: Column containing the 1st sample of the pair in the relationship table
+    :param j_col: Column containing the 2nd sample of the pair in the relationship table
+    :param sex_ht: An optional table containing sex information for the samples. If not given, counts are only computed for autosomes.
+    :param is_female_col: An optional column in sex_ht giving the samples sex. If not given, counts are only computed for autosomes.
+    :return: A Table with the sibling shared variant counts
+    """
+    if is_female_col is None or sex_ht is None:
+        logger.warning(
+            "Since no proband sex expression was given to generate_trip_stats_expr, only DNMs in autosomes will be counted."
```
lfrancioli: seems  like this comment  needs to  be removed or adapted.

gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +425,503 @@ def annotate_freq(
...
+                expr,
+                hl.agg.sum(sib_mt.sib1_entry.GT.n_alt_alleles() + sib_mt.sib2_entry.GT.n_alt_alleles())
+            ) for name, expr in strata.items()
+        }
+    )
+
+    return sib_stats
+
+
+def create_sib_mt(
```
lfrancioli: Is this sibs MT used for any other purpose than creating those sibs counts?  If  not, this seems like an awful lot of work that I don't think is necessary. If you only created the `sibs` global annotation, it seems like that'd be  enough to compute these stats without reorganizing the whole MT.

gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +425,503 @@ def annotate_freq(
...
+    :return: A Table with the sibling shared variant counts
+    """
+    sex_ht = sex_ht.annotate(
+        is_female=hl.case()
+            .when(sex_ht.sex_karyotype == "XX", True)
+            .when(sex_ht.sex_karyotype == "XY", False)
+            .or_missing()
+    )
+
+    sib_ht = relatedness_ht.filter(relatedness_ht[relationship_col] == 'Siblings')
```
lfrancioli: change to constant from `relatedness.py`. Also, maybe add in the docstring that it's what this will be using?

lfrancioli: I just realized that this is in another PR! So I guess maybe add a comment to change it?

gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +425,431 @@ def annotate_freq(
...
+    sex_ht = sex_ht.annotate(
+        is_female=hl.case()
+            .when(sex_ht.sex_karyotype == "XX", True)
+            .when(sex_ht.sex_karyotype == "XY", False)
+            .or_missing()
+    )
+
+    # TODO: Change to use SIBLINGS constant when relatedness PR goes in
+    sib_ht = relatedness_ht.filter(relatedness_ht[relationship_col] == 'Siblings')
+    s_to_keep = set(sib_ht[i_col].s.collect()) | set(sib_ht[j_col].s.collect())
```
lfrancioli: would be better written as:
```
s_to_keep = sibs_ht.aggregate(hl.agg.explode(lambda s: hl.agg.collect_as_set(s), [sibs_ht[i_col].s, sibs_ht[j_col].s], _localize=False)
mt = mt.filter_cols(s_to_keep).contains(mt.s)
```

PR #164, opened 2020-02-12
add vep to clinvar import

mike-w-wilson:
Added a vep step to the clinvar import per Beryl's suggestion.

### DIFF
--- a/gnomad_hail/resources/grch38/reference_data.py
+++ b/gnomad_hail/resources/grch38/reference_data.py
@@ -6,7 +6,7 @@ from gnomad_hail.resources.resource_utils import (
     import_sites_vcf,
     NO_CHR_TO_CHR_CONTIG_RECODING
 )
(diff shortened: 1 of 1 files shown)


gnomad_hail/resources/grch38/reference_data.py
```diff
@@ -33,6 +34,12 @@ def _import_purcell_5k(path) -> hl.Table:
     return p5k.key_by('locus')
 
 
+def _import_clinvar(**kwargs) -> hl.Table:
+    clinvar = import_sites_vcf(**kwargs)
+    clinvar = hl.vep(clinvar, vep_config_path(ref='GRCh38'))
```
nawatts: Should we use `gnomad_hail.utils.generic.vep_or_lookup_vep` for this?

mike-w-wilson: Good point! I forgot we had it for 38.

PR #168, opened 2020-02-24
Moved variant annotation hists functions to gnomad_hail

ch-kr:
from gnomad_qc (https://github.com/macarthur-lab/gnomad_qc/blob/master/gnomad_qc/v2/variant_qc/make_var_annot_hists.py)

### DIFF
--- a/gnomad_hail/utils/annotations.py
+++ b/gnomad_hail/utils/annotations.py
@@ -2,9 +2,9 @@ from gnomad_hail import *
 
 
 def pop_max_expr(
-        freq: hl.expr.ArrayExpression,
-        freq_meta: hl.expr.ArrayExpression,
-        pops_to_exclude: Optional[Set[str]] = None
+    freq: hl.expr.ArrayExpression,
+    freq_meta: hl.expr.ArrayExpression,
+    pops_to_exclude: Optional[Set[str]] = None,
 ) -> hl.expr.StructExpression:
     """
     Creates an expression containing popmax: the frequency information about the population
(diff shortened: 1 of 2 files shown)


gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +419,63 @@ def annotate_freq(
 
     # Return MT with freq row annotation
     return mt.annotate_rows(freq=freq_expr).drop('_freq_meta')
+
+
+def get_annotations_hists(
+    ht: hl.Table, annotations_hists: Dict[str, Tuple],
```
lfrancioli: I think moving `annotations_hists` to the next line would increase readability

gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +419,63 @@ def annotate_freq(
...
+    log10_annotations: List[str] = ['DP']
+    ) -> Dict[str, hl.expr.StructExpression]:
+    """
+    Creates histograms for variant metrics.
```
lfrancioli: How about "Creates histograms for variant metrics in `ht.info` ?"

gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +419,63 @@ def annotate_freq(
...
+    :param Table ht: Table with variant metrics
+    :param Dict annotations_hists: Dictionary of metrics names and their histogram values (start, end, bins)
+    :param List log10_annotations: List of metrics to log scale
+    :return: Dictionary of metrics and their histograms
+    :rtype: Dict[str, hl.expr.StructExpression]
+    """
+    hist_dict = {}
+
+    # Check all fields in ht.info and create histograms if they are in annotations_hists dict
+    info_fields = list(ht.row.info)
```
lfrancioli: unnecessary: you can directly write `for field in ht.row.info:`

lfrancioli: Also -and up to you if you like/dislike it- but you could write this like that if you want:
```
return {
    field: hl.agg.hist(
            hl.log10(ht[field] if field in log10_annotations else ht[field), start, end, bin
    ) for field, (start, end, bins) in annotations_hists.items() if field in ht.row.info
}
```

gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +419,63 @@ def annotate_freq(
...
+            start, end, bins = annotations_hists[field]
+            if field in log10_annotations:
+                hist_dict[field] = hl.agg.hist(hl.log10(ht[field]), start, end, bins)
+            else:
+                hist_dict[field] = hl.agg.hist(ht[field], start, end, bins)
+
+    return hist_dict
+
+
+def create_frequency_bins(ht: hl.Table) -> hl.Table:
```
lfrancioli: How about changing signature to be the following?
```
def create_frequency_bins_expr(
   ac: hl.expr.NumericExpression,
   af:  hl.expr.NumericExpression
) -> hl.expr.StringExpression:
```

gnomad_hail/utils/annotations.py
```diff
@@ -419,3 +419,63 @@ def annotate_freq(
...
+    """
+    Creates bins for frequencies in preparation for aggregating QUAL by frequency bin.
```
lfrancioli: I think it's worth adding the list of bins created

gnomad_hail/utils/annotations.py
nawatts: Plug for [Black](https://pypi.org/project/black/) to automatically resolve all formatting discussions.

gnomad_hail/utils/annotations.py
ch-kr: ooh, nice!

ch-kr:
Thanks for the suggestions! Back to you

PR #169, opened 2020-02-25
Refactor argument parsing for import resources script

nawatts:
Currently, import_resources.py creates an optional argument for each resource (and each version for versioned resources). This changes it to use one required argument for the list of resources to import. Thus, usage changes from `import_resources.py --grch37.clinvar.20181028` to `import_resources.py grch37.clinvar.20181028`.

This simplifies the argument handling code and removes restrictions on version IDs created by their being used in argument names (see https://github.com/macarthur-lab/gnomad_hail/pull/160#issuecomment-584714468).


gnomad_hail/resources/import_resources.py
```diff
@@ -38,45 +40,32 @@ def get_module_importable_resources(module, prefix: Optional[str] = None) -> Dic
...
+        resource.import_resource(args.overwrite)
 
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser()
-    grch37_grp = parser.add_argument_group('grch37 resources')
-    add_arguments_from_resources_dict(grch37_grp, grch37_resources)
-    grch38_grp = parser.add_argument_group('grch38 resources')
-    add_arguments_from_resources_dict(grch38_grp, grch38_resources)
+    wrapper = textwrap.TextWrapper(width=100, initial_indent=" " * 2, subsequent_indent=" " * 4)
```
nawatts: The usage output here is significantly wider than argparse's default. Using `100` here allows most resource paths and import paths to fit on one line instead of being wrapped.

gnomad_hail/resources/import_resources.py
```diff
@@ -38,45 +40,32 @@ def get_module_importable_resources(module, prefix: Optional[str] = None) -> Dic
...
+    resources_help = "Resource to import. Choices are:\n" + "\n".join(
+        itertools.chain.from_iterable(
```
lfrancioli: Can you wrap this in a function so that we can reuse this for resources importing in other projects (e.g. gnomad v3)?

PR #171, opened 2020-02-26
Added info ht calculation and split info ht functionality

ch-kr:
...to sparse_mt.py

### DIFF
--- a/gnomad_hail/utils/sparse_mt.py
+++ b/gnomad_hail/utils/sparse_mt.py
@@ -1,6 +1,6 @@
+from .annotations import get_lowqual_expr
 from .generic import *
 from .gnomad_functions import get_adj_expr
-from .variant_qc import get_lowqual_expr
 
 INFO_SUM_AGG_FIELDS= ['QUALapprox']
 INFO_INT32_SUM_AGG_FIELDS = ['VarDP']
@@ -367,6 +367,7 @@ def get_site_info_expr(
 
 def default_compute_info(
     mt: hl.MatrixTable,
+    site_annotations: bool = False,
     n_partitions: int = 5000
 ) -> hl.Table:
     """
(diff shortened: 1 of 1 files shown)


gnomad_hail/utils/sparse_mt.py
```diff
@@ -363,6 +365,90 @@ def get_site_info_expr(
...
+    ).rows()
+
+    # Add lowqual flag
+    info_ht = info_ht.annotate(
+        lowqual=get_lowqual_expr(
+            info_ht.alleles,
+            info_ht.info.QUALapprox,
+            # The indel het prior used for gnomad v3 was 1/10k bases (phred=40).
+            # This value is usually 1/8k bases (phred=39).
+            indel_phred_het_prior=40
```
lfrancioli: we should use the default here I think.

gnomad_hail/utils/sparse_mt.py
```diff
@@ -363,6 +365,90 @@ def get_site_info_expr(
     return info
 
 
+def default_compute_info(
+    mt: hl.MatrixTable,
+    n_partitions: int = 5000
```
lfrancioli: worth adding an option for adding the site-level info too?

gnomad_hail/utils/sparse_mt.py
```diff
@@ -363,6 +365,90 @@ def get_site_info_expr(
...
+            # The indel het prior used for gnomad v3 was 1/10k bases (phred=40).
+            # This value is usually 1/8k bases (phred=39).
+            indel_phred_het_prior=40
+        )
+    )
+
+    return info_ht.naive_coalesce(n_partitions)
+
+
+def split_info(ht: hl.Table) -> hl.Table:
```
lfrancioli: I think this may be better as an expr -> expr function, what do you think?

ch-kr: works for me!

PR #176, opened 2020-03-02
Restructure as namespace packages

nawatts:
Restructures the repository to have `gnomad.resources` and `gnomad.utils` as [namespace packages](https://packaging.python.org/guides/packaging-namespace-packages/).

A consequence of this change is that `from gnomad import ...` no longer works. Imports must be from one of the packages (`from gnomad.resources import ...` etc.). Some imports in util modules had to be changed because of that.


docs/generate_api_reference.py
```diff
@@ -133,4 +133,14 @@ def write_package_doc(package_name):
 
 
 if __name__ == "__main__":
-    write_package_doc("gnomad_hail")
+    packages = ["resources", "utils"]
```
lfrancioli: I'm not 100% sure I understand this so just wanted to double check: resources and utils will still be under the same package, right? Or is this separating them into two?

nawatts: They will be distributed together. pip installing `gnomad` will get you both `gnomad.resources` and `gnomad.utils`.

nawatts: However, since now the top level `gnomad` is a namespace instead of a package, it has to be handled differently here.

lfrancioli: all right, sounds good! And the docs will still be available in the same format together?

nawatts: Yes

lfrancioli: sweet!

PR #181, opened 2020-03-03
Added two annotations from v3 (absent from v2) to annotations hists c…

lfrancioli:
…onstants


lfrancioli:
@ch-kr super small one for you in the context of the annotations hists

lfrancioli:
OK, actually running this over the v3, I found a couple of bugs that I squashed in this PR too!

PR #186, opened 2020-03-05
Added function to generic that returns file size and md5

ch-kr:


gnomad/utils/generic.py
```diff
@@ -1327,3 +1327,36 @@ def rep_on_read(path: str, n_partitions: int) -> hl.MatrixTable:
...
+            size = int(value)
+            if size >= one_gibibyte:
+                info["size"] = f"{round(size / one_gibibyte, 2)} GiB"
+            else:
+                info["size"] = f"{round(size / one_mebibyte, 2)} MiB"
+
+        if label == "Hash (md5)":
+            info["md5"] = base64.b64decode(value).hex()
+
+    return (info["size"], info["md5"])
```
nawatts: It might be useful to return the integer size as well as the GiB/MiB string.

gnomad/utils/generic.py
```diff
@@ -1327,3 +1329,36 @@ def rep_on_read(path: str, n_partitions: int) -> hl.MatrixTable:
...
+            size = int(value)
+            if size >= one_gibibyte:
+                info["size"] = f"{round(size / one_gibibyte, 2)} GiB"
+            else:
+                info["size"] = f"{round(size / one_mebibyte, 2)} MiB"
+
+        if label == "Hash (md5)":
+            info["md5"] = base64.b64decode(value).hex()
+
+    return (int(info["size"]), info["size"], info["md5"])
```
nawatts: `int` won't work here, since the string contains "MiB" or "GiB". This should probably be the original `size` value in bytes.

ch-kr: hehe yes sorry also just realized this and pushed

lfrancioli:
Isn't this redundant with https://hail.is/docs/0.2/utils/index.html?highlight=hadoop#hail.utils.hadoop_stat

ch-kr:
I didn't realize that existed, it doesn't return the md5 does it?

lfrancioli:
no, so still useful!

PR #199, opened 2020-03-16
small enhancement to subset in generic.py

lfrancioli:


lfrancioli:
In addition, I think that naming an arg `gt_expr` that has a `str` type is misleading.. If this isn't used too broadly yet, I think we should either change the name or the type of the arg. Since the default for this field is different for dense vs sparse matrix, I wonder if it would be best to change the type..

ch-kr:
I don't have strong feelings about the GT arg; changing the type sounds good to me

PR #206, opened 2020-03-25
Add missing import statement

gtiao:


gnomad/utils/sample_qc.py
```diff
@@ -3,6 +3,7 @@
 from .gnomad_functions import logger, filter_to_adj
 from .sparse_mt import impute_sex_ploidy
 from gnomad.utils.relatedness import get_duplicated_samples
+from gnomad.utils.generic import filter_low_conf_regions
```
nawatts: This is already imported by `from .generic import *`, but 👍for the specific import.

jkgoodrich: Just realized how long this PR has been here, is it good to merge, should we switch to specific imports for anything else that is needed too, or keep the general import and not this one because it is redundant?

nawatts: Long term, I'd vote for all specific imports. We can also enforce that with PyLint (we currently don't fail on warnings because there are so many of them).

jkgoodrich: Great, thank you! I will modify and ask for a re-review so we can get this merged.

jkgoodrich: actually, I'm going to close this because after the reorganization it's not relevant anymore

PR #207, opened 2020-03-25
Repo reorganization 

mike-w-wilson:
## Summary 
This PR is pretty massive but since it is a reorganization of the entire repo, I couldn't think of another way to do it. This organization was drafted and previously discussed here: 

https://docs.google.com/spreadsheets/d/1pVpru0acCCr1AvSRyf89X-LxPKhwK-mYCBtZvICl9fQ/edit#gid=0 

The changes include import statements, loggers, and where functions live. Six functions have also been removed as they are built into hail or unnecessary for other reasons. They are:

gnomad.utils.generic.reverse_complement_bases(bases)
gnomad.utils.generic.get_sample_data(mt, …)
gnomad.utils.generic.phase_by_transmission(…)
gnomad.utils.generic.phase_trio_matrix_by_transmission(tm)
gnomad.utils.generic.explode_trio_matrix(tm)
gnomad.utils.generic.sample_pcs_uniformly(…)
gnomad.utils.liftover.main(args)

Once this is approved, I can run Black.


gnomad/utils/annotations.py
```diff
@@ -850,3 +850,26 @@ def fs_from_sb(
...
+
+    :param t: Input HT/MT
+    :return: Boolean expression selecting only bi-allelic sites
+    """
+    return (~t.was_split if 'was_split' in t.row else (hl.len(t.alleles) == 2))
+
+
+def unphase_call_expr(call_expr: hl.expr.CallExpression) -> hl.expr.CallExpression:
+    """
+    Generate unphased version of MatrixTable (assumes call is in mt.GT and is diploid or haploid only)
```
mike-w-wilson: This doc needs to be updated now

gnomad/sample_qc/ancestry.py
```diff
@@ -100,6 +99,38 @@ def assign_population_pcs(
     :return: Hail Table or Pandas Dataframe (depending on input) containing sample IDs and imputed population labels, trained random forest model
     """
     from sklearn.ensemble import RandomForestClassifier
+
+    def expand_pd_array_col(
```
mike-w-wilson: I considered this more of a util but I don't have any example of where else it would be used in current pipelines so moving it here works for me.

lfrancioli: yeah, I don't think we've ever used it for anything else. Also we really don't have much pandas tools so I think that nesting it under the sole function that uses it is the right move!

nawatts: If this isn't using anything from the outer function's scope, then unnesting it would make both functions a bit easier to read. If it's not intended to be public, adding a leading underscore to the function name marks it private / undocumented.

mike-w-wilson:
Regarding @nawatts 's comment on `__init__.py`, do we want to adopt `__all__`? If so, are there any submodules we would want to leave out of the imports? I can also make the change suggested and just import the submodules themselves ` from gnomad.sample_qc import ancestry ` I do like that this approach clears the namespace up.  Any thoughts there @lfrancioli ?

gnomad/sample_qc/ancestry.py
lfrancioli: well, this is really just a small helper for this function and it's unrelated to the rest of the module, I think nesting it is best

nawatts: Ok. Generally though, nesting a function makes it more difficult to understand because you can't tell whether or not the nested function uses the outer function's parameters or other variables in the outer function's scope. If it's not nested, then it's easy to see what the inputs are.

lfrancioli:
I like importing the submodules too

gnomad/sample_qc/ancestry.py
lfrancioli: All right, I inlined this. Since we don't need all the extra params and all it only adds 1 line of code to the main function body and I'm pretty sure will make everyone happy :)

nawatts: Excellent, 39 lines of code deleted in 4d1a31e. That looks like the best solution.

gnomad/utils/file_utils.py
```diff
@@ -54,70 +52,6 @@ def select_primitives_from_ht(ht: hl.Table) -> hl.Table:
                         v.dtype in {hl.tstr, hl.tint32, hl.tfloat32, hl.tint64, hl.tfloat64, hl.tbool}})
 
 
-def ht_to_vcf_mt(
```
jkgoodrich: Yeah I guess it does make since to move this to a `utils.vcf` because it is specific to writing to a vcf and since we are planning to add more to the vcf module at somepoint. Good catch

gnomad/utils/vcf.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Annotate with new expression and add 's' empty string field required to cast HT to MT
+    info_ht = info_ht.annotate(
+        info=info_ht.info.annotate(**info_expr),
+        s=hl.null(hl.tstr)
+    )
+
+    # Create an MT with no cols so that we acn export to VCF
+    info_mt = info_ht.to_matrix_table_row_major(columns=['s'], entry_field_name='s')
+    return info_mt.filter_cols(False)
```
jkgoodrich, line 72: looks like github is complaining about no newline at end of file

lfrancioli:
Thanks for weighing in @jkgoodrich ! A couple of things based on your comments:
1 .  Let's keep `ancestry.py` :)
2. I'm in for `random_forest.py`
3. Regarding `random_forest.pretty_print_runs`, let's keep it there then if you're using it

So I guess what I'm saying is that the current state of things is good with me ;)

gnomad/utils/annotations.py
```diff
@@ -1,12 +1,30 @@
...
 
-logger = logging.getLogger("gnomad.utils")
+import hail as hl
+from gnomad.utils.gen_stats import to_phred
+
+logging.basicConfig(format="%(asctime)s (%(name)s %(lineno)s): %(message)s", datefmt='%m/%d/%Y %I:%M:%S %p')
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+ANNOTATIONS_HISTS = {
```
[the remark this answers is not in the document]
lfrancioli: I think I'm fine with having this here as in my experience these values are good starting points at least and the different metrics are pretty standard. Totally open to suggestions though -- did you have another module in mind?
I think adding a comment about this is would be nice  (but I can see this being outside the scope of this PR and part of another effort to document constants)

lfrancioli:
This is awesome @mike-w-wilson 

So I think we should go all lowercase. Also, could you update the names with the more complete ones (and double check the colors match) those here:
https://github.com/macarthur-lab/gnomad_lof/blob/master/R/constants.R

Note that the Amish isn't in the R code as it's new for v3, so needs to be added. If you  see anything else that's interesting to have w.r.t. pops in these constants please transfer over too!

mike-w-wilson:
@Laurent, added the complete pop names and  pop colors but removed any unused pops with clashes.

PR #208, opened 2020-03-27
Updated ambiguous sex annotation (Ambiguous > ambiguous)

ch-kr:


ch-kr:
do you mean the sex karyotype annotations? I actually like having those capitalized because chrX and chrY are typically capitalized

lfrancioli:
I also think that we generally refer to chromosome `X` and not chromosome `x`

PR #210, opened 2020-03-30
Add getting started documentation

nawatts:


konradjk:
Looks great! Can you add a vignette for utils as well?

nawatts:
It looks like most of `utils` is used in generating the public dataset. What would be a good example of something that someone would be run on the public dataset?

konradjk:
I was asked about `process_consequences` ~3 times in the last week, the whole suite of VEP-processing utilities. Also things like `adjust_sex_ploidy` and other QC functions have been shared to the community and used in other non-gnomAD pipelines now as well.

nawatts:
Ok. I'll suggest we leave this page as is, with a very simple example of obtaining gnomAD data that will be applicable to almost every user. And then add a separate section with examples of tasks, such as running VEP, etc. That can also serve as sort of an FAQ. I can start that section in this PR with `process_consequences`.

nawatts:
Since VEP/process_consequences usage will be affected by the reorganization in #207, we may not want to publish that immediately. I'll open a separate PR for it so that this one isn't held up.

Issue #211, opened 2020-03-30
VEP configuration is tied to US region


nawatts:
The VEP configuration files which `get_vep_config` points to are specific to the US region. There are other replicates of VEP data in the Europe and Australia regions. `get_vep_config` should point to the same replicate that Hail uses. That can be found using `/usr/share/google/get_metadata_value attributes/VEP_REPLICATE`.

PR #212, opened 2020-03-30
Add documentation for using VEP utilities

nawatts:
This adds an Examples section to the documentation with instructions on how to use the gnomAD VEP utilities. (There's possibly a better name for this than "Examples").

### DIFF
--- a/docs/examples/vep.rst
+++ b/docs/examples/vep.rst
@@ -1,0 +1,39 @@
+Variant Effect Predictor (VEP)
+==============================
+
+To use the `Ensembl Variant Effect Predictor <https://www.ensembl.org/vep>`_ with Hail on Google Dataproc,
+the ``--vep`` flag must be included when starting the cluster. Note that a cluster's VEP configuration is
+tied to a specific reference genome.
+
+.. code-block:: shell
+
+   hailctl dataproc start cluster-name --vep GRCh37 --requester-pays-allow-all --packages gnomad
+
+.. note::
+
+   VEP data is stored in requester pays buckets. Reading from these buckets will bill charges to the project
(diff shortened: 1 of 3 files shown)


docs/examples/vep.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Import variants into a sites-only Hail Table::
+
+   import hail as hl
+
+   ds = hl.import_vcf("/path/to/data.vcf.gz", reference_genome="GRCh37", drop_samples=True).rows()
+
+Annotate variants with VEP consequences::
+
+   from gnomad.utils.generic import vep_or_lookup_vep
```
nawatts: These import paths will need to change after #207.

docs/examples/vep.rst
```diff
@@ -0,0 +1,39 @@
...
+
+   ds = vep_or_lookup_vep(ds, reference="GRCh37", vep_config="gs://hail-us-vep/vep85-loftee-gcloud.json")
```
nawatts: Changes to the default VEP configuration in #197 have not yet been published to PyPI. Once they are, the `vep_config` argument can be removed here.

konradjk:
Looks great! As an aside, the Hail team is making geo-redundant copies of the VEP cache and will match the copy of the data based on the location of the cluster

nawatts:
Yes. I added the `hail-us-vep` path here because the version of `get_vep_config` currently on PyPI still points to `hail-common`.  I filed an issue (#211) to update `get_vep_config` to return the URL for the appropriate copy.

nawatts:
Updated import paths now that #207 has been merged.

PR #218, opened 2020-04-15
Default compute info fix

jkgoodrich:
Just some small fixes/changes to annotations and variant QC functions from testing with UKBB QC


gnomad/sample_qc/relatedness.py
```diff
@@ -720,7 +720,8 @@ def filter_mt_to_trios(mt: hl.MatrixTable, fam_ht: hl.Table) -> hl.MatrixTable:
 
     mt = mt.filter_cols(hl.is_defined(fam_ht[mt.col_key]))
     mt = filter_to_autosomes(mt)
-    mt = annotate_adj(mt)
+    if "adj" not in list(mt.entry):
```
ch-kr, line 723: I think you might not need `in list(mt.entry)`; `in mt.entry` should work (but not 100%)

gnomad/variant_qc/pipeline.py
```diff
@@ -303,7 +303,8 @@ def generate_sib_stats(
         _localize=False,
     )
     mt = mt.filter_cols(s_to_keep.contains(mt.s))
-    mt = annotate_adj(mt)
+    if "adj" not in list(mt.entry):
```
ch-kr, line 306: same comment about mt.entry here

PR #223, opened 2020-05-11
Small variant annotation fix

jkgoodrich:


gnomad/variant_qc/pipeline.py
```diff
@@ -288,6 +298,8 @@ def generate_sib_stats(
     :param relationship_col: Column containing the relationship for the sample pair as defined in this module constants.
     :return: A Table with the sibling shared variant counts
     """
+    mt = filter_to_autosomes(mt)
```
ch-kr, line 301: should we add to the docstring that this currently only runs on autosomes?

nawatts:
Could you also add a note on the changes to the "Unreleased" section of [CHANGELOG.md](https://github.com/broadinstitute/gnomad_methods/blob/master/CHANGELOG.md)? That will help users see what changed between releases.

gnomad/variant_qc/pipeline.py
```diff
@@ -276,7 +278,6 @@ def generate_trio_stats(mt: hl.MatrixTable,) -> hl.Table:
 def generate_sib_stats(
     mt: hl.MatrixTable,
     relatedness_ht: hl.Table,
-    sex_ht: hl.Table,
```
ch-kr, line 269: are you planning to add this back in the future?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 * Updated usage of included intervals in imputing sex ploidy, also updated interval parameter names [(#209)](https://github.com/broadinstitute/gnomad_methods/pull/209)
 * Updated capitalization in relatedness constants [(#217)](https://github.com/broadinstitute/gnomad_methods/pull/217)
 * Changed interface for Slack notifications [(#219)](https://github.com/broadinstitute/gnomad_methods/pull/219)
+* Modified variant QC pipeline function `generate_sib_stats` to filter to autosomes [(#223)](https://github.com/broadinstitute/gnomad_methods/pull/223)
+* Added singleton de novo counts to variant QC pipeline function `score_bin_agg` [(#223)](https://github.com/broadinstitute/gnomad_methods/pull/223)
```
ch-kr, line 14: this should probably mention that some annotations are now necessary (e.g., ac_qc_samples_unrelated_raw) and some were removed (e.g., unrelated_qc_callstats), right?

gnomad/variant_qc/pipeline.py
jkgoodrich: i don't think we will need it if we are filtering to autosomes now

ch-kr: yup that makes sense, but generate trio stats expr still allows for sex chromosomes. maybe we should change that as well to be consistent?

jkgoodrich: Ah, so just remove `proband_is_female_expr=mt.is_female,` in `generate_trio_stats`

jkgoodrich: or do you think it would be better to require sibling stats to have is_female?

ch-kr: maybe the first one (removing `is_female`). that way, both `generate_trio_stats` and `generate_sib_stats` will default to computing on autosomes only, but users can run `generate_trio_stats_expr` or `generate_sib_stats_expr` if they want sex chromosomes for whatever reason

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Modified variant QC pipeline function `generate_sib_stats` to filter to autosomes [(#223)](https://github.com/broadinstitute/gnomad_methods/pull/223)
```
ch-kr, line 5: this includes `generate_trio_stats` now as well, right?

gnomad/variant_qc/pipeline.py
```diff
@@ -243,10 +258,18 @@ def generate_trio_stats(mt: hl.MatrixTable,) -> hl.Table:
...
     :param mt: A Trio Matrix Table returned from `hl.trio_matrix`. Must be dense
+    :param autosomes_only: If set, only autosomal intervals are used.
+    :param bi_allelic_only: If set, only bi-allelic sites are used for the computation
     :return: Table with trio stats
     """
-    mt = mt.filter_rows(hl.len(mt.alleles) == 2)
+    if autosomes_only:
+        mt = filter_to_autosomes(mt)
+    if bi_allelic_only:
+        mt = mt.filter_rows(bi_allelic_expr(mt))
```
ch-kr, line 271: thanks for adding this!

PR #224, opened 2020-05-12
Add variant QC pipeline functions from gnomad_qc code

jkgoodrich:
This PR is a bit big, but most of it is generalizing or adding functions from gnomad_qc that we will also be using in other pipelines

### DIFF
--- a/gnomad/variant_qc/pipeline.py
+++ b/gnomad/variant_qc/pipeline.py
@@ -14,7 +14,11 @@ from gnomad.sample_qc.relatedness import (
 from gnomad.utils.annotations import annotate_adj
 from gnomad.utils.reference_genome import get_reference_genome
 from gnomad.variant_qc.evaluation import compute_quantile_bin
-from gnomad.variant_qc.random_forest import get_features_importance, test_model, train_rf
+from gnomad.variant_qc.random_forest import (
+    get_features_importance,
+    test_model,
+    train_rf,
+)
 from gnomad_qc.v2.variant_qc.variantqc import sample_rf_training_examples
 
 
@@ -328,17 +332,17 @@ def generate_sib_stats(
 
 
 def generate_rf_training(
-        ht: hl.Table,
-        rf_features: List[str],
-        tp_expr: hl.expr.BooleanExpression,
-        fp_expr: hl.expr.BooleanExpression,
-        fp_to_tp: float = 1.0,
-        num_trees: int = 500,
-        max_depth: int = 5,
-        label_col: str = "rf_label",
-        train_col: str = "rf_train",
-        test_intervals: Union[str, List[str]] = 'chr20',
-        filter_expr: Optional[hl.expr.BooleanExpression] = None,
+    ht: hl.Table,
+    rf_features: List[str],
+    tp_expr: hl.expr.BooleanExpression,
+    fp_expr: hl.expr.BooleanExpression,
+    fp_to_tp: float = 1.0,
+    num_trees: int = 500,
+    max_depth: int = 5,
+    label_col: str = "rf_label",
+    train_col: str = "rf_train",
+    test_intervals: Union[str, List[str]] = "chr20",
+    filter_expr: Optional[hl.expr.BooleanExpression] = None,
 ) -> Tuple[hl.Table, pyspark.ml.PipelineModel]:
     """
     Perform random forest (RF) training using a Table annotated with features and training data.
@@ -374,14 +378,19 @@ def generate_rf_training(
         ht = ht.annotate_globals(test_intervals=test_intervals_locus)
 
     ht = sample_rf_training_examples(
-        ht, tp_col="tp", fp_col='fp', fp_to_tp=fp_to_tp, label_col=label_col, train_col=train_col
+        ht,
+        tp_col="tp",
+        fp_col="fp",
+        fp_to_tp=fp_to_tp,
+        label_col=label_col,
+        train_col=train_col,
     )
     ht = ht.persist()
 
     rf_ht = ht
 
     if filter_expr is not None:
-        logger.info('Filtering training set using filter_expr')
+        logger.info("Filtering training set using filter_expr")
         rf_ht = rf_ht.filter(rf_ht._filter)
     rf_ht.drop("_filter")
 
@@ -395,10 +404,7 @@ def generate_rf_training(
 
     logger.info(
         "Training RF model:\nfeatures: {}\nnum_tree: {}\nmax_depth:{}\nTest intervals: {}".format(
-            ",".join(rf_features),
-            num_trees,
-            max_depth,
-            ",".join(test_intervals_str),
+            ",".join(rf_features), num_trees, max_depth, ",".join(test_intervals_str),
         )
     )
 
@@ -412,9 +418,7 @@ def generate_rf_training(
 
     test_results = None
     if test_intervals:
-        logger.info(
-            f"Testing model on intervals {','.join(test_intervals_str)}"
-        )
+        logger.info(f"Testing model on intervals {','.join(test_intervals_str)}")
         test_ht = hl.filter_intervals(ht, test_intervals_locus, keep=True)
         test_ht = test_ht.filter(hl.is_defined(test_ht[label_col]))
         test_results = test_model(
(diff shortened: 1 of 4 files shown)


gnomad/variant_qc/pipeline.py
```diff
@@ -320,3 +330,354 @@ def generate_sib_stats(
...
+        & (rf_result_ht.rf_probability["TP"] < rf_result_ht.rf_indel_cutoff.min_score)
+    )
+    rf_result_ht = rf_result_ht.annotate(
+        filters=hl.case()
+        .when(rf_filter_criteria, {"RF"})
+        .when(~rf_filter_criteria, hl.empty_set(hl.tstr))
+        .or_error("Missing RF probability!")
+    )
+
+    inbreeding_coeff_filter_criteria = hl.is_defined(rf_result_ht.inbreeding_coeff) & (
```
ch-kr, line 648: what happens to the lines where this isn't defined?

gnomad/variant_qc/pipeline.py
```diff
@@ -320,3 +330,354 @@ def generate_sib_stats(
...
+        rf_result_ht.inbreeding_coeff < inbreeding_coeff_cutoff
+    )
+    rf_result_ht = rf_result_ht.annotate(
+        filters=hl.cond(
```
ch-kr, line 652: cond -> if_else

gnomad/variant_qc/pipeline.py
```diff
@@ -320,3 +330,354 @@ def generate_sib_stats(
...
+            inbreeding_coeff_filter_criteria,
+            rf_result_ht.filters.add("InbreedingCoeff"),
+            rf_result_ht.filters,
+        )
+    )
+
+    rf_result_ht = rf_result_ht.annotate(
```
ch-kr, line 659: does it make sense to combine this annotate with the annotate above (line 651)?

gnomad/variant_qc/random_forest.py
```diff
@@ -159,6 +159,93 @@ def get_columns_quantiles(
     return res
 
 
+def median_impute_features(
+    ht: hl.Table,
+    impute_features_by_variant_type: bool = False,
```
lfrancioli, line 164: should we make this an `expr` ? Seems like it'd make more sense now that this is going into a more generic lib.

lfrancioli: would also remove the need for that note

gnomad/variant_qc/random_forest.py
```diff
@@ -159,6 +159,93 @@ def get_columns_quantiles(
...
+    n_variants_median: int = 50000,
+) -> hl.Table:
+    """
+    Numerical features are median-imputed by Hail's `approx_median`. If `impute_features_by_variant_type` is set,
+    imputation is done based on the median of the variant type.
+
```
lfrancioli, line 170: would be good to add exactly what annotations will be added to the table.

gnomad/variant_qc/pipeline.py
```diff
@@ -320,3 +330,354 @@ def generate_sib_stats(
...
+        features_importance=features_importance,
+        features=rf_features,
+        test_intervals=test_intervals_str,
+        test_results=test_results,
+    )
+
+    return ht, rf_model
+
+
+def generate_final_rf_ht(
```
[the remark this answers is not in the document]
jkgoodrich: So I think I wanted it to be RF-dependent since without that it seems like it is really just `add_filters_expr`. I changed the code to use `add_filters_expr` now. I am starting to wonder if I should just keep this out of `pipeline.py` and put back into just the UKBB files. I removed it for now, but we can talk about it more.

gnomad/variant_qc/pipeline.py
```diff
@@ -320,3 +328,97 @@ def generate_sib_stats(
...
+
+def train_rf_model(
+    ht: hl.Table,
+    rf_features: List[str],
+    tp_expr: hl.expr.BooleanExpression,
+    fp_expr: hl.expr.BooleanExpression,
+    fp_to_tp: float = 1.0,
+    num_trees: int = 500,
+    max_depth: int = 5,
+    test_expr: Optional[
```
lfrancioli, line 341: Is this option necessary? I'd have thought that just a Boolean expr would do. Seems like users can easily do 
```
rf_test_expr = hl.literal(test_expr).any(
            lambda interval: interval.contains(ht.locus)
        )
```
if they want. If using a test chrom it might be as simple as `ht.locus.contig == test_chrom`

gnomad/variant_qc/random_forest.py
```diff
@@ -159,6 +159,72 @@ def get_columns_quantiles(
...
+def median_impute_features(
+    ht: hl.Table, strata: Optional[Dict[str, hl.expr.Expression]] = None
+) -> hl.Table:
+    """
+    Numerical features in the Table are median-imputed by Hail's `approx_median`.
+
+    If a `strata` dict is given, imputation is done based on the median of of each stratification.
+
+    The annotations that are added to the Table are
+        - feature_imputed - A variant annotation indicating if each numerical feature was imputed or not.
```
lfrancioli, line 171: I'm being nit-picky, but this is really a `row` annotation.

gnomad/variant_qc/pipeline.py
```diff
@@ -320,3 +328,97 @@ def generate_sib_stats(
...
+    if isinstance(test_expr, list):
+        rf_test_expr = hl.literal(test_expr).any(
+            lambda interval: interval.contains(ht.locus)
+        )
+    elif isinstance(test_expr, hl.expr.BooleanExpression):
+        rf_test_expr = test_expr
+    else:
+        rf_test_expr = False
+
+    ht = ht.annotate(tp=tp_expr, fp=fp_expr, rf_test=rf_test_expr,)
```
lfrancioli, line 382: I kind of like how you've used `_tp` and `_fp` in `sample_training_examples` as it gives a clue that these annotations won't be returned. Up to you, but I'd support the change here too :)

gnomad/variant_qc/training.py
```diff
@@ -0,0 +1,130 @@
...
+            prob_fp = desired_fp / n_fp
+        else:
+            prob_tp = n_fp / desired_fp
+            prob_fp = 1.0
+
+        logger.info(
+            f"Training examples sampling: tp={prob_tp}*{n_tp}, fp={prob_fp}*{n_fp}"
+        )
+
+        if prob_fp < 1.0:
```
lfrancioli, line 104: So I find the code below quite convoluted. I think the following does what we want and is much easier to read....
```
    test_expr = (
        hl.case(missing_false=True)
        .when(ht._fp & hl.or_else(~ht._tp, True), hl.rand_bool(prob_fp))
        .when(ht._tp & hl.or_else(~ht._fp, True), hl.rand_bool(prob_tp))
        .default(False)
    )
```
Note that `prob_tp = prob_fp = 1.0` will need to be added at line 90 too I think.

gnomad/variant_qc/training.py
```diff
@@ -0,0 +1,130 @@
...
+    fp_expr: hl.BooleanExpression,
+    fp_to_tp: float = 1.0,
+    test_expr: Optional[hl.expr.BooleanExpression] = None,
+) -> hl.Table:
+    """
+    Returns a Table of all positive and negative training examples in `ht` with an annotation for those that should be
+    used for training given a true positive (TP) to false positive (FP) ratio.
+
+    The returned Table has the following annotations:
+        - train: indicates if the variant should be used for training. The only variants labeled as False will be
```
lfrancioli, line 24: There are two other cases for variants to be excluded for training:
- variants that are `True` for both `tp_expr` and `fp_expr` will also be `False`
- variants that are pruned out to obtain the desired `fp_to_tp` ratio.

lfrancioli:
One more thing, seems like we've lost `generate_final_rf_ht` in the process -- was this on purpose?

gnomad/variant_qc/training.py
jkgoodrich: Oh yes, thank you, that is so much easier to read

gnomad/variant_qc/pipeline.py
```diff
@@ -320,3 +328,86 @@ def generate_sib_stats(
...
+        features=rf_features,
+        label="rf_label",
+        num_trees=num_trees,
+        max_depth=max_depth,
+    )
+
+    test_results = None
+    if test_expr is not None:
+        logger.info(f"Testing model on specified variants or intervals...")
+        test_ht = ht.filter(hl.is_defined(ht.rf_label) & ht.rf_test)
```
ch-kr, line 401: this might not be necessary, but it might be nice to add to the docstring that rf_label needs to be present (isn't returned by the function) if `test_expr` is specified

gnomad/variant_qc/random_forest.py
```diff
@@ -159,6 +159,73 @@ def get_columns_quantiles(
...
+        )
+
+    else:
+        ht = ht.annotate_globals(
+            feature_medians=ht.aggregate(median_agg_expr, _localize=False)
+        )
+        feature_median_expr = ht.feature_medians
+
+    ht = ht.annotate(
+        **{f: hl.or_else(ht[f], feature_median_expr[f]) for f in numerical_features},
```
ch-kr, line 220: this only returns the median for the feature if `ht[f]` is missing, right? shouldn't it be the opposite?

gnomad/variant_qc/pipeline.py
jkgoodrich: rf_label is added above by `sample_training_examples`

lfrancioli:
> Yeah I removed `generate_final_rf_ht` intentionally [#224 (comment)](https://github.com/broadinstitute/gnomad_methods/pull/224#discussion_r434726428)

All right. Let's revisit once we have a plan for the finalization script in general.

gnomad/variant_qc/random_forest.py
lfrancioli: this is correct: you want to median-impute values that are missing.

gnomad/variant_qc/pipeline.py
lfrancioli: I'm not sure if I'm not understanding this correctly, but `rf_label` is generated within the function using `tp_expr` and `fp_expr` and it is returned. Am I missing something?

ch-kr: nope, just extra groggy today, sorry!

PR #238, opened 2020-07-14
Runtime variant QC fixes

jkgoodrich:
Added changed that were needed in order to run the common code variant QC functions in the UKBB variant QC scripts

### DIFF
--- a/gnomad/variant_qc/pipeline.py
+++ b/gnomad/variant_qc/pipeline.py
@@ -55,7 +55,7 @@ def create_binned_ht(
     :return: table with bin number for each variant
     """
 
-    def update_bin_expr(
+    def _update_bin_expr(
         bin_expr: Dict[str, hl.expr.BooleanExpression],
         new_expr: hl.expr.BooleanExpression,
         new_id: str,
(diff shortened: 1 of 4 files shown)


gnomad/variant_qc/pipeline.py
```diff
@@ -89,31 +87,27 @@ def update_bin_expr(
...
     if add_substrat:
         for add_id, add_expr in add_substrat.items():
-            bin_expr = update_bin_expr(bin_expr, add_expr, add_id)
+            bin_expr = _update_bin_expr(bin_expr, add_expr, add_id)
 
     bin_ht = compute_quantile_bin(
         ht, score_expr=ht.score, bin_expr=bin_expr, n_bins=n_bins
     )
+
+    ht = ht.select_globals()
```
ch-kr, line 110: basic question, why do you not want the globals?

jkgoodrich: Because the function I pass this to modifies the HT and returns it while adding globals. When I do the join, I only want one copy of the original globals

PR #240, opened 2020-07-17
Allow compute_binned_truth_sample_concordance to compute other bins

jkgoodrich:
Small change so I can perform the binning easily on only well covered intervals

### DIFF
--- a/gnomad/variant_qc/evaluation.py
+++ b/gnomad/variant_qc/evaluation.py
@@ -288,53 +288,76 @@ def compute_grouped_binned_ht(
 
 
 def compute_binned_truth_sample_concordance(
-    ht: hl.Table, binned_score_ht: hl.Table, n_bins: int = 100
+    ht: hl.Table,
+    binned_score_ht: hl.Table,
+    n_bins: int = 100,
+    add_bins: Optional[Dict[str, hl.expr.BooleanExpression]] = None,
 ) -> hl.Table:
     """
     Determines the concordance (TP, FP, FN) between a truth sample within the callset and the samples truth data
     grouped by bins computed using `compute_quantile_bin`.
 
     .. note::
-
         The input 'ht` should contain three row fields:
             - score: value to use for quantile binning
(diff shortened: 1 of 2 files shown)


gnomad/variant_qc/evaluation.py
```diff
@@ -288,54 +288,77 @@ def compute_grouped_binned_ht(
...
             - score: value to use for quantile binning
             - GT: a CallExpression containing the genotype of the evaluation data for the sample
             - truth_GT: a CallExpression containing the genotype of the truth sample
-
         The input `binned_score_ht` should contain:
              - score: value used to bin the full callset
              - bin: the full callset quantile bin
 
+    'add_bins` can be used to add additional global and truth sample binning to the final binned truth sample
+    concordance HT. The keys in `add_bins` must be present in `indexed_binned_score_ht` and the values in `add_bins`
```
ch-kr, line 310: this might be a tiny bit clearer if you change indexed_binned_score_ht to binned_score_ht, since the indexed ht is created within the function and not passed in/referenced elsewhere in the docstring (but not a really important change)

gnomad/variant_qc/evaluation.py
```diff
@@ -288,54 +288,77 @@ def compute_grouped_binned_ht(
...
-    # Explode the global and truth sample bins
-    ht = ht.annotate(
-        bin=[
-            hl.tuple(["global_bin", ht.global_bin]),
-            hl.tuple(["truth_sample_bin", ht.truth_sample_bin]),
+    bin_list = [
+        hl.tuple(["global_bin", ht.global_bin]),
+        hl.tuple(["truth_sample_bin", ht.truth_sample_bin]),
+    ]
+    bin_list.extend(
```
ch-kr, line 349: basic question, is there a reason why you can't create bin_list in one expression (why do you need the two `extend`s?)

ch-kr: oops, nevermind!

PR #241, opened 2020-07-22
Vcf export

ch-kr:
Added constants and some functions that are used during MT/HT>VCF export


gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+def make_hist_bin_edges_expr(ht: hl.Table, prefix: str) -> Dict[str, str]:
+    """
+    Create dictionaries containing variant histogram annotations and their associated bin edges, formatted into a string
+    separated by pipe delimiters.
+
+    :param ht: Table containing histogram variant annotations.
+    :param prefix: Prefix text for age histogram bin edges. 
+    :return: Dictionary keyed by histogram annotation name, with corresponding reformatted bin edges for values.
+    """
+    edges_dict = {
```
jkgoodrich, line 651: Maybe condense to something like: 

```
    edges_dict = {
        f"{prefix}{subset}": "|".join(
            map(lambda x: f"{x:.1f}", ht.head(1)[f'age_hist_{subset}'].collect()[0].bin_edges)
        )
    for subset in ["het", "hom"]
    }
```

gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+        ),
+        f"{prefix}hom": "|".join(
+            map(lambda x: f"{x:.1f}", ht.head(1).age_hist_hom.collect()[0].bin_edges)
+        ),
+    }
+    for hist in HISTS:
+
+        # Parse hists calculated on both raw and adj-filtered data
+        for hist_type in ["adj_qual_hists", "qual_hists"]:
+            hist_name = hist
```
jkgoodrich, line 663: add in outer for loop

gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+    """
+    Create dictionaries containing variant histogram annotations and their associated bin edges, formatted into a string
+    separated by pipe delimiters.
+
+    :param ht: Table containing histogram variant annotations.
+    :param prefix: Prefix text for age histogram bin edges. 
+    :return: Dictionary keyed by histogram annotation name, with corresponding reformatted bin edges for values.
+    """
+    edges_dict = {
+        f"{prefix}het": "|".join(
```
jkgoodrich, line 652: Is prefix expected to have and _ at the end? I think it might be better to just add an _ if prefix is not "" and set prefix default as ""

gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+        ),
+    }
+    for hist in HISTS:
+
+        # Parse hists calculated on both raw and adj-filtered data
+        for hist_type in ["adj_qual_hists", "qual_hists"]:
+            hist_name = hist
+            if "adj" in hist_type:
+                hist_name = f"{hist}_adj"
+            edges_dict[hist_name] = (
```
jkgoodrich, line 666: Can condense to:
```
            edges_dict[hist_name] = (
                "|".join(
                    map(
                        lambda x: f"{x:.2f}" if "ab" in hist else str(int(x)),
                        ht.head(1)[hist_type][hist].collect()[0].bin_edges,
                    )
                )
            )
```

gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+def make_hist_dict(bin_edges: Dict[str, Dict[str, str]], adj: bool) -> Dict[str, str]:
+    """
+    Generate dictionary of Number and Description attributes to be used in the VCF header, specifically for histogram annotations.
+
+    :param bin_edges: Dictionary keyed by histogram annotation name, with corresponding string-reformatted bin edges for values.
+    :param adj: Whether to create a header dict for raw or adj quality histograms.
+    :return: Dictionary keyed by VCF INFO annotations, where values are Dictionaries of Number and Description attributes.
+    """
+    header_hist_dict = {}
+    for hist in HISTS:
```
jkgoodrich, line 693: Just thought about this, but for all functions that use the constant HISTS, make hists a parameter with HISTS as default

gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+                "Description": f"Count of {hist_fields[0].upper()} values falling above highest histogram bin edge {hist_text}",
+            },
+        }
+
+        header_hist_dict.update(hist_dict)
+
+    return header_hist_dict
+
+
+def sample_sum_check(
```
jkgoodrich, line 728: This is something we want to check on the internal HT sanity checks right? I think it should also be moved somewhere else like the other function I mentioned above, but again not positive where to put it

gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+    """
+    combo_AC = [ht.info[f"{prefix}AC_{x}"] for x in make_label_combos(label_groups)]
+    combo_AN = [ht.info[f"{prefix}AN_{x}"] for x in make_label_combos(label_groups)]
+    combo_nhomalt = [
+        ht.info[f"{prefix}nhomalt_{x}"] for x in make_label_combos(label_groups)
+    ]
+
+    group = label_groups.pop("group")[0]
+    alt_groups = "_".join(
+        sorted(label_groups.keys(), key=lambda x: SORT_ORDER.index(x))
```
jkgoodrich, line 756: make SORT_ORDER a parameter with constant as default

gnomad/utils/vcf.py
```diff
@@ -1,11 +1,106 @@
...
+
+from gnomad.sample_qc.ancestry import POP_NAMES
 
 logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+SORT_ORDER = ["popmax", "group", "pop", "subpop", "sex"]
+"""
+Order to sort subgroupings during VCF export.
```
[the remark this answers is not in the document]
ch-kr: yup, that's my understanding. will update

gnomad/utils/vcf.py
```diff
@@ -14,6 +109,158 @@
     "AS_SB_TABLE",
 ]
 
+INFO_DICT = {
```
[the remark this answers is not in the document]
ch-kr: added. Laurent had a note about making these flags all caps in the original code (https://github.com/broadinstitute/gnomad_qc/blob/ff51c685d63e723f3f66f9d2ac88cf1a5f0decc1/gnomad_qc/v2/variant_qc/prepare_data_release.py#L44), do you think we should change that here?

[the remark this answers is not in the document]
ch-kr: I added this function to add the AS fields to the info dict: `add_as_info_dict`, would you prefer I just make another constant?

gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+        ht = ht.flatten()
+        ht.select("locus", "alleles", *display_fields).show()
+    else:
+        logger.info(f"PASSED {check_description} check")
+        if verbose:
+            ht_orig = ht_orig.flatten()
+            ht_orig.select(*display_fields).show()
+
+
+def make_filters_sanity_check_expr(ht: hl.Table) -> Dict[str, hl.expr.Expression]:
```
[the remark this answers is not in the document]
ch-kr: I moved this here because I run this sanity check on the VCF MT before export, but happy to move it again

gnomad/utils/vcf.py
```diff
@@ -92,3 +339,488 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+    """
+    Make Hail Expressions to measure % variants filtered under varying conditions of interest.
+    :param ht: Hail Table containing 'filter' annotation to be examined.
+    :return: Dictionary containing Hail aggregation expressions to measure filter flags.
+    """
+    filters_dict = {
+        "n": hl.agg.count(),
+        "frac_any_filter": hl.agg.count_where(ht.is_filtered) / hl.agg.count(),
```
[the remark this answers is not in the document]
ch-kr: I vaguely remember this being important for callrate, but probably fine to use fraction here. will update

gnomad/utils/vcf.py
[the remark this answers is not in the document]
ch-kr: oops just saw your comment, resolving this

gnomad/utils/vcf.py
```diff
@@ -92,3 +345,536 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+    :return: Dictionary with reset annotations
+    """
+    metrics = list(t.row.info)
+    female_metrics = [x for x in metrics if "_female" in x]
+    female_metrics = [
+        x for x in female_metrics if ("nhomalt" in x) or ("AC" in x) or ("AN" in x)
+    ]
+
+    if faf:
+        female_metrics.extend([x for x in female_metrics if "faf" in x])
```
gtiao, line 868: I'm confused by this line: it looks like you're extending the existing list `female_metrics` by faf elements that are already in the list

gnomad/utils/vcf.py
```diff
@@ -92,3 +344,536 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+        - Fraction of variants removed only because of AC0 filter
+        - Fraction of variants removed only because of random forest filtering
+
+
+    :param ht: Table containing 'filter' annotation to be examined.
+    :return: Dictionary containing Hail aggregation expressions to measure filter flags.
+    """
+    filters_dict = {
+        "n": hl.agg.count(),
+        "frac_any_filter": hl.agg.fraction(hl.len(ht.filters) == 0),
```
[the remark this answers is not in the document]
ch-kr: thanks for pointing this out, this should be `!=`

gnomad/utils/vcf.py
```diff
@@ -14,6 +112,160 @@
...
+    "transmitted_singleton": {
+        "Description": "Variant was a callset-wide doubleton that was transmitted within a family from a parent to a child (i.e., a singleton amongst unrelated samples in cohort)"
+    },
+    "original_alleles": {"Description": "Alleles before splitting multiallelics"},
+    "variant_type": {
+        "Description": "Variant type (snv, indel, multi-snv, multi-indel, or mixed)"
+    },
+    "allele_type": {
+        "Number": "A",
+        "Description": "Allele type (snv, insertion, deletion, or mixed)",
```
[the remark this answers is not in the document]
ch-kr: good point. I took this straight out of the gnomAD release code. I'll remove the "Number": "A" since this is a site annotation (we add it with https://github.com/broadinstitute/gnomad_methods/blob/e78dea469af3d90097edbe2d9d763216d874ee4b/gnomad/utils/annotations.py#L716 before splitting multiallelics)

gnomad/utils/vcf.py
```diff
@@ -92,3 +344,536 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+                "Number": "A",
+                "Description": f"Count of age values falling below lowest histogram bin edge for heterozygous individuals{description_text}",
+            },
+            f"{prefix}age_hist_het_n_larger": {
+                "Number": "A",
+                "Description": f"Count of age values falling above highest histogram bin edge for heterozygous individuals{description_text}",
+            },
+            f"{prefix}age_hist_hom_bin_freq": {
+                "Number": "A",
+                "Description": f"Histogram of ages of homozygous alternate individuals{description_text}; bin edges are: {bin_edges['het']}; total number of individuals of any genotype bin: {age_hist_data}",
```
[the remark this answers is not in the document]
ch-kr: the bin edges should be consistent between the two. I can change this to be `bin_edges['hom']` to be less confusing though

gnomad/utils/vcf.py
ch-kr: faf elements aren't in the list -- `female_metrics` first pulls all annotations that contain `_female` and then corrects that list to keep only elements with both `_female` and `AC`, `AN`, or `nhomalt`. the faf annotations (`faf95_female`) don't contain AC, AN, or nhomalt, so they aren't kept in that step.

that said, why do we set AC, AN, and nhomalt to 0 but leave AF? I'm going to change this to just be `female_metrics = [x for x in metrics if "_female" in x]` if that works with you and @jkgoodrich

gnomad/utils/vcf.py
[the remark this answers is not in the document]
jkgoodrich: I'm not sure. @lfrancioli what are your thoughts, what was the reasoning behind the note to switch to all caps?

gnomad/utils/vcf.py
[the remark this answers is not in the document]
jkgoodrich: good point, it is really more of a site_type. @gtiao, do you think we should we change it at this point, or keep as is to be consistent with past call sets? I lean to keep, but fine either way.

gnomad/utils/vcf.py
```diff
@@ -92,3 +340,534 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+
+        for combo in combos:
+            combo_fields = combo.split("_")
+            group_dict = dict(zip(group_types, combo_fields))
+
+            if not faf:
+                combo_dict = {
+                    f"{prefix}AC_{combo}": {
+                        "Number": "A",
+                        "Description": f"Alternate allele count{make_combo_header_text('for', group_dict, prefix)}",
```
jkgoodrich, line 595: Not necessary, but `make_combo_header_text('for', group_dict, prefix)` and `make_combo_header_text('to', group_dict, prefix)` are called several times. Could make them variables and use those here.

gnomad/utils/vcf.py
```diff
@@ -92,3 +340,534 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+    :param prefix: String indicating sample subset.
+    :param label_groups: Dictionary containing an entry for each label group, where key is the name of the grouping,
+        e.g. "sex" or "pop", and value is a list of all possible values for that grouping (e.g. ["male", "female"] or ["afr", "nfe", "amr"]).
+    :param verbose: If True, show top values of annotations being checked, including checks that pass; if False,
+        show only top values of annotations that fail checks.
+    :param subpop: Subpop abbreviation, supplied only if subpopulations are included in the annotation groups being checked.
+    :param sort_order: List containing order to sort label group combinations. Default is SORT_ORDER.
+    :return: None
+    """
+    combo_AC = [ht.info[f"{prefix}AC_{x}"] for x in make_label_combos(label_groups)]
```
jkgoodrich, line 801: Maybe make ` make_label_combos(label_groups)` a variable since it is used 3 times here?

gnomad/utils/vcf.py
```diff
@@ -92,3 +340,534 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+    """
+    metrics = list(t.row.info)
+    female_metrics = [x for x in metrics if "_female" in x]
+
+    female_metrics_dict = {}
+    for metric in female_metrics:
+        female_metrics_dict.update(
+            {
+                f"{metric}": hl.or_missing(
+                    (~t.locus.contig.in_y_nonpar() | ~t.locus.contig.in_y_par()),
```
jkgoodrich, line 868: I am likely just missing something here, but why can't it just be `~(t.locus.contig == 'Y')`?

gnomad/utils/vcf.py
jkgoodrich: hmm, interesting that AF is not set to 0. I will defer to @gtiao for this one, there may have been motivation for this that I am not aware of.

gnomad/utils/vcf.py
```diff
@@ -92,3 +340,534 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+    verbose: bool,
+) -> None:
+    """
+    Check a generic logical condition involving annotations in a Hail Table and print the results to terminal.
+
+    Displays the number of rows in the Table that return False for a condition (`cond_expr`). 
+
+    .. note::
+        This function checks for the number of rows that violate the `cond_expr` and prints the desired condition (`check_description`) to the terminal.
+        `cond_expr` and `check_description` are opposites and should never be the same.
```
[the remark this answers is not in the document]
ch-kr: reworded docstring and added example

gnomad/utils/vcf.py
```diff
@@ -92,3 +340,534 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+        - Fraction of variants removed only because of random forest filtering
+
+
+    :param ht: Table containing 'filter' annotation to be examined.
+    :return: Dictionary containing Hail aggregation expressions to measure filter flags.
+    """
+    filters_dict = {
+        "n": hl.agg.count(),
+        "frac_any_filter": hl.agg.fraction(hl.len(ht.filters) != 0),
+        "frac_inbreed_coeff": hl.agg.fraction(ht.filters.contains("inbreeding_coeff")),
```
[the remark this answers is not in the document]
ch-kr: I forgot about this and just changed it to InbreedingCoeff, happy to change back if necessary

gnomad/utils/vcf.py
ch-kr: I could change it, but then I'd need to infer the build to get the name of the Y contig. I thought keeping the built in hail functions might be more flexible

jkgoodrich: Ah, yes, that is a good reason

gnomad/utils/vcf.py
```diff
@@ -92,3 +340,534 @@ def get_pipe_expr(array_expr: hl.expr.ArrayExpression) -> hl.expr.StringExpressi
...
+        ht = ht.flatten()
+        ht.select("locus", "alleles", *display_fields).show()
+    else:
+        logger.info(f"PASSED {check_description} check")
+        if verbose:
+            ht_orig = ht_orig.flatten()
+            ht_orig.select(*display_fields).show()
+
+
+def make_filters_sanity_check_expr(ht: hl.Table) -> Dict[str, hl.expr.Expression]:
```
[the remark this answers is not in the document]
jkgoodrich: I don't see anything for this flag in the gnomAD common code, which is completely fine, but do you think we should add this check to the UKBB (sorry if it was already added and I just missed it)?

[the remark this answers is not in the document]
ch-kr: ah sorry I didn't add this yet

gnomad/utils/vcf.py
```diff
@@ -639,43 +619,45 @@ def make_vcf_filter_dict(
...
+        prefix += "_"
+
     edges_dict = {
-        f"{prefix}het": "|".join(
-            map(lambda x: f"{x:.1f}", ht.head(1).age_hist_het.collect()[0].bin_edges)
-        ),
-        f"{prefix}hom": "|".join(
-            map(lambda x: f"{x:.1f}", ht.head(1).age_hist_hom.collect()[0].bin_edges)
-        ),
+        f"{prefix}{call_type}": "|".join(
```
gtiao, line 639: Another nice consolidation of code

gnomad/assessment/sanity_checks.py
```diff
@@ -0,0 +1,166 @@
...
+    display_fields: List[str],
+    verbose: bool,
+) -> None:
+    """
+    Check a generic logical condition involving annotations in a Hail Table and print the results to terminal.
+
+    Displays the number of rows in the Table that match the `cond_expr` and fail to be the desired condition (`check_description`). 
+    If the number of rows that match the `cond_expr` is 0, then the Table passes that check; otherwise, it fails.
+
+    .. note::
```
gtiao, line 27: Thank you for adding this note -- it's probably something easily missed when setting up the generic field check, so it's great to have this additional reminder

PR #244, opened 2020-07-30
Update cache and setup-python Actions

nawatts:
Attempting to resolve or at least reproduce the PySpark install error in checks on #241.

https://github.com/broadinstitute/gnomad_methods/pull/241/checks?check_run_id=929532907

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,11 +12,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Setup Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: 3.7
       - name: Use pip cache
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
           path: ~/.cache/pip
           key: pip-${{ hashFiles('**/requirements*.txt') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -12,18 +12,19 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Setup Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
```
ch-kr, line 15: what is the difference between using @v1 and @v2 here?

.github/workflows/ci.yml
```diff
@@ -12,18 +12,19 @@ jobs:
...
         with:
           python-version: 3.7
       - name: Use pip cache
-        uses: actions/cache@v1
+        uses: actions/cache@v2
```
ch-kr, line 19: same question here

nawatts:
The release notes for the two updated actions can be found in their repositories:
https://github.com/actions/setup-python/releases
https://github.com/actions/cache/releases

Updating those actions was partly just to see if the PySpark installation error was reproducible on another PR. The actual fix for the issue is the second commit that installs the [wheel](https://pypi.org/project/wheel/) package. 

The error message for PySpark installation in [checks on #241](https://github.com/broadinstitute/gnomad_methods/runs/929532907) and [this PR](https://github.com/broadinstitute/gnomad_methods/runs/929746240) included `error: invalid command 'bdist_wheel'` and the wheel package contains "A setuptools extension for building wheels that provides the bdist_wheel setuptools command."

PR #249, opened 2020-08-13
VCF fix

ch-kr:
Fixes for bugs found at runtime

### DIFF
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -48,7 +48,7 @@ AS_FIELDS = [
     "AS_ReadPosRankSum",
     "AS_SOR",
     "AS_VarDP",
-    "InbreedingCoeff",
+    "AS_InbreedingCoeff",
 ]
 """
 Allele-specific variant annotations.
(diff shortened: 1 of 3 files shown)


gnomad/utils/vcf.py
```diff
@@ -583,7 +594,7 @@ def add_as_info_dict(
             ] = f"Allele-specific {first_letter}{rest_of_description}"
 
         except KeyError:
-            raise ValueError(f"{field} is not present in input info dictionary!")
+            logger.warning(f"{field} is not present in input info dictionary!")
```
jkgoodrich, line 597: Assuming this is so that you can see them all at once, which is great, but I still think we should error out so maybe add an error if any fields were missing, but print all that were missing

gnomad/utils/vcf.py
```diff
@@ -611,7 +622,7 @@ def make_vcf_filter_dict(
         },
         "InbreedingCoeff": {"Description": f"InbreedingCoeff < {inbreeding_cutoff}"},
         "RF": {
-            "Description": f"Failed random forest filtering thresholds of {snp_cutoff}, {indel_cutoff} (probabilities of being a true positive variant) for SNPs, indels"
+            "Description": f"Failed random forest filtering thresholds of {snp_cutoff}, {indel_cutoff} for SNPs, indels (probabilities of being a true positive variant)"
```
jkgoodrich, line 625: I still like "Description": f"Failed random forest filtering thresholds of {snp_cutoff} for SNVs and {indel_cutoff} for indels (probabilities of being a true positive variant)" better, but it's a really minor thing and I am happy to keep as is if you prefer

gnomad/utils/vcf.py
ch-kr: I'm not sure we should error out here with how this function is designed, since it's searching for site fields in INFO_DICT to generate the AS-version of that annotation

jkgoodrich: Ah, OK, yes that makes sense

PR #254, opened 2020-09-10
Add 1 to REF-block DP expression in impute_sex_ploidy

chrisvittal:
A REF-block may have an end equal to its position. The old code
undercounts depth by treating REF-block intervals as half open, when
they are closed. This change adds in order to get their true size.


ch-kr (approved):
looks good to me, thanks for catching that!

PR #257, opened 2020-09-16
Add Clinvar pathogenic resources

williamphu:
Added clinvar_pathogenic code to `reference_data.py` for both `resources/grch37` and `resources/grch38`.


gnomad/resources/grch37/reference_data.py
```diff
@@ -96,6 +96,23 @@
     },
 )
 
+clinvar_pathogenic = VersionedTableResource(
+    default_version="20181028",
+    versions={
+        "20181028": TableResource(
+            path="gs://gnomad-public/resources/grch37/clinvar/clinvar_pathogenic_20181028.ht",
+            import_func=import_sites_vcf,
```
jkgoodrich, line 104: It might be best to add the pathogenic filtering function here instead to directly filter from the HT. Maybe we can add the filtering function to resource utils?

gnomad/resources/grch37/reference_data.py
```diff
@@ -96,6 +96,23 @@
...
+            import_args={
+                "path": "gs://gnomad-public/resources/grch37/clinvar/clinvar_pathogenic_20181028.vcf.bgz",
```
jkgoodrich, line 106: Goes with comment above. This would keep us from needing to export the vcf and then reimport it. Actually, I just got thinking that maybe the best way to do this is have a single clinvar resource, but add a pathogenic annotation to that resource on loading. That way we only store one table, but have the pathogenic info already annotated

gnomad/resources/grch38/reference_data.py
```diff
@@ -42,6 +42,12 @@ def _import_clinvar(**kwargs) -> hl.Table:
     return clinvar
 
 
+def _import_clinvar_pathogenic(**kwargs) -> hl.Table:
```
jkgoodrich, line 45: If we did the above suggestion we actually wont need this anymore, though `_import_clinvar` would now call the function for annotating clinvar with pathogenic information

gnomad/resources/grch38/reference_data.py
```diff
@@ -123,6 +129,24 @@ def _import_clinvar(**kwargs) -> hl.Table:
     },
 )
 
+clinvar_pathogenic = VersionedTableResource(
```
jkgoodrich, line 132: same thoughts as those for GRCh37

jkgoodrich (changes requested):
Thank you so much for adding these resources William! I think the big thing to change is that it would be good to add the code that does the filtering to pathogenic variants to the common code possibly to the resource utils. The other thought is that we can maybe change this to only store the clinvar table, but with an annotation for the path/likely path variants that is annotated on import. However, I have already successfully used this resource to look at the pathogenic variants in the new gnomAD callset and it worked very well, so thank you so much for getting this resource in!!

williamphu:
Just updated the code to do filtering instead of importing a precompiled table - run `from gnomad.resources.[genome_build].reference_data import get_clinvar_pathogenic` to import the function.

gnomad/resources/grch37/reference_data.py
```diff
@@ -270,3 +270,62 @@ def get_truth_ht() -> hl.Table:
         .repartition(200, shuffle=False)
         .persist()
     )
+
+
+def get_clinvar_pathogenic(version: str = None) -> hl.Table:
```
jkgoodrich, line 275: I think we might want to rethink this a little. Currently the same function is basically copied for 38 and 37, but the actual filtering really doesn't need to be in each resource file. I wonder if we should add a function to utils/filtering and call that instead. I am interested to hear what @mike-w-wilson  thinks too since he handled all the reorganization. So I am thinking add something like this to filtering.

```
def filter_to_clinvar_pathogenic(
    clnrevstat_expr: hl.ArrayExpression, 
    clnsig_expr: hl.ArrayExpression, 
    clnsigconf_expr: hl.ArrayExpression,
    remove_no_assertion: bool = True,
    remove_conflicting: bool = True,
) -> Union[hl.MatrixTable, hl.Table]:
    """
    docstring...
    """
    logger.info(f"Found {t.count_rows() if isinstance(t, hl.MatrixTable) else t.count()} variants before filtering")
    path_expr = t.info.CLNSIG.map(lambda x: x.lower()).map(lambda x: x.contains("pathogenic")).any(lambda x: x)
    
    if remove_no_assertion:
        path_expr = path_expr & (hl.set(t.info.CLNREVSTAT).intersection(no_star_assertions).length() == 0) 
        no_star_assertions = hl.literal(
            {
                "no_assertion_provided",
                "no_assertion_criteria_provided",
                "no_interpretation_for_the_single_variant",
            }
        )
    
    if remove_conflicting:
        path_expr = path_expr & hl.is_missing(t.info.CLNSIGCONF)

    if isinstance(t, hl.MatrixTable):
        t = t.filter_rows(path_expr)
    else:
        t = t.filter(path_expr)

    logger.info(
        f"Found {t.count_rows() if isinstance(t, hl.MatrixTable) else t.count()} variants after filtering to clinvar pathogenic variants"
    )

    return t
```

Use in this function would then be:
```
clinvar_path_ht = filter_to_clinvar_pathogenic(
    clinvar_ht,
    clinvar_ht.info.CLNREVSTAT, 
    clinvar_ht.info.CLNSIG, 
    clinvar_ht.info.CLNSIGCONF,
)
```

gnomad/resources/grch37/reference_data.py
```diff
@@ -270,3 +270,62 @@ def get_truth_ht() -> hl.Table:
...
+    """
+    Returns a table that filters the clinvar data to pathogenic and likely pathogenic variants.
+    Accepts an optional string corresponding to the version of clinvar. If no value is given, defaults to the default version of clinvar.
+
+    return: A table with the corresponding version of clinvar filtered.
+    """
+    if version:
+        if not (isinstance(version, str)):
+            raise TypeError(
```
jkgoodrich, line 284: Use f string instead of `+`

gnomad/resources/grch37/reference_data.py
```diff
@@ -270,3 +270,62 @@ def get_truth_ht() -> hl.Table:
...
+                + str(type(clinvar.default_version))
+                + ", "
+                + "given "
+                + str(type(version))
+                + "."
+            )
+        if version in clinvar.versions:
+            clinvar_ht = clinvar.versions[version].ht()
+        else:
+            raise ValueError(version + " is not a supported version.")
```
jkgoodrich, line 295: Use f string instead of `+`

gnomad/resources/grch37/reference_data.py
```diff
@@ -270,3 +270,62 @@ def get_truth_ht() -> hl.Table:
...
+    else:
+        print("Using default clinvar version: " + clinvar.default_version + ".")
```
jkgoodrich, line 297: Use f string instead of `+` and `logger` instead of `print`

gnomad/resources/grch38/reference_data.py
```diff
@@ -251,3 +251,62 @@ def get_truth_ht() -> Table:
         .repartition(200, shuffle=False)
         .persist()
     )
+
+
+def get_clinvar_pathogenic(version: str = None) -> Table:
```
jkgoodrich, line 256: Same comments as in `grch37`

jkgoodrich (changes requested):
Still thinking about the best way to handle this. See main comment for an idea. It would be good to get @mike-w-wilson to provide thoughts on it too

gnomad/resources/grch37/reference_data.py
mike-w-wilson: I like this idea best and I know the above function needs to be cleaned up a bit but I think adding defaults for 
```
clnrevstat_expr: hl.ArrayExpression, 
clnsig_expr: hl.ArrayExpression, 
clnsigconf_expr: hl.ArrayExpression,``` 

in the final version would make sense since the I think we can expect us/users to annotate a HT/MT with our clinvar resource. Alternatively, we can keep them optional and adjust the code to not rely on them.

Comment (jkgoodrich) [reply gnomad/resources/grch37/reference_data.py:275]: @mike-w-wilson I fully agree it would be great to have defaults. Probably a question I should know the answer to, but can we add defaults for hail expressions? I have always switched to strings of column names to add defaults which works, but I do like keeping it as expressions if possible.

Comment (mike-w-wilson) [reply gnomad/resources/grch37/reference_data.py:275]: @jkgoodrich,  I've been thinking about how to best implement this this afternoon and haven't settled on anything I like. I guess I should have a better understanding of how hail handles these things esp. lazy evaluation vs repeated evaluations and what triggers what. Perhaps this is a better question for zulip or discuss.hail.is

Comment (jkgoodrich) [reply gnomad/resources/grch37/reference_data.py:275]: great idea Mike, @williamphu can you post on discuss.hail.is. If you give a link to the discussion (which is public) they can see what we are trying to do.

Comment (williamphu) [reply gnomad/resources/grch37/reference_data.py:275]: Yep, I support extracting the code out to a more general function. The hail discussion is here: https://discuss.hail.is/t/pass-default-for-hail-expressions-in-function/1874.

To double check, would `filter_to_clinvar_pathogenic` only be applied to clinvar resources (expect users to pass in a clinvar), or would any table with corresponding columns be filterable?

Comment (jkgoodrich) [reply gnomad/resources/grch37/reference_data.py:275]: I think it is likely to only be applied to clinvar resources, but the above suggestion would make it more general and could then be applied to any table with the needed columns.

Comment (williamphu) [reply gnomad/resources/grch37/reference_data.py:275]: Based on the comments from the hail discuss, we'll use the string name of the column for the defaults.

Comment (williamphu) [reply gnomad/resources/grch37/reference_data.py]: New code pushed to reflect this restructuring.

File: gnomad/utils/filtering.py [hunk 10]
```diff
@@ -256,3 +256,66 @@ def subset_samples_and_variants(
...
+    remove_no_assertion: bool = True,
+    remove_conflicting: bool = True,
+) -> Union[hl.MatrixTable, hl.Table]:
+    """
+    Returns a MatrixTable or Table that filters the clinvar data to pathogenic and likely pathogenic variants.
+
+    :param: mt: Input dataset that contains clinvar data, could either be a MatrixTable or Table. Defaults to gnomad's grch38 clinvar version.
+    :param clnrevstat: The field string for the expression that contains the review status of the clinical significance of clinvar variants.
+    :param clnsig: The field string for the expression that contains the clinical signifcance of the clinvar variant.
+    :param clnsigconf: The field string for the expression that contains if there is conflicting clinical significance for the variant. Undefined if there is no conflicting signficance.
```
jkgoodrich, line 275: The field string for the expression that contains the conflicting clinical significance values for the variant. For variants with no conflicting significance, this field should be undefined.

jkgoodrich (changes requested):
Just a few small things

gnomad/utils/filtering.py
```diff
@@ -256,3 +256,67 @@ def subset_samples_and_variants(
...
+
+    :param: t: Input dataset that contains clinvar data, could either be a MatrixTable or Table.
+    :param clnrevstat_field: The field string for the expression that contains the review status of the clinical significance of clinvar variants.
+    :param clnsig_field: The field string for the expression that contains the clinical signifcance of the clinvar variant.
+    :param clnsigconf_field: The field string for the expression that contains the conflicting clinical significance values for the variant. For variants with no conflicting significance, this field should be undefined.
+    :param remove_no_assertion: Flag for removing entries in which the clnrevstat (clinical significance) has no assertions (zero stars).
+    :param remove_conflicting: Flag for removing entries with conflicting clinical interpretations.
+    :return: Filtered MatrixTable or Table
+
+    from gnomad.resources.grch38.reference_data import clinvar
```
jkgoodrich, line 280: Move this up, so full docstring would be:

```
    Returns a MatrixTable or Table that filters the clinvar data to pathogenic and likely pathogenic variants.

    Example use:

    .. code-block:: python

        from gnomad.resources.grch38.reference_data import clinvar
        clinvar_ht = clinvar.ht()
        clinvar_ht = filter_to_clinvar_pathogenic(clinvar_ht)

    :param: t: Input dataset that contains clinvar data, could either be a MatrixTable or Table.
    :param clnrevstat_field: The field string for the expression that contains the review status of the clinical significance of clinvar variants.
    :param clnsig_field: The field string for the expression that contains the clinical signifcance of the clinvar variant.
    :param clnsigconf_field: The field string for the expression that contains the conflicting clinical significance values for the variant. For variants with no conflicting significance, this field should be undefined.
    :param remove_no_assertion: Flag for removing entries in which the clnrevstat (clinical significance) has no assertions (zero stars).
    :param remove_conflicting: Flag for removing entries with conflicting clinical interpretations.
    :return: Filtered MatrixTable or Table
```

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * Fix for error in `compute_stratified_sample_qc` where `gt_expr` caused error [(#259)](https://github.com/broadinstitute/gnomad_methods/pull/259)
 * Add reference genome to call of `has_liftover` in `get_liftover_genome` [(#259)](https://github.com/broadinstitute/gnomad_methods/pull/259)
 * Added fix for MQ calculation in `_get_info_agg_expr`, switched `RAW_MQ` and `MQ_DP` in calculation [(#262)](https://github.com/broadinstitute/gnomad_methods/pull/262)
+* Add importable method for filtering clinvar to pathogenic sites for grch37 and grch38 [(#257)](https://github.com/broadinstitute/gnomad_methods/pull/257)
```
jkgoodrich, line 20: Maybe just "Add importable method for filtering clinvar to pathogenic sites"

jkgoodrich (approved):
I think just two small changes and then this is good to go

PR #258, opened 2020-09-16
Reduce memory usage of `get_as_info_expr`.

tpoterba:
Using a range inside of an agg_array allocates that range for each
entry. At a site with 4567 alleles, this led to ~12G being allocated.

Doesn't hugely affect runtime, though.

### DIFF
--- a/gnomad/utils/sparse_mt.py
+++ b/gnomad/utils/sparse_mt.py
@@ -319,11 +319,16 @@ def get_as_info_expr(
     if "AS_SB" in agg_expr:
         agg_expr["AS_SB_TABLE"] = agg_expr.pop("AS_SB")
 
+    if 'alt_alleles_range_array' not in mt.row or mt['alt_alleles_range_array'].dtype != hl.dtype('array<int32>'):
+        msg = f"'get_as_info_expr' expects a row field 'alt_alleles_range_array' of type array<int32>"
+        logger.error(msg)
+        raise RuntimeError(msg)
+
     # Modify aggregations to aggregate per allele
     agg_expr = {
         f: hl.agg.array_agg(
             lambda ai: hl.agg.filter(mt.LA.contains(ai), expr),
-            hl.range(1, hl.len(mt.alleles)),
+            mt.alt_alleles_range_array,
         )
         for f, expr in agg_expr.items()
     }
@@ -430,6 +435,7 @@ def default_compute_info(
     """
     # Move gvcf info entries out from nested struct
     mt = mt.transmute_entries(**mt.gvcf_info)
+    mt = mt.annotate_rows(alt_alleles_range_array=hl.range(1, hl.len(mt.alleles)))
 
     # Compute AS info expr
     info_expr = get_as_info_expr(mt)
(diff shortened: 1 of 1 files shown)


gnomad/utils/sparse_mt.py
```diff
@@ -319,11 +319,16 @@ def get_as_info_expr(
     if "AS_SB" in agg_expr:
         agg_expr["AS_SB_TABLE"] = agg_expr.pop("AS_SB")
 
+    if 'alt_alleles_range_array' not in mt.row or mt['alt_alleles_range_array'].dtype != hl.dtype('array<int32>'):
```
lfrancioli, line 322: I think I'd like it better to have this as an argument with a default value. Or alternatively mention what this array is in the doc. Happy to modify but depends what the team thinks in terms of interface.

jkgoodrich, line 322: @tpoterba Sorry if this is a stupid question, what is the motivation for having this instead of the same fix as in the `gnomad_qc` repo where you added.

```
mt = mt.annotate_rows(alt_alleles_range_array=hl.range(1, hl.len(mt.alleles)))
```
And then just drop the annotation after we use it rather than making it already be on the MT?

gnomad/utils/sparse_mt.py
```diff
@@ -280,6 +280,7 @@ def get_as_info_expr(
     array_sum_agg_fields: Union[
         List[str], Dict[str, hl.expr.ArrayNumericExpression]
     ] = INFO_ARRAY_SUM_AGG_FIELDS,
+    alt_alleles_range_array: str = "alt_alleles_range_array",
```
nawatts, line 283: For consistency with other arguments that take field names, maybe name this `alt_alleles_range_array_field`?

gnomad/utils/sparse_mt.py
```diff
@@ -319,11 +322,18 @@ def get_as_info_expr(
     if "AS_SB" in agg_expr:
         agg_expr["AS_SB_TABLE"] = agg_expr.pop("AS_SB")
 
+    if alt_alleles_range_array not in mt.row or mt[
+        alt_alleles_range_array
+    ].dtype != hl.dtype("array<int32>"):
+        msg = f"'get_as_info_expr' expected a row field '{alt_alleles_range_array}' of type array<int32>"
+        logger.error(msg)
+        raise RuntimeError(msg)
```
nawatts, line 330: Nitpick, but `ValueError` seems like it would be slightly more appropriate here.
https://docs.python.org/3/library/exceptions.html#ValueError

gnomad/utils/sparse_mt.py
```diff
@@ -319,11 +322,18 @@ def get_as_info_expr(
...
+
     # Modify aggregations to aggregate per allele
     agg_expr = {
         f: hl.agg.array_agg(
             lambda ai: hl.agg.filter(mt.LA.contains(ai), expr),
-            hl.range(1, hl.len(mt.alleles)),
+            mt.alt_alleles_range_array,
```
nawatts, line 336: Should this be `mt[alt_alleles_range_array]`?

jkgoodrich: Yes, thank you!!

nawatts (approved):

PR #259, opened 2020-09-17
Sample qc bug fixes

jkgoodrich:


gnomad/utils/liftover.py
```diff
@@ -52,7 +52,7 @@ def get_liftover_genome(
         chain = GRCH37_to_GRCH38_CHAIN
 
     logger.info("Adding liftover chain to input build...")
-    if source.has_liftover():
+    if source.has_liftover(input_build):
```
ch-kr, line 55: wouldn't this have to be `target` instead?

gnomad/sample_qc/filtering.py
```diff
@@ -185,22 +185,22 @@ def compute_stratified_sample_qc(
...
     :param tmp_ht_prefix: Optional path prefix to write the intermediate strata results to (recommended for larger datasets)
-    :param gt_expr: Optional entry field storing the genotype (if not specified, then it is assumed that it is stored in mt.GT)
+    :param gt_col: Name of entry field storing the genotype. Default: 'GT'
     :return: Sample QC table, including strat-specific numbers
     """
     mt = mt.select_rows(**strata)
 
-    if gt_expr is not None:
-        mt = mt.select_entries(GT=gt_expr)
+    if gt_col is not None:
```
ch-kr, line 202: would this not work if you set gt_expr to None as a default?

ch-kr (approved):

PR #261, opened 2020-09-18
Removes VersionedVersionedTableResource class

mike-w-wilson:
Removes VersionedVersionedTableResource class and also sorts imports.


nawatts (approved):
Since there will be no change between published versions, it would be better to remove mention of VersionedVersionedTableResource from the changelog instead of including both "Added VersionedVersionedTableResource" and "Removed VersionedVersionedTableResource".

Issue #263, opened 2020-09-24
Create resources PR for new Amazon resources


gtiao:
We should discuss creating a resources library for gnomAD files hosted on Amazon for any users that are writing pipelines on AWS

nawatts:
As long as the buckets on each cloud provider keeps the same structure, resource classes could accept a cloud provider and path (relative to the bucket root) and construct the absolute URL based on those by joining the appropriate URL schema and bucket name for the selected provider and the path (which does not change between providers).

nawatts:
Closing this as a duplicate of #255.

Issue #255, opened 2020-09-11
Update resource paths to new Google/AWS/Azure public hosted data


gtiao:
Since we'll end up with 3 versions (one on each of the major cloud providers) for each public file, we'll need to implement a new structure for specifying which cloud version to use

nawatts:
#373 added support for reading resources from the Google Cloud Public Datasets bucket.

See https://github.com/broadinstitute/gnomad_methods/pull/373/commits/f44ae96abcaf958800f41746efebae5703fcc0a5 for an example of how to add support for AWS and Azure.

nawatts:
It's not immediately clear what the URL scheme for files on AWS would be.

https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/index.html and https://spark.apache.org/docs/3.1.1/cloud-integration.html#recommended-settings-for-writing-to-object-stores suggest `s3a://`.

https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/index.html#Other_S3_Connectors mentions an `s3://` client from Amazon EMR.

jkgoodrich:
Need to update gnomad_methods resource code for AWS and Azure: https://github.com/broadinstitute/gnomad_methods/blob/master/gnomad/resources/resource_utils.py

PR #265, opened 2020-10-01
Support multiple resource buckets and switch to Google Cloud Public Datasets bucket

### LINKED ISSUE #264: Update gnomAD resources to reflect free Google Cloud-hosted paths
Prefixes should change from `gs://gnomad-public/` or `gs://gnomad-public-requester-pays/` to `gs://gcp-public-data--gnomad`

nawatts:
This adds support for multiple resource buckets in order to provide access to resources hosted on different cloud providers through our resources framework (#255, #263, #264). This also sets the default resource bucket to the Google Cloud Public Datasets bucket: `gs://gcp-public-data--gnomad` (resolves #264).

Some design goals:
1. Do not change interfaces. Everyone has more pressing things to do right now than update pipelines with changes to resources.
2. The resource "directory" structure is the same in each bucket. Because of this, we can define resource paths in one place and parameterize the URL schema (`gs://`, `s3://`, etc.) and the bucket name instead of having different resource modules for each cloud provider.
3. Easily select a resource bucket to use for all resources in a pipeline.

Because our resource modules export objects, my [initial idea](https://github.com/broadinstitute/gnomad_methods/issues/263#issuecomment-698397151) of adding a cloud provider parameter to Resource classes was unworkable since it would require changing the library's interface (against goal 1) or adding separate modules for each provider (against goals 2 and 3).

To meet goal 2, this changes the `path` attribute of `Resource` objects from an absolute URL to a relative (within a bucket) path and adds a `resources_root` parameter to `Resource` methods that read the resource (`TableResource.ht`, `PedigreeResource.pedigree`, etc.) and `import_resource`. This provides flexibility: users can mix and match resource buckets within a pipeline (for example, if a file in `gnomad-public` hasn't yet been copied to `gcp-public-data--gnomad`) or use resource buckets unknown to us (for example, if someone sets up a copy of `gnomad-public` in Australia). `resources_root` is optional (goal 1) and the expectation is that most users will leave it unspecified.

Setting a default behavior is not as simple as setting `"gs://gnomad-public"` as the default value for `resources_root`. This partly because our resource files are split between two buckets: `gnomad-public` and `gnomad-public-requester-pays` and partly because of goal 3: a pipeline using resources from AWS shouldn't have to pass `resources_root` every time a resource is read.

Thus, `Resource` methods call `get_resource_url` with the `Resource`'s `path` attribute, the given `resources_root`, and a new `Resource` attribute: `gnomad_bucket`. The `gnomad_bucket` attribute is necessary to distinguish resources stored in `gnomad-public` vs `gnomad-public-requester-pays`.

In turn, `get_resource_url` looks at a new setting: the "default resource provider". This controls how resource URLs are constructed when no `resources_root` is provided (which is expected to usually be the case). With this, the resource bucket to use for all resources in a pipeline can be set by changing the `default_resource_provider` attribute on `gnomad.resources.config.gnomad_resource_configuration`. And support for other cloud providers can be added by adding to the `GnomadResourceProvider` enum and updating `get_resource_url` to accordingly.

### Import paths

The `gnomad/resources/import_resources.py` script is used to convert VCF, TSV, etc. files into Hail formats. The `import_args` attribute on `Resource` is not aware of `resources_root` and thus URLs there for files in `gnomad-public` or `gnomad-public-requester-pays` are not changed. This is not likely to be a problem, since we will have limited access to the various cloud providers' buckets and thus the workflow will probably be to import resources in our own buckets and then copy from there to other buckets. Likewise, in `import_resources.py`, the `resources_root` is hard coded to `gs://gnomad-public`.

### Usage

To load resources from the Google Cloud Public Datasets bucket:

```python
from gnomad.resources.grch37 import gnomad

ds = gnomad.public_release("exomes").ht()
```

To load resources from gnomAD buckets:

```python
from gnomad.resources.grch37 import gnomad
from gnomad.resources import GnomadResourceProvider, gnomad_resource_configuration

gnomad_resource_configuration.default_resource_provider = GnomadResourceProvider.GNOMAD

ds = gnomad.public_release("exomes").ht()
```

or

```python
from gnomad.resources.grch37 import gnomad

ds = gnomad.public_release("exomes").ht(resources_root="gs://gnomad-public-requester-pays")
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@
 * Added function, `default_generate_gene_lof_matrix`, to generate gene matrix [(#290)](https://github.com/broadinstitute/gnomad_methods/pull/290)
 * Added function `default_generate_gene_lof_summary` to summarize gene matrix results [(#292)](https://github.com/broadinstitute/gnomad_methods/pull/292)
 * Modified `create_truth_sample_ht` to add adj annotation information in the returned Table if present in the supplied MatrixTables [(#300)](https://github.com/broadinstitute/gnomad_methods/pull/300)
+* Added support for loading resources from alternate sources [(#265)](https://github.com/broadinstitute/gnomad_methods/pull/265)
+* Default to loading resources from Google Cloud Public Datasets [(#265)](https://github.com/broadinstitute/gnomad_methods/pull/265)
 
 ## Version 0.4.0 - July 9th, 2020
 
--- a/docs/examples/vep.rst
+++ b/docs/examples/vep.rst
@@ -7,7 +7,7 @@ tied to a specific reference genome.
 
 .. code-block:: shell
 
-   hailctl dataproc start cluster-name --vep GRCh37 --packages gnomad --requester-pays-allow-buckets gnomad-public-requester-pays
+   hailctl dataproc start cluster-name --vep GRCh37 --packages gnomad
 
 .. note::
 
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -6,10 +6,10 @@ Getting Started
     pip install hail
 
 2. Use ``hailctl`` to start a `Google Dataproc <https://cloud.google.com/dataproc/>`_ cluster with the
-   ``gnomad`` package installed and allow reading from the ``gnomad-public-requester-pays`` storage bucket
-   (see `Hail on the Cloud <https://hail.is/docs/0.2/hail_on_the_cloud.html>`_ for more detail on hailctl)::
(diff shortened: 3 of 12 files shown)


nawatts:
Realized that, as is, this would break uses of resource classes in gnomad_qc.

jkgoodrich:
ooh, didn't know about marking PRs as draft, thanks Nick!

gtiao:
What do we need to do to make it work?

nawatts:
I have a different version of this working but it requires #359.

nawatts:
Replaced by #373

PR #267, opened 2020-10-05
Add dbSNP B154 to resources

jkgoodrich:
Small PR to add new dbSNP release to resources

### DIFF
--- a/gnomad/resources/grch38/reference_data.py
+++ b/gnomad/resources/grch38/reference_data.py
@@ -152,7 +152,7 @@ dbsnp = VersionedTableResource(
                 "min_partitions": 400,
                 "reference_genome": "GRCh38",
             },
-        )
+        ),
     },
 )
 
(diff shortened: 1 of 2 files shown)


gnomad/resources/grch38/reference_data.py
```diff
@@ -138,7 +139,20 @@ def _import_clinvar(**kwargs) -> hl.Table:
...
+            import_args={
+                "path": "gs://gnomad-public/resources/grch38/dbsnp/dbsnp_b154_grch38_all_GCF_000001405.38_20200514.vcf.bgz",
+                "header_file": "gs://gnomad-public/resources/grch38/dbsnp/dbsnp_b154_grch38_all_GCF_000001405.38_20200514.vcf.header",
+                "force_bgz": True,
+                "contig_recoding": DBSNP_B154_CHR_CONTIG_RECODING,
+                "skip_invalid_loci": True,
+                "min_partitions": 400,
+                "reference_genome": "GRCh38",
+            },
+        ),
```
mike-w-wilson, line 155: In other repos, we've put the latest version at the top of the resource dictionary, closest to the default version. I think flipping this with b151 to match that set up would be good.

mike-w-wilson (approved):
One small change: just the ordering in the dictionary but other than that looks good to me!

PR #279, opened 2021-01-11
densify_sites update

ch-kr:
Sparse MTs were previously only keyed by locus by default. They are now keyed by locus and alleles by default


jkgoodrich (commented):
Should we make it an option with default False incase using an older MT?

jkgoodrich (commented):
Will both now be keyed by locus and alleles? will this always work?
`last_END_positions_ht[sites_ht.key].last_END_position`
what if sites_ht has different alleles but you still want to keep the locus?

ch-kr:
that's a good point. what if I change `compute_last_ref_block_end` instead to always key by locus?

jkgoodrich:
Yeah both would need to be re_keyed by locus, so I think that will work

jkgoodrich (approved):
LGTM

PR #281, opened 2021-01-13
Add sor

jkgoodrich:
Adding in SOR calculation to info HT creation.

### DIFF
--- a/gnomad/utils/sparse_mt.py
+++ b/gnomad/utils/sparse_mt.py
@@ -2,7 +2,12 @@ import logging
 from typing import Dict, List, Optional, Union
 
 import hail as hl
-from gnomad.utils.annotations import fs_from_sb, get_adj_expr, get_lowqual_expr
+from gnomad.utils.annotations import (
+    fs_from_sb,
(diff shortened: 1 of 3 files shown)


gnomad/utils/sparse_mt.py
```diff
@@ -401,6 +409,9 @@ def get_site_info_expr(
     # This is done outside of _get_info_agg_expr as the behavior is different in site vs allele-specific versions
     agg_expr["FS"] = fs_from_sb(agg_expr["SB"])
 
+    # Add SOR if SB is present
```
mike-w-wilson, line 412: Should there be a check here and also in line 408? Something similar to the check for DP in L395. The comment makes it seem like SB may not be in the entry or gvcf info fields but the module will fail in 413 without an SB field. Or maybe the comment needs to be updated to remove the "if SB..."

mike-w-wilson (approved):
Just a few questions regarding the docs/comments and one possible conditional. For docs, I'm not sure how descriptive to be here but made a couple of suggestions on the wordier side...

PR #288, opened 2021-01-28
Fixes to variant qc evaluation

jkgoodrich:
First pass at a PR to fix variant QC evaluation bugs found by Grace and revert back to ranking then binning instead of using approx quantile.

### DIFF
--- a/gnomad/variant_qc/evaluation.py
+++ b/gnomad/variant_qc/evaluation.py
@@ -48,6 +48,7 @@ def compute_ranked_bin(
         and singleton specific binning, the following could be used:
 
         .. code-block:: python
+        
             bin_expr={
                 'biallelic_bin': ~ht.was_split,
                 'singleton_bin': ht.singleton
--- a/gnomad/variant_qc/evaluation.py
+++ b/gnomad/variant_qc/evaluation.py
@@ -25,7 +25,7 @@ def compute_ranked_bin(
 
     The bin is computed by dividing the `score_expr` into `n_bins` bins containing approximately equal numbers of elements.
     This is done by ranking the rows by `score_expr` (and a random number in cases where multiple variants have the same score)
-    and then assigning the variant to a bin based on it's ranking.
+    and then assigning the variant to a bin based on its ranking.
 
     If `compute_snv_indel_separately` is True all items in `bin_expr` will be stratified by snv / indels for the ranking and
     bin calculation. Because SNV and indel rows are mutually exclusive, they are re-combined into a single annotation. For
@@ -48,7 +48,7 @@ def compute_ranked_bin(
         and singleton specific binning, the following could be used:
 
         .. code-block:: python
-        
+
             bin_expr={
                 'biallelic_bin': ~ht.was_split,
                 'singleton_bin': ht.singleton
@@ -56,7 +56,7 @@ def compute_ranked_bin(
 
     :param ht: Input Table
     :param score_expr: Expression containing the score
-    :param bin_expr: Specific row grouping(s) to perform ranking and binning on to be computed (see note)
+    :param bin_expr: Specific row grouping(s) to perform ranking and binning on (see note)
     :param compute_snv_indel_separately: Should all `bin_expr` items be stratified by SNVs / indels
     :param n_bins: Number of bins to bin the data into
     :param desc: Whether to bin the score in descending order
@@ -82,7 +82,7 @@ def compute_ranked_bin(
     )
 
     logger.info(
-        "Sorting the HT by score_expr followed by a random float between 0 and 1."
+        "Sorting the HT by score_expr followed by a random float between 0 and 1. "
         "Then adding a row index per grouping defined by bin_expr..."
     )
     bin_ht = bin_ht.order_by("_score", "_rand")
@@ -109,7 +109,7 @@ def compute_ranked_bin(
         )
     )
 
-    logger.info(f"Binning ranked rows into {n_bins}...")
+    logger.info(f"Binning ranked rows into {n_bins} bins...")
     bin_ht = bin_ht.select(
         "snv",
         **{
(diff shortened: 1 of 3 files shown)


jkgoodrich:
The current PR replaces the quantile approx functionality, but I am also open to adding it in with a warning about unequal binning when there are duplicate boundaries

gnomad/variant_qc/evaluation.py
```diff
@@ -116,108 +74,78 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
-                bin_ht.bin_stats[bin_id].global_bin_indices[bin_ht[bin_id]]
-                + hl.int(
-                    hl.rand_unif(
-                        0, bin_ht.bin_stats[bin_id].merged_bins[bin_ht[bin_id]] + 1
+            bin_id: hl.int(
+                hl.floor(
+                    (
+                        n_bins
+                        * (
+                            bin_ht[f"{bin_id}_rank"]
```
gtiao, line 121: There are actually 101 bins the way the code is written, because the largest rank (say it's 1002 in a given grouping) will give a value of 1 * n_bins = 100 (because the rank 1002 will be divided by the total number of variants in the grouping, 1002). Basically, you're starting a new bin for the single final variant in each grouping.

I think if you want to only have 100 bins, you'll need to subtract 1 off the `{bin_id}_rank` in line 121 before dividing by `hl.float64(bin_ht.bin_group_variant_counts[bin_id]` on line 122. This will handle the other end (the first variant in the grouping) fine, because `hl.floor()` will still return zero for the first variant.

gnomad/variant_qc/evaluation.py
```diff
@@ -116,108 +74,78 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
+                )
             )
             for bin_id in bin_expr
         },
     )
 
     if desc:
         bin_ht = bin_ht.annotate(
-            **{bin_id: n_bins - bin_ht[bin_id] for bin_id in bin_expr}
+            **{bin_id: n_bins - bin_ht[bin_id] + 1 for bin_id in bin_expr}
```
gtiao, line 134: Note that currently, because there are 101 bins labeled (1,101) in ascending order, this code gives bins labeled (100,0) in descending order. Once the previous comment is addressed and we have 100 bins labeled (1,100) in ascending order, this code will work as intended and give bins labeled (100,1) in descending order.

gnomad/variant_qc/evaluation.py
```diff
@@ -116,108 +74,78 @@ def quantiles_to_bin_boundaries(quantiles: List[int]) -> Dict:
...
         )
 
     # Because SNV and indel rows are mutually exclusive, re-combine them into a single bin.
-    # Update the global bin_stats struct to reflect the change in bin names in the table
+    # Update the global bin_group_variant_counts struct to reflect the change in bin names in the table
     if compute_snv_indel_separately:
-        bin_expr_no_snv = {bin_id.rsplit("_", 1)[0] for bin_id in bin_ht.bin_stats}
+        bin_expr_no_snv = {
+            bin_id.rsplit("_", 1)[0] for bin_id in bin_ht.rank_variant_counts
```
gtiao, line 141: I don't see this annotation defined in this script -- is this a field that's present in the input `ht`?

gnomad/variant_qc/evaluation.py
```diff
@@ -242,7 +170,7 @@ def compute_grouped_binned_ht(
     bin_ht: hl.Table, checkpoint_path: Optional[str] = None,
 ) -> hl.GroupedTable:
     """
-    Groups a Table that has been annotated with bins based on quantiles (`compute_quantile_bin` or
+    Groups a Table that has been annotated with bins based on quantiles (`compute_ranked_bin` or
```
gtiao, line 173: This function needs to be adjusted since `bin_stats` no longer exists as a global annotation (e.g., line 191). And does it still make sense to name the annotation on line 188 `quantile_bins`, now that we're not using a quantile approach? Might be less confusing to name it something else

gtiao (changes requested):
Overall this is much easier and simpler to read -- thanks for putting this in so quickly!

gnomad/variant_qc/evaluation.py
jkgoodrich: I'm not sure I agree, the rank is actually 0 to (n_variants)

Notebook test:
[cleanup_new_binning_function.html.zip](https://github.com/broadinstitute/gnomad_methods/files/5895478/cleanup_new_binning_function.html.zip)

gnomad/variant_qc/evaluation.py
jkgoodrich: same comment as above

gnomad/variant_qc/evaluation.py
jkgoodrich: oops, I changed the name after testing, but forgot one apparently

gnomad/variant_qc/evaluation.py
jkgoodrich: Thank you, will make changes now

gnomad/variant_qc/evaluation.py
gtiao: Can you show what the `{bin_id}_rank` is in the notebook, and not just the final bin assignment? I agree if the rank is 0 to (n_variants - 1) then the code as written is fine (basically the ranking already has 1 subtracted off it). It's just that I assumed `hl.scan.count_where(bin_ht[f"_filter_{bin_id}"])` would be an actual count, which starts a 1 and not 0.

gtiao (commented):
Adjustments all look good, just want to understand how the ranking code works

gnomad/variant_qc/evaluation.py
jkgoodrich: Screen Shot 2021-01-29 at 2 09 05 PM
Screen Shot 2021-01-29 at 2 09 12 PM

jkgoodrich: I thought it was too at first, but got the wrong binning when I tested so changed it

gtiao: Thanks! I wonder if the behavior changes when it's a `scan` -- I know the sum mechanism for `hl.scan()` doesn't count the current row, so maybe that's what's happening here.

In any case, the code is working as intended, so no changes required.

gtiao (approved):

PR #289, opened 2021-01-29
Expanded summary stat counts and added AC summary

ch-kr:


gnomad/assessment/summary_stats.py
```diff
@@ -87,6 +96,64 @@ def get_summary_counts_dict(
...
+
+    .. warning:: 
+        Assumes `allele_expr` contains only two variants (multi-allelics have been split).
+
+    :param allele_expr: ArrayExpression containing alleles.
+    :param lof_expr: StringExpression containing LOFTEE annotation.
+    :param no_lof_flags_expr: BooleanExpression indicating whether LoF variant has any flags.
+    :return: Dict of variant categories and their total allele counts.
+    """
+    logger.warning("This function expects that multi-allelic variants have been split!")
```
konradjk, line 141: You could confirm this with a `hl.agg.max(hl.len(allele_expr))`

konradjk (approved):

gnomad/assessment/summary_stats.py
```diff
@@ -197,6 +197,11 @@ def get_summary_counts(
     :param index: Which index of freq_expr to use for annotation. Default is 0. 
     :return: Table grouped by frequency bin and aggregated across summary count categories. 
 	"""
+    logger.info("Checking if multi-allelic variants have been split...")
+    max_alleles = hl.agg.max(hl.len(ht.alleles))
```
konradjk, line 201: You'd have to execute this in an aggregate to check its length

konradjk (changes requested):

gnomad/assessment/summary_stats.py
```diff
@@ -95,37 +162,46 @@ def get_summary_counts(
...
+    :param freq_field: Name of field in HT containing frequency annotation (array of structs). Default is "freq".
+    :param filter_field: Name of field in HT containing variant filter information. Default is "filters".
+    :param filter_decoy: Whether to filter decoy regions. Default is False.
+    :param index: Which index of freq_expr to use for annotation. Default is 0. 
+    :return: Table grouped by frequency bin and aggregated across summary count categories. 
 	"""
+    logger.info("Checking if multi-allelic variants have been split...")
+    max_alleles = ht.aggregate(hl.agg.max(hl.len(ht.alleles)))
+    if max_alleles > 2:
+        ht = hl.split_multi(ht)
```
nawatts, line 203: Is `split_multi` sufficient here? Does it split VEP consequences as well? Since the code below relies on VEP transcript consequences (for at least `get_most_severe_consequence_for_summary`), would this also need to filter VEP consequences based on `a_index`?

ch-kr: good point. I added a filter, but do you (or @konradjk) know if the VEP `allele_num`field is the same as `a_index`?

konradjk: Yep, definitely a good point. But `split_multi_hts` implements this so you can just switch to that

ch-kr: I need the hail docs back

konradjk (approved):

PR #292, opened 2021-02-02
Gene matrix summary

ch-kr:


gnomad/assessment/summary_stats.py
```diff
@@ -452,3 +451,168 @@ def default_generate_gene_lof_matrix(
...
+                num_hets=hl.agg.sum(mt.num_hets),
+                defined_sites=hl.agg.sum(mt.defined_sites),
+            )
+            .result()
+        )
+
+    ht = mt.annotate_rows(
+        **get_het_hom_summary_dict(
+            csq_name="lof",
+            csq_set=lof_csq_set,
```
konradjk, line 580: Is there a loftee filter somewhere? I don't see it in this code, but maybe it was filtered upstream. Apologies if I missed it

ch-kr: `default_generate_gene_lof_matrix` upstream filters to loftee HC and no flags by default. I can also add the option here if you think it makes sense?

konradjk: Do you run these one after the other? If so, yes. But might be worth the option too. Not sure

konradjk (approved):

PR #300, opened 2021-02-09
Keep adj annotation if present in create_truth_sample_ht

jkgoodrich:


ch-kr (approved):
lgtm, though maybe this should have a changelog entry? since any truth sample hts created after this is merged will have an extra annotation

PR #301, opened 2021-02-09
Add pre-commit hook to run Black

nawatts:
There are [a lot of commits that are just running Black](https://github.com/broadinstitute/gnomad_methods/search?q=black&type=commits).

To save time, this adds a [pre-commit](https://pre-commit.com/) hook that runs Black. After setting it up (`pip install -r requirements-dev.txt`, `pre-commit install`), `git commit` will automatically run Black and the commit will be cancelled if Black made any changes.


jkgoodrich (approved):
Thank you for adding Nick!

PR #303, opened 2021-02-10
Added delimiter parameter to label functions and those that use them

mike-w-wilson:
These changes allow the user to pass a delimiter when creating and working with frequency labels. v2 used underscores while v3.1 switched to hyphens. This change will allow the code to work with either version and give us flexibility moving forward.


jkgoodrich (approved):
lgtm, thanks for adding Mike!

PR #306, opened 2021-02-16
Add `compare_row_counts` to sanity checks

jkgoodrich:
For use in UKBB QC and gnomAD QC


gnomad/assessment/sanity_checks.py
```diff
@@ -172,3 +173,18 @@ def sample_sum_check(
                 ],
                 verbose,
             )
+
+
+def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
+    """
+    Check if the row counts in two Tables are the same.
+
+    :param Table ht1: First Table to be checked
```
ch-kr, line 182: these don't need type here, right?

ch-kr (approved):
small comment, but lgtm

PR #307, opened 2021-02-18
Add sample count global in `annotate_freq`

jkgoodrich:
Requested as an addition to gnomAD release HT

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -352,6 +352,11 @@ def annotate_freq(
     Each element of the list contains metadata on a frequency stratification and the index in the list corresponds
     to the index of that frequency stratification in the `freq` row annotation.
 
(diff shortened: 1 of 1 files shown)


gnomad/utils/annotations.py
```diff
@@ -487,14 +492,20 @@ def annotate_freq(
     )
 
     # Annotate columns with group_membership
+    freq_sample_count = mt.aggregate_cols(
```
mike-w-wilson, line 495: Sorry can't comment on the L494 in github but code the current comment on L494 be moved to L497 and updated for the freq_sample_count? Or removed?

mike-w-wilson (approved):
Two minor things. I cannot comment on it within the PR but should the general description of the method be updated to include the counts(L326/327)? Other than that, this LGTM.

PR #339, opened 2021-02-22
Resolve Pylint errors

nawatts:
On #265, Pylint threw some errors that appear to be false positives (on code that was not changed in the PR).

https://github.com/broadinstitute/gnomad_methods/pull/265/checks?check_run_id=1954044098

The `invalid-unary-operand-type` error has come up before: https://the-tgg.slack.com/archives/C0295AMDW/p1608218693011800

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -56,6 +56,8 @@ def pop_max_expr(
     :return: Popmax struct
     """
     _pops_to_exclude = hl.literal(pops_to_exclude)
+
+    # pylint: disable=invalid-unary-operand-type
     popmax_freq_indices = hl.range(0, hl.len(freq_meta)).filter(
         lambda i: (hl.set(freq_meta[i].keys()) == {"group", "pop"})
         & (freq_meta[i]["group"] == "adj")
@@ -166,6 +168,8 @@ def faf_expr(
     _pops_to_exclude = (
         hl.literal(pops_to_exclude) if pops_to_exclude is not None else {}
     )
+
+    # pylint: disable=invalid-unary-operand-type
     faf_freq_indices = hl.range(0, hl.len(freq_meta)).filter(
         lambda i: (freq_meta[i].get("group") == "adj")
         & (
--- a/gnomad/utils/plotting.py
+++ b/gnomad/utils/plotting.py
@@ -215,11 +215,14 @@ def plot_multi_hail_hist(
 
         hist_source = ColumnDataSource(data)
 
+        # fmt: off
+        hide_zeros_filter = BooleanFilter(
+            [top > 0 for top in hist_source.data["top"]]  # pylint: disable=unsubscriptable-object
+        )
+        # fmt: on
+
         view = (
-            CDSView(
-                source=hist_source,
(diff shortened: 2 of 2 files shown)


gnomad/utils/plotting.py
```diff
@@ -215,11 +215,14 @@ def plot_multi_hail_hist(
 
         hist_source = ColumnDataSource(data)
 
+        # fmt: off
+        hide_zeros_filter = BooleanFilter(
+            [top > 0 for top in hist_source.data["top"]]  # pylint: disable=unsubscriptable-object
+        )
+        # fmt: on
```
nawatts, line 222: Had to jump through some hoops here to prevent Black from moving the comment to a different line.

gnomad/utils/annotations.py
```diff
@@ -56,6 +56,8 @@ def pop_max_expr(
     :return: Popmax struct
     """
     _pops_to_exclude = hl.literal(pops_to_exclude)
+
+    # pylint: disable=invalid-unary-operand-type
```
jkgoodrich, line 60: Is this needed because of the `~` below? Will we need to add this comment to other places where we use the `~`?

gnomad/utils/plotting.py
```diff
@@ -215,11 +215,14 @@ def plot_multi_hail_hist(
 
         hist_source = ColumnDataSource(data)
 
+        # fmt: off
```
jkgoodrich, line 218: so the `# fmt: off` and `# fmt: on` are to disable the Black check for the lines in between them? What is the reason that`# pylint: disable=unsubscriptable-object` specifically has to be on this line rather than above (line 218) like done for the `# pylint: disable=invalid-unary-operand-type` in the other file?

jkgoodrich (commented):
Just a couple questions so I understand how it's working. Thanks for adding!

gnomad/utils/annotations.py
nawatts: Yes, Pylint incorrectly detects the `~` below. It doesn't look like every `~` results in a false positive. I didn't spend much time looking for a pattern.

Alternatively, we could disable this message in the project-level Pylint config.

gnomad/utils/plotting.py
nawatts: > so the # fmt: off and # fmt: on are to disable the Black check for the lines in between them? 
Yes, it doesn't format lines between `# fmt: off` and `# fmt: off`. https://black.readthedocs.io/en/stable/the_black_code_style.html

> What is the reason that # pylint: disable=unsubscriptable-object specifically has to be on this line rather than above (line 218) like done for the # pylint: disable=invalid-unary-operand-type in the other file?
It doesn't. I fixed this message first and it was only during fixing the the `invalid-unary-operand-type` messages that I found out about [block disables](http://pylint.pycqa.org/en/latest/user_guide/message-control.html#block-disables). Will change.

gnomad/utils/annotations.py
nawatts: I figured that since it wasn't causing a false positive on *every* instance of `~`, it could still provide some value.

jkgoodrich (approved):
lgtm

PR #346, opened 2021-03-16
Continue with docs build if unable to download Hail's objects.inv

nawatts:
Currently, we use [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) to link to Hail docs from references to Hail classes, functions, etc. in our docs. However, while that's a nice to have feature, it's not essential. If we can't download `objects.inv` from Hail's website for whatever reason (like if the website is temporarily down) that shouldn't block PR checks or prevent updates to our docs.

Currently, if `https://hail.is/docs/0.2/objects.inv` is unavailable, the docs build will fail. This changes the docs build to check if that file is available before configuring the intersphinx mapping for Hail. If it's not, the docs will build anyway, just without links to Hail docs.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,7 @@
 import sys
 from pathlib import Path
 
+import requests
 import sphinx_autodoc_typehints
 from sphinx.ext import autosummary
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -73,10 +74,12 @@ def extract_summary(doc, document):
...
 # https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html
-#intersphinx_mapping = {
-#    "hail": ("https://hail.is/docs/0.2", None),
-#}
+intersphinx_mapping = {}
 
+if requests.head("https://hail.is/docs/0.2/objects.inv").status_code == 200:
+    intersphinx_mapping["hail"] = ("https://hail.is/docs/0.2", None)
+else:
+    print("Unable to link to Hail docs (cannot access objects.inv)", file=sys.stderr)
```
jkgoodrich, line 82: probably a stupid question, for the main gnomad_methods we use logging rather than printing, is there a reason that print should be used here instead of logging?

jkgoodrich (approved):
Just a question, but otherwise lgtm

docs/conf.py
nawatts: I figured the timestamps and filtering options provided by logging wouldn't be as useful here as in pipelines, since this script runs in under a minute and other output from Sphinx is not timestamped.

jkgoodrich: Sounds good, thank you for explaining!

PR #347, opened 2021-03-17
Changed or to and in set_female_metrics_to_na

ch-kr:
Function previously was not correctly setting metrics to missing on chrY (XX samples only). For example:
```
+----------+---------+
| locus         | alleles     | ac_xx_adj | y_par | y_nonpar | y_conds |
+---------------+-------------+-----------+-------+----------+---------+
| locus<GRCh38> | array<str>  |     int32 |  bool |     bool |    bool |
+---------------+-------------+-----------+-------+----------+---------+
| chrY:2786709  | ["G","T"]   |        NA | false |     true |    true |
| chrY:2786719  | ["C","A"]   |         0 | false |     true |    true |
```

- `y_par` is an annotation showing whether the locus is in a pseudoautosomal region of chromosome Y
- `y_nonpar` is an annotation showing whether the locus is in a non-pseudoautosomal region of chromosome Y
- `y_conds` is an annotation with the following condition: `(~locus.in_y_nonpar() | ~locus.in_y_par())`

`ac_xx_adj` should be NA on both lines, but you can see that on both lines, `y_conds` evaluates to True, which means [line 779](https://github.com/broadinstitute/gnomad_methods/blob/master/gnomad/utils/vcf.py#L779) will return the original value of this metric (and not missing).


mike-w-wilson (approved):
LGTM!

PR #348, opened 2021-03-18
Updated make_combo_header_text to return description text if combo_di…

ch-kr:
…ct length is 1, group is adj, and 'gnomad' is in prefix

### DIFF
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -436,6 +436,8 @@ def make_combo_header_text(
 
     if len(combo_dict) == 1:
         if combo_dict["group"] == "adj":
+            if "gnomad" in prefix:
+                return f" {preposition} samples in gnomAD"
             return ""
 
     if "sex" in combo_dict:
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -436,8 +436,6 @@ def make_combo_header_text(
 
     if len(combo_dict) == 1:
         if combo_dict["group"] == "adj":
-            if "gnomad" in prefix:
-                return f" {preposition} samples in gnomAD"
             return ""
 
     if "sex" in combo_dict:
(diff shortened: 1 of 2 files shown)


gnomad/utils/vcf.py
```diff
@@ -436,6 +436,8 @@ def make_combo_header_text(
 
     if len(combo_dict) == 1:
         if combo_dict["group"] == "adj":
+            if "gnomad" in prefix:
+                return f" {preposition} samples in gnomAD"
```
jkgoodrich, line 440: should we at a postfix option instead of having a check for gnomad in the prefix?

gnomad/utils/vcf.py
```diff
@@ -436,6 +436,8 @@ def make_combo_header_text(
 
     if len(combo_dict) == 1:
         if combo_dict["group"] == "adj":
```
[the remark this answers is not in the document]
ch-kr: yup! this code previously read 
```
if len(combo_dict) == 1:
    if combo_dict["group"] == "adj":
        return ""
```

and this block downstream corrects raw:
```
    if "group" in combo_dict:
        if combo_dict["group"] == "raw":
            header_text = header_text + ", before removing low-confidence genotypes"
```

gnomad/utils/vcf.py
ch-kr: do you mean update the ukbb code instead? this function is called in `make_info_dict`, so we'd need to add a fix after this:
```
                for label_group in faf_label_groups:
                    vcf_info_dict.update(
                        make_info_dict(
                            prefix=subset,
                            pop_names=gnomad_exomes_pops,
                            label_groups=label_group,
                            faf=True,
                        )
                    )
``` 
(also after the same call for the genomes)

ch-kr: I moved the fix to `make_info_dict`, let me know what you think!

jkgoodrich (approved):
lgtm

PR #354, opened 2021-04-02
Simplify workflow for updating docs site

nawatts:
When a commit is made to the default branch, a [GitHub Actions](https://docs.github.com/en/actions) workflow builds the docs site and pushes the site content to the `gh-pages` branch, where it is then picked up by [GitHub Pages](https://docs.github.com/en/pages).

Currently, we use an SSH key stored in repository [secrets](https://docs.github.com/en/actions/reference/encrypted-secrets) together with a [deploy key](https://docs.github.com/en/developers/overview/managing-deploy-keys#deploy-keys) to authenticate the workflow's push to that branch. However, workflows include an [access token](https://docs.github.com/en/actions/reference/authentication-in-a-workflow) for the current repository and the [checkout action](https://github.com/actions/checkout) configures git to use that token for authentication. Thus, there's no need for our workflow to reconfigure authentication.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,8 +57,6 @@ jobs:
         run: ./docs/build.sh
       - name: Publish to GitHub Pages
         if: github.event_name == 'push'
-        env:
-          DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
         run: |
           # Create empty gh-pages branch
           git checkout --orphan gh-pages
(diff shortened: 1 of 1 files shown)


williamphu (approved):
That's good to know. Thanks Nick! LGTM.

nawatts:
Someone with admin access to this repository can delete the deploy key and secret. @broadinstitute/gnomad-admins

PR #357, opened 2021-04-06
Added readviz utils

ch-kr:
Added two functions from https://github.com/broadinstitute/gnomad-readviz/blob/master/step1__select_samples.py:
`het_hom_hemi_take_expr` (renamed `readviz_struct_expr`) and `hemi_expr`

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -975,18 +975,3 @@ def hemi_expr(
         # mt.GT[0] is alternate allele
         gt.is_haploid() & (sex_expr == male_str) & (gt[0] == 1),
     )
-
-
-def readviz_struct_expr(
-    s: hl.expr.StringExpression, gq: hl.expr.Int32Expression
-) -> hl.expr.StructExpression:
-    """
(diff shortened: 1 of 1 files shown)


gnomad/utils/annotations.py
```diff
@@ -950,3 +950,43 @@ def unphase_call_expr(call_expr: hl.expr.CallExpression) -> hl.expr.CallExpressi
...
+    """
+    return hl.or_missing(
+        locus.in_x_nonpar() | locus.in_y_nonpar(),
+        # Haploid genotypes have a single integer, so checking if
+        # mt.GT[0] is alternate allele
+        gt.is_haploid() & (sex_expr == male_str) & (gt[0] == 1),
+    )
+
+
+def readviz_struct_expr(
```
bw2, line 980: There may be other readviz-related structs, so I'd vote for renaming to something that describes the object like `sample_id_and_gq_struct_expr(..)` but this would be longer than the underlying expression, so I'd actually vote for getting rid of the method and just using the expression directly.

bw2 (approved):

PR #360, opened 2021-04-08
Refactor resource path validation

nawatts:
Each type of resource has some convention for file names (`.ht` for TableResource, `.mt` for MatrixTableResource, etc). Currently, the `__init__` method of BaseResource takes a list of file extensions to use to validate the path and each subclass overloads `__init__` to pass the list of expected file extensions for that subclass.

This moves the list of expected file extensions to a class variable on each subclass, which removes the need to overload `__init__`.

### DIFF
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -16,16 +16,16 @@ class BaseResource(ABC):
     :param path: The resource path
     :param import_args: Any sources that are required for the import and need to be kept track of (e.g. .vcf path for an imported VCF)
     :param import_func: A function used to import the resource. `import_func` will be passed the `import_args` dictionary as kwargs.
-    :param expected_file_extensions: A list of all expected file extensions. If path doesn't end with one of these, a warning if emitted.
     """
 
-    @abstractmethod
+    expected_file_extensions: List[str] = []
+    """Expected file extensions for this resource type. If path doesn't end with one of these, a warning is logged."""
+
     def __init__(
         self,
         path: Optional[str] = None,
         import_args: Optional[Dict[str, Any]] = None,
         import_func: Optional[Callable] = None,
-        expected_file_extensions: Optional[List[str]] = None,
     ):
         if path is None and import_func is None:
             raise ValueError(
@@ -38,13 +38,14 @@ class BaseResource(ABC):
 
         if (
             path is not None
-            and expected_file_extensions
-            and not any(path.endswith(ext) for ext in expected_file_extensions)
+            and self.expected_file_extensions
+            and not any(path.endswith(ext) for ext in self.expected_file_extensions)
         ):
             logger.warning(
-                "Created the following {} with a path that doesn't end with {}: {}".format(
-                    self.__class__.__name__, " or ".join(expected_file_extensions), self
-                )
+                "Created the following %s with a path that doesn't end with %s: %s",
+                self.__class__.__name__,
+                " or ".join(self.expected_file_extensions),
+                self,
             )
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


gnomad/resources/resource_utils.py
```diff
@@ -38,13 +39,14 @@ def __init__(
...
+            and not any(path.endswith(ext) for ext in self.expected_file_extensions)
         ):
             logger.warning(
-                "Created the following {} with a path that doesn't end with {}: {}".format(
-                    self.__class__.__name__, " or ".join(expected_file_extensions), self
-                )
+                "Created the following %s with a path that doesn't end with %s: %s",
+                self.__class__.__name__,
+                " or ".join(self.expected_file_extensions),
+                self,
```
nawatts, line 49: This resolves a [logging-not-lazy](https://pylint.readthedocs.io/en/latest/technical_reference/features.html#logging-checker-messages) message from Pylint.

gtiao: I'm not sure I follow why there's an error here from Pylint -- it looks like there's f-string formatting on line 55. Is this just due to the use of string variables in the logger? It would be nice to be have a consistent style across the repo. Does Pylint still object if we do `"Created the following {self.__class__.__name__} with a path that doesn't end with {" or ".join(expected_file_extensions)}: {self}"`?

gnomad/resources/resource_utils.py
```diff
@@ -168,6 +147,8 @@ class PedigreeResource(BaseResource):
     :param missing: The string used to denote missing values. For case-control, 0, -9, and non-numeric are also treated as missing.
     """
 
+    expected_file_extensions: List[str] = [".fam", ".ped"]
+
     def __init__(
```
gtiao, line 152: Do you keep the __init__ function here because you need to define extra parameters unique to pedigrees (`quant_pheno`, 'delimiter`, 'missing')?

gnomad/resources/resource_utils.py
nawatts: The warning is specifically about f-strings in logger calls. Logging with `logger.info("message about %s", obj)` instead of `logger.info(f"message about {obj}")` or `logger.info("message about %s" % obj)` avoids a crash if there's an error formatting the message and only constructs the message if it will actually be output.

https://medium.com/swlh/why-it-matters-how-you-log-in-python-1a1085851205

gnomad/resources/resource_utils.py
nawatts: Yes

gnomad/resources/resource_utils.py
jkgoodrich: huh, I didn't know this. If we are switching to this method we should make sure the rest of the team is aware because I think most of us use f-strings in logger calls.

nawatts: Ideally, we could fix all the Pylint warnings and then start enforcing them on PR checks.

https://github.com/broadinstitute/gnomad_methods/blob/e0f433aee33adefce837e8406f0eeb6323a71796/.github/workflows/ci.yml#L30-L31

gtiao: Thanks, Nick -- I didn't know about this either, and I support propagating this information to the entire team and enforcing this on PR checks

gtiao (approved):
LGTM

gnomad/resources/resource_utils.py
nawatts: Sounds good. There's a lot of instances of this in existing code which will have to be changed before this can be enabled in PR checks.

jkgoodrich: I'm happy to take another evening next week while watching a show to make these changes like I did for the docstrings. It's satisfying work that doesn't take as much thought.

nawatts: If you decide to, this will filter warnings to those related to lazy formatting for logging.

`./lint --disable=W --enable=logging-not-lazy,logging-format-interpolation,logging-fstring-interpolation`

PR #361, opened 2021-04-09
Add tests to PR checks

nawatts:
Configure PR checks to run tests using [pytest](https://docs.pytest.org/en/stable/) and add a few tests for resource classes.

### DIFF
--- a/tests/resources/test_resource_utils.py
+++ b/tests/resources/test_resource_utils.py
@@ -6,49 +6,53 @@ from gnomad.resources import resource_utils
 class TestTableResource:
     @patch("hail.read_table")
     def test_read_table(self, read_table):
-        resource = resource_utils.TableResource("gs://my-bucket/table.ht")
+        resource = resource_utils.TableResource("gs://gnomad-public/table.ht")
 
         ds = resource.ht()
-        read_table.assert_called_with("gs://my-bucket/table.ht")
+        read_table.assert_called_with("gs://gnomad-public/table.ht")
         assert ds == read_table.return_value
 
 
 class TestMatrixTableResource:
     @patch("hail.read_matrix_table")
     def test_read_matrix_table(self, read_matrix_table):
-        resource = resource_utils.MatrixTableResource("gs://my-bucket/matrix_table.mt")
+        resource = resource_utils.MatrixTableResource(
+            "gs://gnomad-public/matrix_table.mt"
+        )
 
         ds = resource.mt()
-        read_matrix_table.assert_called_with("gs://my-bucket/matrix_table.mt")
+        read_matrix_table.assert_called_with("gs://gnomad-public/matrix_table.mt")
         assert ds == read_matrix_table.return_value
 
 
 class TestPedigreeResource:
     @patch("hail.Pedigree.read")
     def test_read_pedigree(self, read_pedigree):
-        resource = resource_utils.PedigreeResource("gs://my-bucket/pedigree.ped")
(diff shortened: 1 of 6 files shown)


tests/resources/test_resource_utils.py
```diff
@@ -0,0 +1,54 @@
+from unittest.mock import patch
+
+from gnomad.resources import resource_utils
+
+
+class TestTableResource:
+    @patch("hail.read_table")
+    def test_read_table(self, read_table):
+        resource = resource_utils.TableResource("gs://my-bucket/table.ht")
```
williamphu, line 9: When testing, do we need a path to a valid file?

tests/resources/test_resource_utils.py
```diff
@@ -0,0 +1,54 @@
...
+        assert ds == read_matrix_table.return_value
+
+
+class TestPedigreeResource:
+    @patch("hail.Pedigree.read")
+    def test_read_pedigree(self, read_pedigree):
+        resource = resource_utils.PedigreeResource("gs://my-bucket/pedigree.ped")
+
+        ds = resource.pedigree()
+        read_pedigree.assert_called()
```
williamphu, line 32: Is there a reason why this class uses `assert_called()` vs `assert_called_with()`?

tests/resources/test_resource_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    @patch("hail.read_table")
+    def test_read_table(self, read_table):
+        resource = resource_utils.TableResource("gs://my-bucket/table.ht")
+
+        ds = resource.ht()
+        read_table.assert_called_with("gs://my-bucket/table.ht")
+        assert ds == read_table.return_value
+
+
+class TestMatrixTableResource:
```
williamphu, line 16: I might not be understanding the docs for pytest correctly, but are the tests below checked? How would these classes be checked?

williamphu (changes requested):
I think I have a couple questions about the pytests, but everything looks good.

tests/resources/test_resource_utils.py
nawatts: Not really. In this test, [patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) replaces the real `hl.read_table` function with a mock that does nothing. However, if the relevant Hail function isn't patched in a test, I guess it would be preferable to attempt reading from a bucket we control. Will change to a gnomAD bucket.

tests/resources/test_resource_utils.py
nawatts: Pytest automatically discovers tests in `testpaths` directories configured in pytest.ini.

> * In those directories, search for `test_*.py` or `*_test.py` files, imported by their test package name.
> * From those files, collect test items:
>    * `test` prefixed test functions or methods outside of class
>    * `test` prefixed test functions or methods inside `Test` prefixed test classes (without an `__init__` method)

https://docs.pytest.org/en/6.2.x/goodpractices.html#test-discovery

tests/resources/test_resource_utils.py
nawatts: The PedigreeResource methods call `hl.Pedigree.read` and `hl.import_fam` with some additional arguments, but I was only interested in testing the first argument (the path) here.

`assert_called_with("path")` would fail because the function was called with another argument.

williamphu (approved):
Looks good to me!

PR #367, opened 2021-04-23
Fixes to docstring formatting

jkgoodrich:
These are just the easy fixes to docstrings reported by `pydocstyle`. The remaining problems are actual missing docstrings and will take more thought to add.

PR #368, opened 2021-04-23
Add pydocstyle to pre-commit hooks

nawatts:
After #367, this adds the remaining changes necessary to successfully run [pydocstyle](https://www.pydocstyle.org/). It adds ignore comments for missing module/package level docstrings and adds or ignores a few missing function docstrings.

### DIFF
--- a/gnomad/resources/__init__.py
+++ b/gnomad/resources/__init__.py
@@ -1,1 +1,3 @@
+# noqa: D104
+
 from .resource_utils import *
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,3 +12,7 @@ repos:
       - id: black
         language_version: python3
         files: gnomad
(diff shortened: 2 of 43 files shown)


gnomad/resources/__init__.py
```diff
@@ -1 +1,3 @@
+# noqa: D104
```
jkgoodrich, line 1: Just asking for my knowledge, is there a reason its better that we specify D104 in each __init__.py rather than add where D105 and D107 are added in `pydocstylerc`?

jkgoodrich (approved):
lgtm

gnomad/resources/__init__.py
nawatts: If we ignore it individually in existing files, then pydocstyle will still complain about missing module docstrings in new files. Versus if we ignore it in `pydocstylerc`, new files can be created without module docstrings.

PR #369, opened 2021-04-23
Add pydocstyle to PR checks

nawatts:
Now that both #367 and #368 are merged and pydocstyle runs successfully, it can be added to PR checks. This ensures that docstrings follow conventions even if a contributor isn't using the pre-commit hook added in #368.

This also reorders checks so that fast checks (Black and pydocstyle) are run before slower ones (Pylint and pytest). This avoids running the slow checks on PRs that require changes to formatting/docstrings.


jkgoodrich (approved):
lgtm, thank you Nick!

PR #371, opened 2021-04-26
Only ignore Pylint warnings that are present in existing code

nawatts:
Currently, in PR checks, we disable all Pylint warnings and re-enable a few.

https://github.com/broadinstitute/gnomad_methods/blob/9ce5fdbf5a07e80ea96dc2acb6b5c3943c8f9c97/.github/workflows/ci.yml#L34-L35

This changes Pylint configuration to ignore only the specific warnings that are present in existing code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Check docstrings
         run: pydocstyle gnomad
       - name: Run Pylint
-        run: ./lint --disable=W --enable=logging-not-lazy,logging-format-interpolation,logging-fstring-interpolation
+        run: ./lint
       - name: Run tests
         run: python -m pytest
   docs:
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,6 +1,23 @@
 [MESSAGES CONTROL]
 
 disable=
-  unused-wildcard-import,
+  # Ignore refactor and convention categories
   R,
-  C
+  C,
+  # Allow TODO and FIXME comments
+  fixme,
+  # Ignore unnecessary-lambda because it has many false positives from hl.ArrayExpression.map, etc.
+  unnecessary-lambda,
+  # Warnings present in existing code. These should be fixed and removed from this list.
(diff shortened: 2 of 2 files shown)


.pylintrc
```diff
@@ -1,6 +1,23 @@
...
-  unused-wildcard-import,
+  # Ignore refactor and convention categories
   R,
-  C
+  C,
+  # Allow TODO and FIXME comments
+  fixme,
+  # Ignore unnecessary-lambda because it has many false positives from hl.ArrayExpression.map, etc.
+  unnecessary-lambda,
+  # Warnings present in existing code. These should be fixed and removed from this list.
```
jkgoodrich, line 11: So is this something we should fix first and then remove these lines? We can put in this PR, but I am also happy to try to fix these after. I'm assuming I will just see them all when I run ./lint?

.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Check docstrings
         run: pydocstyle gnomad
       - name: Run Pylint
-        run: ./lint --disable=W --enable=logging-not-lazy,logging-format-interpolation,logging-fstring-interpolation
+        run: ./lint
```
jkgoodrich, line 35: Thank you for fixing!

jkgoodrich (approved):

.pylintrc
nawatts: Yes, the code causing the warning will have to be fixed before the warning can be removed from this list. Otherwise, PR checks will fail.

`./lint` will now succeed because all of these warnings are disabled. To enable one and find code that causes it, either remove it from this list and run `./lint` or enable it it Pylint CLI args (ex: `./lint --enable=unused-import`).

.github/workflows/ci.yml
nawatts: This was working fine. Changing from disable all + re-enable some to disable some just clarifies the list of what's remaining to be resolved.

PR #372, opened 2021-04-26
Run pydocstyle on tests

nawatts:
Currently, PR checks run pydocstyle on files in the `gnomad` directory.

https://github.com/broadinstitute/gnomad_methods/blob/b5f83b81d42d95d3c1de3bd1b744b363690ac3ac/.github/workflows/ci.yml#L32-L33

Pre-commit hooks, however, run it on all Python files that changed in the commit.

https://github.com/broadinstitute/gnomad_methods/blob/b5f83b81d42d95d3c1de3bd1b744b363690ac3ac/.pre-commit-config.yaml#L15-L18

Thus, it's currently possible for a change to pass PR checks, get merged, and then cause pre-commit hooks to fail when it is pulled locally. This change brings PR checks and pre-commit hooks closer in line by running pydocstyle on tests in PR checks.

### DIFF
--- a/.pydocstylerc
+++ b/.pydocstylerc
@@ -1,5 +1,6 @@
 [pydocstyle]
 convention = pep257
+match = .*\.py
 add-ignore =
   D105, # Ignore missing docstring for magic methods.
   D107 # Ignore missing docstring for __init__ method. Parameters are documented in the class docstring.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Check formatting
         run: black --check gnomad tests
       - name: Check docstrings
-        run: pydocstyle gnomad
+        run: pydocstyle gnomad tests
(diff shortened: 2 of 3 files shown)


.pydocstylerc
```diff
@@ -1,5 +1,6 @@
 [pydocstyle]
 convention = pep257
+match = .*\.py
```
ch-kr, line 3: basic question, why is this line necessary if you've changed the workflow yaml to `pydocstyle gnomad tests`?

ch-kr (commented):
small question

.pydocstylerc
nawatts: The default for `match` ignores files named `test_*.py`. So without this, `pydocstyle tests` doesn't actually check the test files contained in the `tests` directory.

http://www.pydocstyle.org/en/stable/usage.html#cli-usage

ch-kr (approved):
lgtm!

PR #376, opened 2021-05-06
Fix some Pylint warnings

nawatts:
#371 started checking Pylint warnings (previously, we only checked error and fatal level messages) in PR checks. However, some warnings were left ignored because they were present in existing code. This fixes a few of those warnings and removes them from the ignored list.

Descriptions of Pylint messages can be found at https://docs.pylint.org/en/stable/technical_reference/features.html.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -11,13 +11,10 @@ disable=
   # Warnings present in existing code. These should be fixed and removed from this list.
   cell-var-from-loop,
   dangerous-default-value,
-  f-string-without-interpolation,
   isinstance-second-argument-not-valid-type,
   protected-access,
   raise-missing-from,
   redefined-builtin,
   redefined-outer-name,
-  unused-import,
-  unnecessary-pass,
   unused-variable,
   unused-wildcard-import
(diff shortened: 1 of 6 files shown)


.pylintrc
```diff
@@ -11,13 +11,10 @@ disable=
   # Warnings present in existing code. These should be fixed and removed from this list.
   cell-var-from-loop,
   dangerous-default-value,
-  f-string-without-interpolation,
```
gtiao, line 14: These removals look good, but I looked over the other warnings, and some of them seem like they would also be good to remove (and hence check), e.g., `cell-var-from-loop`, `redefined-builtin`, `unused-variable`, and (eventually) `unused-wildcard-import`

gtiao (commented):
Looks good, just a comment about the other warnings we might want to be checking

.pylintrc
nawatts: These were all ignored because they were present in existing code when we started checking warnings (#371). As noted in the comment in .pylintrc, all of these should eventually be fixed and removed from the ignore list (unless there's a specific reason for ignoring one).

This PR was limited to a few warnings to keep it small and allow for easier/quicker review. Also, trying to fix many unrelated things in one PR risks the entire PR getting held up on discussion of one specific item. With the set of warnings in this PR, that didn't seem likely because all of these warnings were straightforward and only occurred one or a few times.

nawatts: There is not.

nawatts: Filed an issue for it (#378).

gtiao (approved):

Issue #378, opened 2021-05-14
Resolve and enable Pylint warnings


nawatts:
#371 started checking most Pylint warnings in PR checks. However, some warnings were left ignored because they occurred in existing code.

Ignored warnings are listed in [`.pylintrc`](https://github.com/broadinstitute/gnomad_methods/blob/master/.pylintrc).

ch-kr:
closing because this no longer seems relevant

PR #379, opened 2021-05-17
Fix annotation type helper functions

nawatts:
Currently, `annotation_type_is_numeric` and `annotation_type_in_vcf_info` do not work.

`isinstance(t, hl.tint32)`, `isinstance(t, hl.tstr)` etc. raise TypeErrors because `hl.tstr`, `hl.int32`, etc. are not Python types (they are instances of a HailType class).

Pylint detects this issue, but we were ignoring this warning.

Related to #378

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -11,7 +11,6 @@ disable=
   # Warnings present in existing code. These should be fixed and removed from this list.
   cell-var-from-loop,
   dangerous-default-value,
-  isinstance-second-argument-not-valid-type,
   protected-access,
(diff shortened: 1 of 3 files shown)


williamphu (approved):
The changes look good to me! I never realized that hl.tint, tstr, etc. are slightly different from the tset, tarray. Does `isinstance` work for all container types in hail, vs the primitives?

nawatts:
Force-pushed to resolve merge conflicts in CHANGELOG.md.

PR #381, opened 2021-05-20
Add generalized functions from gnomad_qc vcf export

jkgoodrich:


gnomad/utils/filtering.py
```diff
@@ -341,3 +341,17 @@ def filter_to_clinvar_pathogenic(
         t.count_rows() if isinstance(t, hl.MatrixTable) else t.count(),
     )
     return t
+
+
+def remove_fields_from_globals(global_field: List[str], fields_to_remove: List[str]):
```
ch-kr, line 346: now that this function has been moved to the common repo/taken out of the context of VCF export, I wonder if there's a better name? I actually forgot what this did and thought this function was going to remove globals from HTs/MTs. maybe `remove_fields_from_constants` is better?

gnomad/utils/filtering.py
```diff
@@ -341,3 +341,17 @@ def filter_to_clinvar_pathogenic(
...
+    """
+    Remove fields from the pre-defined global field variables.
+
```
ch-kr, line 349: might be helpful to include an example of when we'd use this function:
```suggestion

Used when exporting data to VCF. 
For example, used to remove "decoy" from `REGION_FLAG_FIELDS` in v3.1 VCF export.
```

gnomad/utils/vcf.py
```diff
@@ -493,6 +493,31 @@ def make_combo_header_text(
     return header_text
 
 
+def create_label_groups(
+    pops: List[str],
+    sexes: List[str],
```
ch-kr, line 498: since this is in the common repo now, should we set a default for this field too? possibly to this? https://github.com/broadinstitute/gnomad_methods/blob/master/gnomad/utils/vcf.py#L45 (we should also update this constant to use XX and XY now)

gnomad/utils/vcf.py
```diff
@@ -493,6 +493,31 @@ def make_combo_header_text(
...
+    pop_sex_groups: List[str] = ["adj"],
+) -> List[Dict[str, List[str]]]:
+    """
+    Generate a list of label group dictionaries needed to populate info dictionary.
+
+    Label dictionaries are passed as input to `make_info_dict`.
+
+    :param pops: List of population names.
+    :param sexes: List of sample sexes.
+    :param groups: List of data types (raw, adj). Default is ["raw", "adj"].
```
ch-kr, line 509:
```suggestion
    :param groups: List of data types (raw, adj). Default is `GROUPS`, which is ["raw", "adj"].
```

ch-kr: I'm also wondering if there is a better name for this now that you added `pop_sex_groups`. maybe `all_groups`?

gnomad/utils/vcf.py
```diff
@@ -493,6 +493,31 @@ def make_combo_header_text(
...
+    :param pop_sex_groups: List of only data types to populate with pops and sexes (raw, adj). Default is ["adj"].
```
ch-kr, line 510:
```suggestion
    :param pop_sex_groups: List of data types (raw, adj) to populate with pops and sexes. Default is ["adj"].
```

gnomad/utils/vcf.py
```diff
@@ -493,6 +493,31 @@ def make_combo_header_text(
...
+    :return: List of label group dictionaries.
+    """
+    return [
+        dict(group=groups),
```
ch-kr, line 514: might be helpful to add why we need `groups` and `pop_sex_groups` (it isn't immediately obvious upon reading the code). I had a comment similar to the one below in the UKBB code
```suggestion
        # This is to capture raw frequency fields, which are
        # not stratified by sex or population (e.g., only AC_raw exists, not AC_XX_raw)
        dict(group=groups),
```

gnomad/utils/vcf.py
```diff
@@ -865,3 +890,64 @@ def set_female_y_metrics_to_na(
             }
         )
     return female_metrics_dict
+
+
+def build_vcf_export_reference(
+    name: str,
+    build: str = "GRCh38",
+    contigs_keep: List[str] = [f"chr{i}" for i in range(1, 23)]
```
ch-kr, line 898: `keep_contigs` sounds slightly better to me but no real preference

gnomad/utils/vcf.py
```diff
@@ -865,3 +890,64 @@ def set_female_y_metrics_to_na(
...
+    + ["chrX", "chrY", "chrM"],
+) -> hl.ReferenceGenome:
+    """
+    Create export reference based on reference genome defined by `build`.
+
+    By default this will return a new reference with all non-standard contigs eliminated. Keeps chr 1-22, Y, X, and M.
+
+    Non-standard for example is: ##contig=<ID=chr3_GL000221v1_random,length=155397,assembly=GRCh38>
```
ch-kr, line 906:
```suggestion
    An example of a non-standard contig is: ##contig=<ID=chr3_GL000221v1_random,length=155397,assembly=GRCh38>
```

gnomad/utils/vcf.py
```diff
@@ -865,3 +890,64 @@ def set_female_y_metrics_to_na(
...
+
+    :param name: Name to use for new reference.
+    :param build: Reference genome build to use as starting reference genome.
+    :param contigs_keep: Contigs to keep from reference genome defined by `build`.
```
ch-kr, line 910:
```suggestion
    :param contigs_keep: Contigs to keep from reference genome defined by `build`. Default is autosomes, sex chromosomes, and chrM.
```

gnomad/utils/vcf.py
```diff
@@ -865,3 +890,64 @@ def set_female_y_metrics_to_na(
...
+    :return: Reference genome for VCF export containing only contigs in `contigs_keep`
```
ch-kr, line 911:
```suggestion
    :return: Reference genome for VCF export containing only contigs in `contigs_keep`.
```
(adding period for consistency with other docstring lines)

gnomad/utils/vcf.py
```diff
@@ -865,3 +890,64 @@ def set_female_y_metrics_to_na(
...
+        name=name,
+        contigs=contigs_keep,
+        lengths={contig: ref.lengths[contig] for contig in contigs_keep},
+        x_contigs=ref.x_contigs,
+        y_contigs=ref.y_contigs,
+        par=[
+            (interval.start.contig, interval.start.position, interval.end.position)
+            for interval in ref.par
+        ],
+        mt_contigs=ref.mt_contigs,
```
ch-kr, line 925: just curious, is this still necessary if `contigs_keep` doesn't specify chrM?

ch-kr (changes requested):
just a few small suggestions. Thank you for generalizing these functions!!

gnomad/utils/vcf.py
jkgoodrich: I don't think so, but it will just return an empty list which is what the default is for initialization, so it will still work

gnomad/utils/filtering.py
jkgoodrich: I changed some of the other info in the docstring, let me know if you still think an example of when to use is needed

gnomad/utils/filtering.py
```diff
@@ -341,3 +341,21 @@ def filter_to_clinvar_pathogenic(
         t.count_rows() if isinstance(t, hl.MatrixTable) else t.count(),
     )
     return t
+
+
+def remove_fields_from_constant(
+    constant: List[str], fields_to_remove: List[str]
+) -> List[str]:
+    """
+    Remove fields from a list indicating which fields were missing from the original list.
```
ch-kr, line 350:
```suggestion
    Remove fields from a list and display any field(s) missing from the original list.
```

gnomad/utils/filtering.py
ch-kr: looks good to me! I added a slight rewording of the description (seems clearer to me, but no strong feelings)

gnomad/utils/vcf.py
```diff
@@ -42,7 +42,7 @@
 Global populations that are included in filtering allele frequency (faf) calculations. Used in VCF export.
 """
 
-SEXES = ["male", "female"]
+SEXES = ["XX", "XY"]
 """
 Sample sexes used in VCF export.
```
ch-kr, line 47:
```suggestion
Sample sexes used in VCF export.

```
(not sure if there needs to be a space here)

gnomad/utils/vcf.py
```diff
@@ -42,7 +42,7 @@
...
 Used to stratify frequency annotations (AC, AN, AF) for each sex.
```
ch-kr, line 48:
```suggestion
Used to stratify frequency annotations (AC, AN, AF) for each sex.
Note that sample sexes in gnomAD v3 and earlier were 'male' and 'female'. 
```

ch-kr (approved):
a couple more tiny edits but LGTM!

PR #388, opened 2021-06-21
Improve error message for when a public resource is unavailable

nawatts:
Some resources may not be available from all sources due to delays in syncing, etc. In those cases, it would be preferable to provide a helpful error message (directing the user to try reading the resource from the gnomAD buckets) instead of simply letting the read fail.

This wraps all methods of GnomadPublicResource subclasses that read the resource (GnomadPublicTableResource.ht, GnomadPublicMatrixTableResource.mt, etc.) and adds a check for if the resource is available from the selected source. If the resource is not available, an exception with a helpful message is thrown.

### DIFF
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -2,8 +2,8 @@
 
 import logging
 from abc import ABC, abstractmethod
-from functools import reduce
-from typing import Any, Callable, Dict, List, Optional
+from functools import reduce, wraps
+from typing import Any, Callable, Dict, Iterable, List, Optional
 
 import hail as hl
 from hail.linalg import BlockMatrix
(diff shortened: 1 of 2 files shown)


nawatts:
Using [`__init_subclass__`](https://docs.python.org/3/reference/datamodel.html#customizing-class-creation) for this seemed like it would require less boilerplate than overriding methods in GnomadPublicResource subclasses, which would require keeping method signatures and docstrings in sync with the base Resource subclasses.

gnomad/resources/resource_utils.py
```diff
@@ -373,9 +373,47 @@ def __init__(self, default_version: str, versions: Dict[str, BlockMatrixResource
...
+            @wraps(original_method)
+            def read_resource(self, *args, **kwargs):
+                # If one of the known sources is selected, check if the resource is available.
+                # For custom sources, skip the check and attempt to read the resource.
+                resource_source = gnomad_public_resource_configuration.source
+                if (
+                    isinstance(resource_source, GnomadPublicResourceSource)
+                    and resource_source != GnomadPublicResourceSource.GNOMAD
+                ):
+                    if not self.is_resource_available():
```
mike-w-wilson, line 402: Do you think it makes sense to check if the gnomAD resource exists here( given how occasionally the repo lags behind buckets)?  I'm wondering if it makes sense to add functionality to load from the gnomAD source on failure? Or if that gets more complicated as we add other cloud providers public buckets.

mike-w-wilson (approved):
This looks good to me. I just had one question on resource failure and if it makes sense to add functionality to load from gnomAD if a resource is not found. That is likely another piece of work/PR though.

gnomad/resources/resource_utils.py
nawatts: This was assuming that resources will always be present in the gnomAD bucket, since they are copied from there to the cloud provider buckets.

I did not consider the case where an older version of this code points to a resource that no longer exists in the gnomAD buckets. Ideally, those URLs would not change but, practically, they do. That's a good point. We could catch that and suggest users update to the most recent version of gnomad_methods.

I'll claim that we should not automatically load from the gnomAD source if the resource doesn't exist. Most Hail resources are in the requester pays bucket, so if a user has configured their pipeline to use one of the cloud provider buckets, automatically falling back to the gnomAD buckets could result in unexpected egress/operation costs for that user. Also, they may not have Hail configured to allow requester pays access. I think the best we can do is point them towards the solution.

mike-w-wilson: I did not think of the cost associated with the requester-pays bucket, it absolutely makes sense to not automatically load the gnomAD source.

nawatts: Filed #390 for this.

PR #389, opened 2021-06-25
Adding common sanity checks

mike-w-wilson:
This moves a number of sanity checks from the gnomad_qc repo and merges them with Julia's updated aggregation approach for faster runtimes. I think the filters check still needs a bit of work and could be approached a bit differently but I figured I'll iterate after feedback. There is a general method to run all checks. It takes in a lot of arguments which I don't love but I think it is necessary for making sure all the check's parameters can be passed through. This was tested on a test subset from both the full callset and the hgdp_tgp subset using the prepare_vcf_data release script from gnomad_qc. I also removed the histogram sanity checks as the histogram fields it checked, n_smaller and n_larger, were removed and it no longer served any purpose.


gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    metrics_frac_missing = {}
+    for x in info_metrics:
+        metrics_frac_missing[x] = hl.agg.sum(hl.is_missing(t.info[x])) / n_sites
+    for x in non_info_metrics:
+        metrics_frac_missing[x] = hl.agg.sum(hl.is_missing(t[x])) / n_sites
+    output = t.aggregate(hl.struct(**metrics_frac_missing))
+
+    n_fail = 0
+    for metric, value in dict(output).items():
+        if value > missingness_threshold:
```
jkgoodrich, line 787: I wonder for these loggers we should also print the number of sites missing, it might be nice to confirm none missing for vep even though the percent might pass the missingness threshold so with the number of sites missing printed I can confirm that easily

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +116,83 @@ def make_filters_sanity_check_expr(
...
+    t: Union[hl.MatrixTable, hl.Table],
+    subset: str,
     label_groups: Dict[str, List[str]],
-    verbose: bool,
-    subpop: bool = None,
     sort_order: List[str] = SORT_ORDER,
-) -> None:
+    delimiter: str = "-",
+    metric_first_label: bool = True,
+) -> Tuple[dict, dict]:
```
jkgoodrich, line 125: PART OF LARGE SUGGESTION

```suggestion
) -> dict:
```

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +116,83 @@ def make_filters_sanity_check_expr(
...
-        show only top values of annotations that fail checks.
-    :param subpop: Subpop abbreviation, supplied only if subpopulations are included in the annotation groups being checked.
+    :param t: Input MatrixTable or Table containing annotations to be summed.
+    :param subset: String indicating sample subset.
+    :param label_groups: Dictionary containing an entry for each label group, where key is the name of the grouping, e.g. "sex" or "pop", and value is a list of all possible values for that grouping (e.g. ["XY", "XX"] or ["afr", "nfe", "amr"]).
     :param sort_order: List containing order to sort label group combinations. Default is SORT_ORDER.
-    :return: None
+    :param delimiter: String to use as delimiter when making group label combinations.
+    :param metric_first_label: If True, metric precedes label group, e.g. AC-afr-male. If False, label group precedes metric, afr-male-AC.
+    :return: Tuple of dictionaries
```
jkgoodrich, line 137: PART OF LARGE SUGGESTION
```suggestion
    :return: Dictionary of sample sum field check expressions and display fields
```

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +116,83 @@ def make_filters_sanity_check_expr(
...
+            if field in info_fields:
+                subfield_values.append(t.info[field])
+            else:
+                logger.info("%s is not in table's info field", field)
 
-    ht = ht.annotate(**annot_dict)
+        annot_dict[f"sum_{subfield}"] = hl.sum(subfield_values)
 
+    field_check_expr = {}
+    field_check_details = {}
```
jkgoodrich, line 170: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param verbose: If True, show top values of annotations being checked, including checks that pass; if False, show only top values of annotations that fail checks.
+    :param show_percent_sites: If true, show the percentage and count of overall sites that fail; if False, only show the number of sites that fail.
+    :param delimiter: String to use as delimiter when making group label combinations.
+    :param metric_first_label: If True, metric precedes label group, e.g. AC-afr-male. If False, label group precedes metric, afr-male-AC.
+    :return: None
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
+
+    field_check_expr = {}
+    field_check_details = {}
```
jkgoodrich, line 405: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param sexes: List of sexes in table.
+    :param subset_pops: Dict with subset (keys) and list of populations within subset (values). An empty string, e.g. "", should be passed as key with callset pops as value to test entire callset.
+    :param verbose: If True, show top values of annotations being checked, including checks that pass; if False, show only top values of annotations that fail checks.
+    :param metric_first_label: If True, metric precedes label group, e.g. AC-afr-male. If False, label group precedes metric, afr-male-AC.
+    :return: None
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
+
+    field_check_expr = {}
+    field_check_details = {}
```
jkgoodrich, line 479: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param subsets: List of sample subsets.
+    :param verbose: If True, show top values of annotations being checked, including checks that pass; if False, show only top values of annotations that fail checks.
+    :param delimiter: String to use as delimiter when making group label combinations.
+    :param metric_first_label: If True, metric precedes label group, e.g. AC-afr-male. If False, label group precedes metric, afr-male-AC.
+    :return: None
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
+
+    field_check_expr = {}
+    field_check_details = {}
```
jkgoodrich, line 582: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    t_x = hl.filter_intervals(t, [hl.parse_locus_interval("chrX")])
+    t_xnonpar = t_x.filter(t_x.locus.in_x_nonpar())
+    n = t_xnonpar.count()
+    logger.info("Found %d X nonpar sites", n)
+
+    logger.info("Check (nhomalt == nhomalt_xx) for X nonpar variants:")
+    xx_metrics = [x for x in xx_metrics if "nhomalt" in x]
+
+    field_check_expr = {}
+    field_check_details = {}
```
jkgoodrich, line 729: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +116,83 @@ def make_filters_sanity_check_expr(
...
+            cond_expr=t.info[check_field_left] != annot_dict[f"sum_{subfield}"],
+            display_fields=hl.struct(
+                **{
+                    check_field_left: t.info[check_field_left],
+                    f"sum_{subfield}": annot_dict[f"sum_{subfield}"],
+                }
+            ),
+        )
+
+    return field_check_expr, field_check_details
```
jkgoodrich, line 195: PART OF LARGE SUGGESTION
```suggestion
        field_check_expr[f"{check_field_left} = {check_field_right}"] = {
            "expr": hl.agg.count_where(
                t.info[check_field_left] != annot_dict[f"sum_{subfield}"]
            ),
            "display_fields": hl.struct(
                **{
                    check_field_left: t.info[check_field_left],
                    f"sum_{subfield}": annot_dict[f"sum_{subfield}"],
                }
            ),
        }

    return field_check_expr
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :return: Tuple of dictionaries
+    """
+    field_check_expr[check_description] = hl.agg.filter(cond_expr, hl.agg.count())
+    field_check_details[check_description] = hl.struct(
+        cond_expr=cond_expr, display_fields=display_fields
+    )
+
+    return field_check_expr, field_check_details
+
+
```
jkgoodrich, line 344: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+def generic_field_check_loop(
+    ht: hl.Table,
+    field_check_expr: dict,
+    field_check_details: dict,
```
jkgoodrich, line 348: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    show_percent_sites: bool = False,
+    ht_count: int = None,
+):
+    """
+    Loop through all conditional checks for a given hail Table.
+
+    This loop allows aggregation across the hail Table once, as opposed to aggregating during every conditional check. 
+    :param ht: Table containing annotations to be checked.
+    :param field_check_expr: Dictionary whose keys are conditions being checked and values are the expressions for filtering to condition.
+    :param field_check_details: Dictionary whose keys are the the check descriptions and values are struct expressions used for check and what to display for the check in the terminal.
```
jkgoodrich, line 359: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+        generic_field_check(
+            ht,
+            cond_expr=field_check_details[check_description].cond_expr,
+            check_description=check_description,
+            n_fail=n_fail,
+            display_fields=field_check_details[check_description].display_fields,
+            verbose=verbose,
+            show_percent_sites=show_percent_sites,
+            ht_count=ht_count,
+        )
```
jkgoodrich, line 375: PART OF LARGE SUGGESTION
```suggestion
    ht_field_check_counts = ht.aggregate(
        hl.struct(**{k: v['expr'] for k, v in field_check_expr.items()})
    )
    for check_description, n_fail in ht_field_check_counts.items():
        generic_field_check(
            ht,
            check_description=check_description,
            n_fail=n_fail,
            display_fields=field_check_expr[check_description]['display_fields'],
            verbose=verbose,
            show_percent_sites=show_percent_sites,
            ht_count=ht_count,
        )
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+                        field_check_details=field_check_details,
+                        check_description=f"{check_field_left} != {check_field_right}",
+                        cond_expr=t.info[check_field_left] == t.info[check_field_right],
+                        display_fields=hl.struct(
+                            **{
+                                check_field_left: t.info[check_field_left],
+                                check_field_right: t.info[check_field_right],
+                            }
+                        ),
+                    )
```
jkgoodrich, line 430: PART OF LARGE SUGGESTION
```suggestion

                    field_check_expr[f"{check_field_left} != {check_field_right}"] = {
                        "expr": hl.agg.count_where(
                            t.info[check_field_left] == t.info[check_field_right]
                        ),
                        "display_fields": hl.struct(
                            **{
                                check_field_left: t.info[check_field_left],
                                check_field_right: t.info[check_field_right],
                            }
                        ),
                    }
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+
+    generic_field_check_loop(
+        t,
+        field_check_expr,
+        field_check_details,
```
jkgoodrich, line 435: PART OF LARGE SUGGESTION
```suggestion
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
+
+    field_check_expr = {}
+    field_check_details = {}
+    for subset, pops in sample_sum_sets_and_pops.items():
+        pop_names = pops
+
+        # We do not store the raw callstats for the pop or sex groupings of any subset so only check adj here.
+        field_check_expr_s, field_check_details_s = sample_sum_check(
```
jkgoodrich, line 484: PART OF LARGE SUGGESTION
```suggestion
        field_check_expr_s = sample_sum_check(
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+            t,
+            subset,
+            dict(group=["adj"], pop=pop_names),
+            sort_order,
+            delimiter,
+            metric_first_label,
+        )
+        field_check_expr.update(field_check_expr_s)
+        field_check_details.update(field_check_details_s)
+        field_check_expr_s, field_check_details_s = sample_sum_check(
```
jkgoodrich, line 494: PART OF LARGE SUGGESTION
```suggestion
        field_check_expr_s = sample_sum_check(
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+            t,
+            subset,
+            dict(group=["adj"], sex=sexes),
+            sort_order,
+            delimiter,
+            metric_first_label,
+        )
+        field_check_expr.update(field_check_expr_s)
+        field_check_details.update(field_check_details_s)
+        field_check_expr_s, field_check_details_s = sample_sum_check(
```
jkgoodrich, line 504: PART OF LARGE SUGGESTION
```suggestion
        field_check_expr_s = sample_sum_check(
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+            subset,
+            dict(group=["adj"], pop=list(set(pop_names)), sex=sexes),
+            sort_order,
+            delimiter,
+            metric_first_label,
+        )
+        field_check_expr.update(field_check_expr_s)
+        field_check_details.update(field_check_details_s)
+
+        generic_field_check_loop(t, field_check_expr, field_check_details, verbose)
```
jkgoodrich, line 515: PART OF LARGE SUGGESTION
```suggestion

        generic_field_check_loop(t, field_check_expr, verbose)
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+        check_field = f"{subfield}{delimiter}adj"
+        field_check_expr, field_check_details = make_field_check_dicts(
+            field_check_expr=field_check_expr,
+            field_check_details=field_check_details,
+            check_description=f"{check_field} >= 0",
+            cond_expr=t.info[check_field] < 0,
+            display_fields=hl.struct(
+                **{check_field: t.info[check_field], "filters": t.filters}
+            ),
+        )
```
jkgoodrich, line 603: PART OF LARGE SUGGESTION
```suggestion

        field_check_expr[f"{check_field} > 0"] = {
            "expr": hl.agg.count_where(t.info[check_field] <= 0),
            "display_fields": hl.struct(**{check_field: t.info[check_field]}),
        }

        check_field = f"{subfield}{delimiter}adj"
        field_check_expr[f"{check_field} >= 0"] = {
            "expr": hl.agg.count_where(t.info[check_field] < 0),
            "display_fields": hl.struct(
                **{check_field: t.info[check_field], "filters": t.filters}
            ),
        }
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+
+    # Check adj AN >= 0
+    check_field = f"AN{delimiter}adj"
+    field_check_expr, field_check_details = make_field_check_dicts(
+        field_check_expr=field_check_expr,
+        field_check_details=field_check_details,
+        check_description=f"{check_field} >= 0",
+        cond_expr=t.info[check_field] < 0,
+        display_fields=hl.struct(**{check_field: t.info[check_field]}),
+    )
```
jkgoodrich, line 623: PART OF LARGE SUGGESTION
```suggestion
    field_check_expr[f"{check_field} > 0"] = {
        "expr": hl.agg.count_where(t.info[check_field] <= 0),
        "display_fields": hl.struct(**{check_field: t.info[check_field]}),
    }

    # Check adj AN >= 0
    check_field = f"AN{delimiter}adj"
    field_check_expr[f"{check_field} >= 0"] = {
        "expr": hl.agg.count_where(t.info[check_field] < 0),
        "display_fields": hl.struct(**{check_field: t.info[check_field]}),
    }

```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+            field_check_details=field_check_details,
+            check_description=f"{check_field_left} >= {check_field_right}",
+            cond_expr=t.info[check_field_left] < t.info[check_field_right],
+            display_fields=hl.struct(
+                **{
+                    check_field_left: t.info[check_field_left],
+                    check_field_right: t.info[check_field_right],
+                }
+            ),
+        )
```
jkgoodrich, line 639: PART OF LARGE SUGGESTION
```suggestion

        field_check_expr[f"{check_field_left} >= {check_field_right}"] = {
            "expr": hl.agg.count_where(
                t.info[check_field_left] < t.info[check_field_right]
            ),
            "display_fields": hl.struct(
                **{
                    check_field_left: t.info[check_field_left],
                    check_field_right: t.info[check_field_right],
                }
            ),
        }
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+                cond_expr=t.info[check_field_left] < t.info[check_field_right],
+                display_fields=hl.struct(
+                    **{
+                        check_field_left: t.info[check_field_left],
+                        check_field_right: t.info[check_field_right],
+                    }
+                ),
+            )
+
+    generic_field_check_loop(t, field_check_expr, field_check_details, verbose)
```
jkgoodrich, line 666: PART OF LARGE SUGGESTION
```suggestion

            field_check_expr[f"{check_field_left} >= {check_field_right}"] = {
                "expr": hl.agg.count_where(
                    t.info[check_field_left] < t.info[check_field_right]
                ),
                "display_fields": hl.struct(
                    **{
                        check_field_left: t.info[check_field_left],
                        check_field_right: t.info[check_field_right],
                    }
                ),
            }

    generic_field_check_loop(t, field_check_expr, verbose)
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +207,766 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+            != t_xnonpar.info[check_field_right],
+            display_fields=hl.struct(
+                **{
+                    check_field_left: t_xnonpar.info[check_field_left],
+                    check_field_right: t_xnonpar.info[check_field_right],
+                }
+            ),
+        )
+
+    generic_field_check_loop(t_xnonpar, field_check_expr, field_check_details, verbose)
```
jkgoodrich, line 750: PART OF LARGE SUGGESTION
```suggestion
        field_check_expr[f"{check_field_left} == {check_field_right}"] = {
            "expr": hl.agg.count_where(
                t_xnonpar.info[check_field_left] != t_xnonpar.info[check_field_right]
            ),
            "display_fields": hl.struct(
                **{
                    check_field_left: t_xnonpar.info[check_field_left],
                    check_field_right: t_xnonpar.info[check_field_right],
                }
            ),
        }

    generic_field_check_loop(t_xnonpar, field_check_expr, verbose)
```

jkgoodrich (changes requested):
OK, I haven't looked completely though everything, but I have one larger suggestion where if you decide to go with it then everything marked `PART OF LARGE SUGGESTION` can be added.

gnomad/assessment/sanity_checks.py
```diff
@@ -18,50 +19,57 @@ def generic_field_check(
...
-            ht_orig = ht_orig.flatten()
-            ht_orig.select(*display_fields).show()
+            ht.select(**display_fields).show()
 
 
 def make_filters_sanity_check_expr(
-    ht: hl.Table, extra_filter_checks: Optional[Dict[str, hl.expr.Expression]] = None
+    ht: hl.Table,
+    extra_filter_checks: Optional[Dict[str, hl.expr.Expression]] = None,
+    rf: bool = False,
```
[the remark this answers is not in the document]
mike-w-wilson: Yes --I went back and forth on this and choose the boolean only because I thought it would be more rigid but  I think the string approach is better since it is more flexible. Ill add the suggestion.

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +116,83 @@ def make_filters_sanity_check_expr(
...
-    Results from checking the sum of the specified annotations are printed in the terminal.
+    For example, if pop1 consists of pop1, pop2, and pop3, check that t.info.AC-subset1 == sum(t.info.AC-subset1-pop1, t.info.AC-subset1-pop2, t.info.AC-subset1-pop3).
 
-    :param ht: Table containing annotations to be summed.
-    :param prefix: String indicating sample subset.
-    :param label_groups: Dictionary containing an entry for each label group, where key is the name of the grouping,
-        e.g. "sex" or "pop", and value is a list of all possible values for that grouping (e.g. ["male", "female"] or ["afr", "nfe", "amr"]).
-    :param verbose: If True, show top values of annotations being checked, including checks that pass; if False,
-        show only top values of annotations that fail checks.
-    :param subpop: Subpop abbreviation, supplied only if subpopulations are included in the annotation groups being checked.
```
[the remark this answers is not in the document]
mike-w-wilson: I think subsets should be added in the future PR. We would want to add add a check so we can confrim something like if the continental pop "eur" has the sub pops "italy" and "britain" we'd check AC-eur == (AC-eur-italy + AC-eur-britain)

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +116,83 @@ def make_filters_sanity_check_expr(
...
+            else:
+                field = f"{subset}{subfield}{delimiter}{x}"
+
+            if field in info_fields:
+                subfield_values.append(t.info[field])
+            else:
+                logger.info("%s is not in table's info field", field)
 
-    ht = ht.annotate(**annot_dict)
+        annot_dict[f"sum_{subfield}"] = hl.sum(subfield_values)
```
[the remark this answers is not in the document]
mike-w-wilson: Good catch, yes it should

mike-w-wilson (commented):
Going to work on the large suggestion now just wanted to get these things in.

gnomad/assessment/sanity_checks.py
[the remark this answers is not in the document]
jkgoodrich: OK, that works for me, maybe add a TODO?

gnomad/assessment/sanity_checks.py
```diff
@@ -18,50 +19,57 @@ def generic_field_check(
     cond_expr: hl.expr.BooleanExpression,
```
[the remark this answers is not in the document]
mike-w-wilson: I think making the expr optional makes sense. I understand it may break some UKBB code but it would be a simple fix?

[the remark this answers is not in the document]
jkgoodrich: Yes, I looked yesterday and I think it is actually only a single line of code

mike-w-wilson:
@jkgoodrich, this is ready for another round of review. Thanks for the large suggestion - it really cleans up the code!

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +118,86 @@ def make_filters_sanity_check_expr(
...
+            if metric_first_label:
+                field = f"{metric}{delimiter}{subset}{label}"
+            else:
+                field = f"{subset}{metric}{delimiter}{label}"
 
-    ht = ht.annotate(**annot_dict)
+            if field in info_fields:
+                sum_group_exprs.append(t.info[field])
+            else:
+                logger.info("%s is not in table's info field", field)
```
jkgoodrich, line 169: I get these in the output, I think this would have identified that we were missing whole groups in our dataset, thank you so much for adding!!
```
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-papuan-adj is not in table's info field.   !!!! Is this OK, or not? I also think we are missing some of the needed subsets to check here if comparing to what we used to have 
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-biakapygmy-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-san-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-mbutipygmy-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-melanesian-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-papuan-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-biakapygmy-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-san-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-mbutipygmy-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-melanesian-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-papuan-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-biakapygmy-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-san-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-mbutipygmy-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-melanesian-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-papuan-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-papuan-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-biakapygmy-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-biakapygmy-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-san-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-san-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-mbutipygmy-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-mbutipygmy-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-melanesian-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AC-hgdp-melanesian-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-papuan-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-papuan-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-biakapygmy-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-biakapygmy-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-san-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-san-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-mbutipygmy-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-mbutipygmy-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-melanesian-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): AN-hgdp-melanesian-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-papuan-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-papuan-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-biakapygmy-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-biakapygmy-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-san-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-san-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-mbutipygmy-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-mbutipygmy-XY-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-melanesian-XX-adj is not in table's info field
INFO (gnomad.assessment.sanity_checks 169): nhomalt-hgdp-melanesian-XY-adj is not in table's info field
```

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +118,86 @@ def make_filters_sanity_check_expr(
...
+        annot_dict[
+            f"sum{delimiter}{metric}{delimiter}{group}{delimiter}{sum_group}"
+        ] = hl.sum(sum_group_exprs)
+
+    field_check_expr = {}
+    for metric in ["AC", "AN", "nhomalt"]:
+        if metric_first_label:
+            check_field_left = f"{metric}{delimiter}{subset}{group}"
+            check_field_right = f"sum{delimiter}{metric}{delimiter}{subset}{group}{delimiter}{sum_group}"
         else:
```
jkgoodrich, line 180: I just realized this could be further simplified
```suggestion
            check_field_left = f"{metric}{delimiter}{subset}{group}"
        else:
```

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +118,86 @@ def make_filters_sanity_check_expr(
...
-                f"{prefix}{subfield}_{subpop}_{group} = sum({subfield}_{group}_{alt_groups})",
-                [
-                    f"info.{prefix}{subfield}_{subpop}_{group}",
-                    f"sum_{subfield}_{group}_{alt_groups}",
-                ],
-                verbose,
-            )
+            check_field_left = f"{subset}{metric}{delimiter}{group}"
+            check_field_right = f"sum{delimiter}{subset}{metric}{delimiter}{group}{delimiter}{sum_group}"
+
```
jkgoodrich, line 183: More for the simplification
```suggestion
            check_field_left = f"{subset}{metric}{delimiter}{group}"

        check_field_right = f"sum{delimiter}{check_field_left}{delimiter}{sum_group}"
```

gnomad/assessment/sanity_checks.py
```diff
@@ -104,78 +118,86 @@ def make_filters_sanity_check_expr(
...
+        field_check_expr[f"{check_field_left} = {check_field_right}"] = {
+            "expr": hl.agg.count_where(
+                t.info[check_field_left]
+                != annot_dict[
+                    f"sum{delimiter}{metric}{delimiter}{group}{delimiter}{sum_group}"
```
jkgoodrich, line 188: This is redundant in a few places below, maybe make `{delimiter}{metric}{delimiter}{group}{delimiter}{sum_group}` a parameter above and then use that in place that it is used?

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+        "Checking distributions of variant type and region type amongst variant filters..."
+    )
+    _filter_agg_order(
+        t,
+        {
+            "allele_type": t.info.allele_type,
+            "in_problematic_region": t.in_problematic_region,
+        },
+        50,
+        140,
```
jkgoodrich, line 322: These are parameters in ` _filter_agg_order` do you think they should be parameters to `filters_sanity_check` also? If not then do they need to be parameters in ` _filter_agg_order`?

jkgoodrich (changes requested):
A few more comments and a couple changes, but it's getting close. I do want @ch-kr to take a look over it though since she has looked over and made changes to this code more than I have.

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    """
+    Check that all VCF fields and descriptions are present in input Table and VCF header dictionary.
+
+    :param t: Input MatrixTable or Tableto be exported to VCF.
+    :param header_dict: VCF header dictionary.
+    :param row_annotations: List of row annotations in MatrixTable.
+    :param hists: List of variant histogram annotations. Default is HISTS.
+    :return: Boolean with whether all expected fields and descriptions are present.
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
```
ch-kr, line 797: subsetting to rows here means that this function will only check sites VCFs, right? it might be helpful to leave the option to check genotypes for the future (not sure if this is relevant for anything other than HGDP/TGP for gnomad though)

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+
+    # Confirm all VCF fields/descriptions are present before exporting
+    hist_fields = []
+    for hist in hists:
+        hist_fields.extend(
```
ch-kr, line 802: this was written before the extra hists were dropped from the release file. maybe we should stop adding the extra hists here to `hist_fields` and then update the check below (line 838)?

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+                        "%s in T info field has empty description in VCF header!", field
+                    )
+                    temp_missing_descriptions.append(field)
+            except KeyError:
+                logger.warning(
+                    "%s in T info field does not exist in VCF header!", field
+                )
+                # NOTE: some hists are not exported, so ignoring here
+                # END entry is also not exported (removed during densify)
+                if (field not in hist_fields) and (field != "END"):
```
ch-kr, line 838: `field != "END"` only needs to be checked if the input is a MT, right? also hopefully the `(field not in hist_fields)` check can be removed

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+
+    All annotations must be within an info struct, e.g. t.info.AC-raw.
+
+    :param t: Input MatrixTable or Table containing variant annotations to check.
+    :param subsets: List of subsets to be checked.
+    :param missingness_threshold: Upper cutoff for allowed amount of missingness. Default is 0.5.
+    :param monoallelic_expr: When passed, log how many monoallelic sites are in the Table.
+    :param verbose: If True, display top values of relevant annotations being checked, regardless of whether check conditions are violated; if False, display only top values of relevant annotations if check conditions are violated.
+    :param show_percent_sites: Show percentage of sites that fail checks. Default is False.
+    :param delimiter: String to use as delimiter when making group label combinations.
```
ch-kr, line 895:
```suggestion
    :param delimiter: String to use as delimiter when making group label combinations. Default is "-".
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param metric_first_label: If True, metric precedes label group, e.g. AC-afr-male. If False, label group precedes metric, afr-male-AC.
```
ch-kr, line 896:
```suggestion
    :param metric_first_label: If True, metric precedes label group, e.g. AC-afr-male. If False, label group precedes metric, afr-male-AC. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param sexes: List of sexes in table.
```
ch-kr, line 897:
```suggestion
    :param sexes: List of sexes in table. Default is SEXES.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param sample_sum_sets_and_pops: Dict with subset (keys) and populations within subset (values) for sample sum check. An empty string, e.g. "", should be passed as key for entire callset.
```
ch-kr, line 898:
```suggestion
    :param sample_sum_sets_and_pops: Dict with subset (keys) and populations within subset (values) for sample sum check. An empty string, e.g. "", should be passed as key for entire callset. Default is {"": POPS}.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param sort_order: List containing order to sort label group combinations. Default is SORT_ORDER.
+    :param variant_filter_field: String of variant filtration used in the filters annotation on `ht` (e.g. RF, VQSR, AS_VQSR).
```
ch-kr, line 900:
```suggestion
    :param variant_filter_field: String of variant filtration used in the filters annotation on `ht` (e.g. RF, VQSR, AS_VQSR). Default is "RF".
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param summarize_variants_check: When true, runs the summarize_variants method.
```
ch-kr, line 901:
```suggestion
    :param summarize_variants_check: When true, runs the summarize_variants method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param filters_check: When true, runs the filters_sanity_check method.
```
ch-kr, line 902:
```suggestion
    :param filters_check: When true, runs the filters_sanity_check method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param raw_adj_check: When true, runs the raw_and_adj_sanity_checks method.
```
ch-kr, line 903:
```suggestion
    :param raw_adj_check: When true, runs the raw_and_adj_sanity_checks method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param subset_freq_check: When true, runs the subset_freq_sanity_checks method.
```
ch-kr, line 904:
```suggestion
    :param subset_freq_check: When true, runs the subset_freq_sanity_checks method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param samples_sum_check: When true, runs the sample_sum_sanity_checks method.
```
ch-kr, line 905:
```suggestion
    :param samples_sum_check: When true, runs the sample_sum_sanity_checks method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param sex_chr_check: When true, runs the sex_chr_sanity_checks method.
```
ch-kr, line 906:
```suggestion
    :param sex_chr_check: When true, runs the sex_chr_sanity_checks method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param missingness_check: When true, runs the missingness_sanity_checks method.
```
ch-kr, line 907:
```suggestion
    :param missingness_check: When true, runs the missingness_sanity_checks method. Default is True.
```

ch-kr (changes requested):
mostly small suggestions. also, I took another look at the v2 code (https://github.com/broadinstitute/gnomad_qc/blob/master/gnomad_qc/v2/variant_qc/prepare_data_release.py#L297) and was wondering if we should add the histograms sanity check back? it looks like the v2 code just double checked to make sure the n_smaller and n_larger fields were 0 as a sanity check

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param check_lcr: Add the low complexity region's boolean expression to the problematic region check.
+    :param check_segdup: Add the segmental duplication region's boolean expression to the problematic region check.
+    :param check_nonpar: Add the non-pseudoautosomal region's boolean expression to the problematic region check.
+    :param check_decoy: Add the decoy region's boolean expression to the problematic region check.
+    :return: None
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
+
+    filters = t.aggregate(hl.agg.counter(t.filters))
+    logger.info("hl.agg.counter filters: %s", filters)
```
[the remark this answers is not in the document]
ch-kr: I like the new output more (the previous code did an `explode` before counting) but don't have super strong feelings about this

gnomad/assessment/sanity_checks.py
ch-kr: I wondered about this too during my review. these two numbers are only used for the slightly more complex shows (calling `_filter_agg_order` with two or more `group_exprs`) and are slightly arbitrary (I took these directly from the v2 code). I guess we could make these parameters to `filters_sanity_check`? maybe something like `large_n_rows: int = 50` and `large_n_cols: int = 140`, with this in the docstring
```
:param large_n_rows: Number of rows to show when showing percentages of filtered variants grouped by multiple conditions. Default is 50.
:param large_n_cols: Number of columns to show when showing percentages of filtered variants grouped by multiple conditions. Default is 140.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+        }
+
+    # Check raw AN > 0
+    check_field = f"AN{delimiter}raw"
+    field_check_expr[f"{check_field} > 0"] = {
+        "expr": hl.agg.count_where(t.info[check_field] <= 0),
+        "display_fields": hl.struct(**{check_field: t.info[check_field]}),
+    }
+
+    # Check adj AN >= 0
```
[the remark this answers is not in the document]
mike-w-wilson: Similar to above.

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+        "Dataset has %d variants distributed across the following contigs: %s",
+        var_summary.n_variants,
+        var_summary.contigs,
+    )
+
+    for contig in var_summary.contigs:
+        if var_summary.contigs[contig] == 0:
+            logger.warning("%s has no variants called", var_summary.contigs)
+
+    if monoallelic_expr is not None:
```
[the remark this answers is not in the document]
mike-w-wilson: Yeah, that is probably a better place for it.

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+            ),
+        )
+    )
+    logger.info("Frequency spot check counts: %s", hl.eval(freq_counts))
+
+
+def sample_sum_sanity_checks(
+    t: Union[hl.MatrixTable, hl.Table],
+    sexes: List[str] = SEXES,
+    sample_sum_sets_and_pops: Dict[str, List[str]] = {"": POPS},
```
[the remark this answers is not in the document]
mike-w-wilson: Yeah, I thought this was a bit more general but understand it requires the user to build the dictionary with something like `sample_sum_sets_and_pops={subset:POPS for subset in SUBSET_LIST_FOR_VCF})`. So you would prefer a subpops parameter and a default pops parameter and any subsets that require their own pops can be passed in the existing dictionary? Then just check what was passed to determine how to handle them?

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+            ),
+            total_defined_AC_raw=hl.agg.count_where(
+                hl.is_defined(t.info[f"AC{delimiter}raw"])
+            ),
+        )
+    )
+    logger.info("Frequency spot check counts: %s", hl.eval(freq_counts))
+
+
+def sample_sum_sanity_checks(
```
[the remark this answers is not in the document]
mike-w-wilson: hmm, I tried using the "sanity" in the main functions but now I'm reconsidering that as since they will all live in a sanity check module. I think I need to think on this one a bit more but will propose alternatives. Maybe the method above could be sum_group_callstats or group_callstat_check?

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+                                check_field_right: t.info[check_field_right],
+                            }
+                        ),
+                    }
+
+    generic_field_check_loop(
+        t, field_check_expr, verbose, show_percent_sites=show_percent_sites,
+    )
+
+    freq_counts = t.aggregate(
```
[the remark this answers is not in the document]
mike-w-wilson: I don't think this is needed. Missingness check should catch any missing sites and I dont see how the adj callstats could have missing values while raw does not or vice versa in gnomAD.

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+        hl.struct(
+            total_defined_AC=hl.agg.count_where(
+                hl.is_defined(t.info[f"AC{delimiter}adj"])
+            ),
+            total_defined_AC_raw=hl.agg.count_where(
+                hl.is_defined(t.info[f"AC{delimiter}raw"])
+            ),
+        )
+    )
+    logger.info("Frequency spot check counts: %s", hl.eval(freq_counts))
```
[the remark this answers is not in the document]
mike-w-wilson: Correct!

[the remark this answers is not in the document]
mike-w-wilson:
```suggestion
    logger.info("Frequency spot check counts: %s", freq_counts)
```

mike-w-wilson (commented):
I've gone through the comments and there are a few things I was hoping to discuss more. I also am going to commit almost all of the small docstring comments and Julia's simplifying field check to make this a bit less crowded.

gnomad/assessment/sanity_checks.py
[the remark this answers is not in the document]
mike-w-wilson: I can update this to include both counts, with and without the explode.

gnomad/resources/config.py
```diff
@@ -17,7 +17,7 @@ class GnomadPublicResourceSource(Enum):
 class _GnomadPublicResourceConfiguration:
     """Configuration for public gnomAD resources."""
 
-    __source: Union[
+    __source: Union[  # pylint: disable=unused-private-member
```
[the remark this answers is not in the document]
ch-kr: ooh, thanks!

gnomad/assessment/sanity_checks.py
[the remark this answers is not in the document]
ch-kr: good point about the sanity check module haha. I like sum_group_callstats!

gnomad/assessment/sanity_checks.py
```diff
@@ -190,3 +212,744 @@ def compare_row_counts(ht1: hl.Table, ht2: hl.Table) -> bool:
...
+    :param info_metrics: List of metrics in info struct of input Table.
+    :param contigs: List of contigs present in input Table.
+    :param verbose: If True, show top values of annotations being checked, including checks that pass; if False, show only top values of annotations that fail checks.
+    :param delimiter: String to use as the delimiter in XX metrics
+    :return: None
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
+
+    xx_metrics = [
+        x for x in info_metrics if f"{delimiter}female" in x or f"{delimiter}XX" in x
```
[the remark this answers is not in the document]
jkgoodrich: Move to just XX

gnomad/assessment/sanity_checks.py
ch-kr: I think it would make sense since we already have pieces of the 'format' field check in the ukbb repo, and`hl.export_vcf` doesn't require sites vcfs to be mts with empty cols anymore (so hopefully users will be exporting hts instead of mts if they have sites files)

mike-w-wilson:
@ch-kr and @jkgoodrich , I've gone through and made the larger changes. I added back a default pop and subset for the sample sum checks. Now the function can accept a list of subsets, it will apply the default pops, and then combined it with the subset specific pops if provided. I've also changed all he hard coded metrics to passed parameters where KC pointed out and one other function. The filter count now has an option to explode the filters and aggregate to give single filter counts. I also changed some function names to what I believe is more representative of the check and removed the redundant "sanity_check" when a function did a single thing. I struggled with better names on other functions like `raw_and_adj_sanity_checks` so suggestions welcome if you dont like it!

jkgoodrich (changes requested):
Just a few more small things, and `make_group_sum_expr_dict` looks very different from before so I think something is wrong with it (which could be my fault because I think I had a comment on this previously so I wonder if I messed it up with that comment), but almost there in my opinion. One other thing is KC's comment about histogram checks: " I took another look at the v2 code (https://github.com/broadinstitute/gnomad_qc/blob/master/gnomad_qc/v2/variant_qc/prepare_data_release.py#L297) and was wondering if we should add the histograms sanity check back? it looks like the v2 code just double checked to make sure the n_smaller and n_larger fields were 0 as a sanity check"

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,850 @@ def make_filters_sanity_check_expr(
...
+        :param n_cols: Number of columns to show.
+        :param extra_filter_checks: Optional dictionary containing filter condition name (key) and extra filter expressions (value) to be examined.
+        :param variant_filter_field: String of variant filtration used in the filters annotation on `ht` (e.g. RF, VQSR, AS_VQSR). Default is "RF".
+        :return: None
+        """
+        t = t.rows() if isinstance(t, hl.MatrixTable) else t
+        # NOTE: make_filters_expr_dict returns a dict with %ages of variants filtered
+        t.group_by(**group_exprs).aggregate(
+            **make_filters_expr_dict(t, extra_filter_checks, variant_filter_field)
+        ).order_by(hl.desc("n")).show(n_rows, n_cols)
```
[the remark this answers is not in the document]
mike-w-wilson: Yes you are right!

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,832 @@ def make_filters_sanity_check_expr(
...
+        else default_pop_subset
+    )
+    for subset, pops in sample_sum_sets_and_pops.items():
+        pop_names = pops
+
+        # We do not store the raw callstats for the pop or sex groupings of any subset so only check adj here.
+        field_check_expr_s = make_group_sum_expr_dict(
+            t,
+            subset,
+            dict(group=["adj"], pop=pop_names),
```
ch-kr, line 476: maybe `group=["adj"]` should be an argument to this function?

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,832 @@ def make_filters_sanity_check_expr(
...
+    :param sum_metrics: List of metrics to sum and compare to annotationed versions and between subsets and entire callset. Default is ["AC", "AN", "nhomalt"].
+    :param sexes: List of sexes in table. Default is SEXES.
+    :param sample_sum_sets_and_pops: Dict with subset (keys) and populations within subset (values) for sample sum check.
+    :param sort_order: List containing order to sort label group combinations. Default is SORT_ORDER.
+    :param variant_filter_field: String of variant filtration used in the filters annotation on `ht` (e.g. RF, VQSR, AS_VQSR). Default is "RF".
+    :param problematic_regions: List of regions considered problematic to run filter check in. Default is ["lcr", "segdup", "nonpar"].
+    :param single_filter_count: If True, explode the Table's filter column and give a supplement total count of each filter. Default is False.
+    :param summarize_variants_check: When true, runs the summarize_variants method. Default is True.
+    :param filters_check: When true, runs the summarize_variant_filters method. Default is True.
```
ch-kr, line 880:
```suggestion
    :param filters_check: When True, runs the summarize_variant_filters method. Default is True.
```
(just for consistency with the above `If True`s)

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,832 @@ def make_filters_sanity_check_expr(
...
+    :param raw_adj_check: When true, runs the raw_and_adj_sanity_checks method. Default is True.
```
ch-kr, line 881:
```suggestion
    :param raw_adj_check: When True, runs the raw_and_adj_sanity_checks method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,832 @@ def make_filters_sanity_check_expr(
...
+    :param subset_freq_check: When true, runs the compare_subset_freqs method. Default is True.
```
ch-kr, line 882:
```suggestion
    :param subset_freq_check: When True, runs the compare_subset_freqs method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,832 @@ def make_filters_sanity_check_expr(
...
+    :param samples_sum_check: When true, runs the sum_group_callstats method. Default is True.
```
ch-kr, line 883:
```suggestion
    :param samples_sum_check: When True, runs the sum_group_callstats method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,832 @@ def make_filters_sanity_check_expr(
...
+    :param sex_chr_check: When true, runs the sex_chr_sanity_checks method. Default is True.
```
ch-kr, line 884:
```suggestion
    :param sex_chr_check: When True, runs the sex_chr_sanity_checks method. Default is True.
```

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,832 @@ def make_filters_sanity_check_expr(
...
+    :param missingness_check: When true, runs the compute_missingness method. Default is True.
```
ch-kr, line 885:
```suggestion
    :param missingness_check: When True, runs the compute_missingness method. Default is True.
```

ch-kr (changes requested):
a few more small things

mike-w-wilson:
In regards to the histogram check. I believe all qual hist n_larger and n_smaller are not unfurled before vcf export looking at the release code in review here: https://github.com/broadinstitute/gnomad_qc/blob/c6eed221dd4cbc12df025cd94f0165bafa6eac40/gnomad_qc/v3/create_release/prepare_vcf_data_release.py#L483.  This entire module relies on annotations existing in the info field so I think the only four fields we unfurl and keep are the `age_hist_het_n_smaller`, `age_hist_het_n_larger`, `age_hist_hom_n_smaller`, and `age_hist_hom_n_larger`. I can add the !=0 check to these four fields, if that makes sense to you both?

ch-kr:
that's a good point -- the qual hists n_smaller/larger aren't unfurled or exported (with the exception of the n_larger DP hist). the check for those hist fields would have to be run on the nested release HT, so maybe that is out of the scope of this PR

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,850 @@ def make_filters_sanity_check_expr(
...
+            - Only inbreeding coefficient filter
+            - Only AC0 filter
+            - Only `variant_filter_field` filtering
+
+    :param t: Input MatrixTable or Table to be checked.
+    :param variant_filter_field: String of variant filtration used in the filters annotation on `ht` (e.g. RF, VQSR, AS_VQSR). Default is "RF".
+    :param problematic_regions: List of regions considered problematic to run filter check in. Default is ["lcr", "segdup", "nonpar"].
+    :param single_filter_count: If True, explode the Table's filter column and give a supplement total count of each filter. Default is False.
+    :param monoallelic_expr: Optional boolean expression of monoallelic status that logs how many monoallelic sites are in the Table.
+    :param large_n_rows: Number of rows to show when showing percentages of filtered variants grouped by multiple conditions. Default is 50.
```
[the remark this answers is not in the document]
mike-w-wilson: The downstream code had these defaults but it they removed via a suggestion. I'll revert the change.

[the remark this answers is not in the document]
jkgoodrich: Sorry that was my fault, I see now why you wanted None for some of them

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,832 @@ def make_filters_sanity_check_expr(
...
+
+    filters = t.aggregate(hl.agg.counter(t.filters))
+    logger.info("Variant filter counts: %s", filters)
+
+    if single_filter_count:
+        ex_t = t.explode(t.filters)
+        filters = ex_t.aggregate(hl.agg.counter(ex_t.filters))
+        logger.info("Exploded variant filter counts: %s", filters)
+
+    if monoallelic_expr is not None:
```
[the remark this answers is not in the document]
mike-w-wilson: hmm good question. The gnomAD dataset doesn't include monoallelic as a filter, it is a separate annotation but this would still be doable. @jkgoodrich , what do you think?

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,850 @@ def make_filters_sanity_check_expr(
...
-                ),
-                f"{prefix}{subfield}_{group} = sum({subfield}_{group}_{alt_groups})",
-                [
-                    f"info.{prefix}{subfield}_{group}",
-                    f"sum_{subfield}_{group}_{alt_groups}",
-                ],
-                verbose,
-            )
+        annot_dict[
+            f"sum{delimiter}{metric}{delimiter}{group}{delimiter}{sum_group}"
```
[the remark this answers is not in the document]
jkgoodrich: Ah I see, the previous message was actually not what was really being checked because it didn't include the subset info, so now it is more accurate. As for the repeated displays, if there is now way around the repeat then I would prefer to keep it faster and have the repeat.

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,850 @@ def make_filters_sanity_check_expr(
...
-    :param verbose: If True, show top values of annotations being checked, including checks that pass; if False,
-        show only top values of annotations that fail checks.
-    :param subpop: Subpop abbreviation, supplied only if subpopulations are included in the annotation groups being checked.
+    :param t: Input MatrixTable or Table containing call stats annotations to be summed.
+    :param subset: String indicating sample subset.
+    :param label_groups: Dictionary containing an entry for each label group, where key is the name of the grouping, e.g. "sex" or "pop", and value is a list of all possible values for that grouping (e.g. ["XY", "XX"] or ["afr", "nfe", "amr"]).
     :param sort_order: List containing order to sort label group combinations. Default is SORT_ORDER.
-    :return: None
+    :param delimiter: String to use as delimiter when making group label combinations. Default is "-".
+    :param metric_first_label: If True, call statistic precedes label group, e.g. AC-subset-male. If False, label group precedes metric, subset-AC-male. Default is True.
```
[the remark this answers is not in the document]
mike-w-wilson: Yeah I think the variable name is actually misleading since `label` is used in the function but has nothing to do with the variable `metric_first_label`. I propose this is changed to `metric_first_field` to better represent what is going on and I will update the docstring again to better explain it. Thoughts?

gnomad/assessment/sanity_checks.py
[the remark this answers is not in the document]
jkgoodrich: Actually, I think there might be an extra indent on generic_field_check_loop(t, field_check_expr, verbose)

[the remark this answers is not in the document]
mike-w-wilson: You're right!

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,836 @@ def make_filters_sanity_check_expr(
...
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
+
+    field_check_expr = {}
+    for subset in subsets:
+        if subset:
+            for metric in metrics:
+                for group in ["adj", "raw"]:
+                    logger.info(
+                        "Comparing subset %s frequencies to entire callset", subset
```
jkgoodrich, line 399: Maybe move this up before metric for loop, or add the metric into the logger

gnomad/assessment/sanity_checks.py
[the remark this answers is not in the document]
jkgoodrich: I think we can leave as is since we decided not to actually have this as a filter and we can revisit if we decide to change that for future callsets

gnomad/assessment/sanity_checks.py
```diff
@@ -103,90 +117,850 @@ def make_filters_sanity_check_expr(
...
+    )
+
+    for contig in var_summary.contigs:
+        if var_summary.contigs[contig] == 0:
+            logger.warning("%s has no variants called", var_summary.contigs)
+
+    return var_summary
+
+
+def raw_and_adj_sanity_checks(
```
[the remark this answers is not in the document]
jkgoodrich: `check_raw_and_adj_callstats` sounds good and `check_sex_chr_metrics`?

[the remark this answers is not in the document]
mike-w-wilson: Sounds good to me.

mike-w-wilson:
@ch-kr  and @jkgoodrich , a small number of changes around function names, docstrings, variables, and comments. I tried cleaning up the `make_group_sum_expr_dict` function a bit more and removed an unecessary variable and also added comments per KC's suggestion. I also made ["adj"] a param for sum_group_stats. I think it's getting very close. I imagine there will be some comments on comments but after that, I will add to the changelog and rename the function to `validty_checks.py`

gnomad/assessment/sanity_checks.py
```diff
@@ -384,7 +393,7 @@ def compare_subset_freqs(
     :param verbose: If True, show top values of annotations being checked, including checks that pass; if False, show only top values of annotations that fail checks.
     :param show_percent_sites: If True, show the percentage and count of overall sites that fail; if False, only show the number of sites that fail.
     :param delimiter: String to use as delimiter when making group label combinations. Default is "-".
-    :param metric_first_label: If True, metric precedes label subset, e.g. AC-non_v2-XY. If False, subset precedes metric, non_v2-AC-XY. Default is True.
+    :param metric_first_field: If True, metric precedes subset, e.g. AC-non_v2-. If False, subset precedes metric, non_v2-AC-XY. Default is True.
```
jkgoodrich, line 396: Is this supposed to have an XY?

```suggestion
    :param metric_first_field: If True, metric precedes subset, e.g. AC-non_v2-XY. If False, subset precedes metric, non_v2-AC-XY. Default is True.
```

jkgoodrich (approved):
I only have one very tiny doc change, but I think this looks great to me. Amazing work Mike!

gnomad/assessment/sanity_checks.py
mike-w-wilson: Ah yes originally! I definitely got sidetracked here and forgot to go back.  I was trying to give an example for the types of comparisons the module does with the addition of the XY but I removed the a similar label in `make_group_sum_expr_dict`, which I think reduces confusion. It read 
```
:param metric_first_field: If True, metric precedes subset in the Table's fields, e.g. AC-hgdp. If False, subset precedes metric, hgdp-AC. Default is True.
``` 
Do you think removing the last -XY would reduce confusion here as well? I couldnt decide if the exta label. -XY, was helpful.

jkgoodrich: yeah, maybe just remove both, I think that gets the point across

ch-kr (changes requested):
so close!! just a few more tiny nitpicks

ch-kr (approved):
LGTM, this comment (https://github.com/broadinstitute/gnomad_methods/pull/389#discussion_r675681995) is really minor, so I'll leave it up to you if you want to change

mike-w-wilson:
Closing this PR. #395 is the same code with a cleaner commit history.

PR #394, opened 2021-07-12
New file exists

ch-kr:
Added function to check if large number of files exist (current function in repo is too slow to check more than a few files). Dan King wrote this function for when I needed to check a large number of files in a batch job


gnomad/utils/file_utils.py
```diff
@@ -6,15 +6,69 @@
...
+
+                    return unapplied_function
+
+                file_existence_checks = [
+                    create_unapplied_function(fname) for fname in fnames
+                ]
+                file_existence = await bounded_gather(
+                    *file_existence_checks, parallelism=750
+                )
+    return {fname: exists for fname, exists in zip(fnames, file_existence)}
```
jkgoodrich, line 69: can this just be:
```suggestion
    return dict(zip(fnames, file_existence))
```

gnomad/utils/file_utils.py
```diff
@@ -6,15 +6,69 @@
...
+    Check whether a large number of files exist.
+
+    Created for use with hail Batch jobs.
+    Normal `file_exists` function is very slow when checking a large number of files.
+
+    :param fnames: List of file names to check.
+    :return: Dictionary of file names (str) and whether the file exists (boolean).
+    """
+
+    async def low_level_async_file_exists(fs: AsyncFS, url: str):
```
jkgoodrich, line 33: Is there any reason to add documentation to this? I know it is an inner function so I don't think it is needed, but we have typically added them in the past. I'm fine either way

gnomad/utils/file_utils.py
```diff
@@ -6,15 +6,69 @@
...
+        try:
+            await fs.statfile(url)
+        except FileNotFoundError:
+            return False
+        else:
+            return True
+
+    async def async_file_exists(fs: AsyncFS, fname: str):
```
jkgoodrich, line 41: Same doc comment as above

gnomad/utils/file_utils.py
```diff
@@ -6,15 +6,69 @@
...
+
+    with tqdm(
+        total=len(fnames), desc="check files for existence", disable=False
+    ) as pbar:
+        with ThreadPoolExecutor() as thread_pool:
+            async with RouterAsyncFS(
+                "file", [LocalAsyncFS(thread_pool), GoogleStorageAsyncFS()]
+            ) as fs:
+
+                def create_unapplied_function(fname):
```
jkgoodrich, line 55: Maybe add a comment for why an unapplied function is needed?

gnomad/utils/file_utils.py
```diff
@@ -6,15 +6,69 @@
...
+                        pbar.update(1)
+                        return x
+
+                    return unapplied_function
+
+                file_existence_checks = [
+                    create_unapplied_function(fname) for fname in fnames
+                ]
+                file_existence = await bounded_gather(
+                    *file_existence_checks, parallelism=750
```
jkgoodrich, line 67: Make `750` a parameter?

jkgoodrich (commented):
Will work as is, just a couple suggestions

gnomad/utils/file_utils.py
```diff
@@ -6,15 +6,93 @@
...
+                    :param fname: Path to file to check.
+                    :return: Function that checks for file existence and updates progress bar.
+                    """
+
+                    async def unapplied_function():
+                        x = await async_file_exists(fs, fname)
+                        pbar.update(1)
+                        return x
+
+                    return unapplied_function
```
danking, line 85: A bit of context on this rigamarole:

When using asyncio, it's important to not have too many live coroutines. Every time you call an async function, you create a coroutine. For example, calling `create_unapplied_function('foo')` does not create a coroutine because we never invoke an async function. On the other hand, `create_unapplied_function('foo')()` *does* create a coroutine because it invokes the returned async function. A coroutine completes when we `await` it (or use one of the other asyncio waiting primitives, like `asyncio.wait`).

For this reason, `bounded_gather` accepts async functions as arguments instead of directly accepting coroutines. For example:
```python3
# this creates a coroutine for each argument to bad_bounded_gather
await bad_bounded_gather(async_do_something('argument'), ...)

# this allows bounded gather to bound the number of live coroutines (by waiting for
# previous coroutines to complete before invoking more)
async def async_delayed_do_something():
    return await async_do_something('argument')
await bounded_gather(async_delayed_do_something, ...)
```

In the first case, we create all the coroutines *before even calling `bad_bounded_gather`*. In the second case, `bounded_gather` can decide when to create coroutines (by invoking one of its arguments).

---

Usually, Hail team uses this pattern:

```python3
await bounded_gather(*[functools.partial(async_file_exists, fs, fname)
                       for fname in fnames],
                     parallelism=parallelism)
```

But we also want to update the progress bar after we check for file existence. I should maybe have called `create_unapplied_function` `check_existence_and_update_pbar_thunk`. I use the term "thunk" to refer to these zero-argument functions that delay the creation of coroutines.

jkgoodrich: Thank you so much for this detailed description Dan!!

ch-kr: seconded, thank you for this!! I added a small note to the docstring, but let me know if you think I should add a longer summary @jkgoodrich

jkgoodrich (approved):
LGTM

danking (approved):
LGTM!

PR #397, opened 2021-07-29
Run pylint on gnomad_qc in PR checks

### LINKED ISSUE #396: Run pylint on gnomad_qc in PR checks
[gnomad_qc](https://github.com/broadinstitute/gnomad_qc) depends on the `master` branch of this repository.
https://github.com/broadinstitute/gnomad_qc/blob/d741289569e0faf0a6d065883ea7fde82cdb0f1e/requirements.txt#L3

Thus, it's easy to accidentally break gnomad_qc code by making changes to gnomad_methods. Furthermore, those breakages likely will not be noticed immediately and the next time someone works with the relevant part of gnomad_qc, they'll encounter unexpected problems.

An example of this was seen recently, where a file renamed in #395 broke an import in gnomad_qc. This was only found in [PR checks on an unrelated change](https://github.com/broadinstitute/gnomad_qc/runs/3182396028) to gnomad_qc, where it had to be dealt with in order to get PR checks to pass (https://github.com/broadinstitute/gnomad_qc/pull/209/commits/a767271b4850109009d7f1e74db5b19a3b1627a7).

We can identify some issues, such as broken imports caused by renamed files or bugs caused by changes to function interfaces, by adding a PR check that runs pylint on gnomad_qc using the modified version of gnomad_methods in the PR. These issues should be easiest to fix when they are immediately brought to the attention of the person that has the most context for the change (the PR author).

nawatts:
For gnomad_methods PRs, run Pylint on gnomad_qc using the version of gnomad_methods from the PR. This quickly identifies cases where a change to gnomad_methods breaks gnomad_qc. Currently, such cases are often not found until later when someone is working on an unrelated change to gnomad_qc.

As seen from the checks on https://github.com/broadinstitute/gnomad_methods/pull/397/commits/683d2e660fdc26c8892e6c2186987cf338f00487 ([results](https://github.com/broadinstitute/gnomad_methods/pull/397/checks?check_run_id=3195395459)), this can catch some issues caused by things like renamed modules or changed function interfaces.

Note that this job is not required to pass in order to merge a PR. This still allows breaking changes to be merged, but provides the author with a notice that they should update gnomad_qc accordingly.

Resolves #396


.github/workflows/ci.yml
```diff
@@ -89,3 +89,37 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout gnomad_methods
+        uses: actions/checkout@v2
+        with:
+          path: gnomad_methods
+      - name: Checkout gnomad_qc
+        uses: actions/checkout@v2
+        with:
+          repository: broadinstitute/gnomad_qc
```
jkgoodrich, line 103: For my knowledge, why does this one need the repository, but gnomad_methods doesn't. Is it because this is within gnomad_methods?

jkgoodrich (approved):
Thanks for adding Nick! This LGTM. Just one question for my knowledge

.github/workflows/ci.yml
nawatts: Correct. If no repository is specified, the checkout action uses the one in the workflow's context.

* https://github.com/actions/checkout#usage
* https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

jkgoodrich: Great, thank you!

PR #398, opened 2021-08-04
Added util to call parallel_file_exists

ch-kr:


nawatts:
Since we generally don't use async functions in pipelines, should the synchronous wrapper get the `parallel_file_exists` name instead?

```python
async def parallel_file_exists_async(fnames):
    ...

def parallel_file_exists(fnames):
    return asyncio.get_event_loop().run_until_complete(parallel_file_exists_async(fnames))
```

gnomad/utils/file_utils.py
```diff
@@ -85,6 +86,17 @@ async def unapplied_function():
     return dict(zip(fnames, file_existence))
 
 
+def parallel_file_exists(fnames: List[str]) -> Dict[str, bool]:
+    """
+    Call `parallel_file_exists_async` to check whether large number of files exist.
+
+    :param fnames: List of file names to check.
```
bw2, line 93: Would add a sentence on whether these can be local or gs:// file paths (or both). 
Also maybe rename fnames => fpaths?

bw2 (changes requested):

bw2 (approved):

gnomad/utils/file_utils.py
```diff
@@ -19,70 +20,83 @@
...
-    return dict(zip(fnames, file_existence))
+    return dict(zip(fpaths, file_existence))
+
+
+def parallel_file_exists(fpaths: List[str]) -> Dict[str, bool]:
+    """
+    Call `parallel_file_exists_async` to check whether large number of files exist.
+
+    :param fpaths: List of file paths to check. Files can be in local or Google cloud storage.
+    :param parallelism: Integer that sets parallelism of file existence checking task. Default is 750.
```
bw2, line 94: Just noticed the function doesn't have a parallelism arg..

PR #399, opened 2021-08-10
Sanity check fix for subsets

jkgoodrich:
Found a small issue when testing on the HGDP + TGP subset VCF export


mike-w-wilson (approved):
Good catch!

Issue #402, opened 2021-08-13
Resources that do not exist in gnomAD buckets


nawatts:
Some resources no longer exist in the gnomAD buckets (gnomad-public and gnomad-public-requester-pays).

Some examples:
* [2.1 PCA](https://github.com/broadinstitute/gnomad_methods/blob/a389dfb1f0c0e7ff9a4dc7c05942474348f0d15a/gnomad/resources/grch37/gnomad.py#L85-L93)
* Release VCFs ([v2](https://github.com/broadinstitute/gnomad_methods/blob/a389dfb1f0c0e7ff9a4dc7c05942474348f0d15a/gnomad/resources/grch37/gnomad.py#L206-L216) and [v3](https://github.com/broadinstitute/gnomad_methods/blob/a389dfb1f0c0e7ff9a4dc7c05942474348f0d15a/gnomad/resources/grch38/gnomad.py#L353-L364))
* [Coverage TSVs](https://github.com/broadinstitute/gnomad_methods/blob/a389dfb1f0c0e7ff9a4dc7c05942474348f0d15a/gnomad/resources/grch38/gnomad.py#L323-L350)

The PCA files are ok, since they are loaded through the [GnomadPublicResource class](https://github.com/broadinstitute/gnomad_methods/blob/a389dfb1f0c0e7ff9a4dc7c05942474348f0d15a/gnomad/resources/resource_utils.py#L380-L463) and can thus be loaded from cloud provider buckets, where they do exist.

However, the VCF and coverage TSV URLs do not go through GnomadPublicResource, so they are currently broken.

nawatts:
I suspect the VCF and coverage TSV paths may only be used in creating a release. There's not much reason for users of this library to read those files when the data is available is available in Hail format. We could possibly move them from here to the appropriate scripts used to create a release.

nawatts:
> The PCA files are ok, since they are loaded through the GnomadPublicResource class and can thus be loaded from cloud provider buckets, where they do exist.

Anyway, the long term plan is not to maintain any files in gnomAD buckets, so this case has to be supported. #264 is the long term solution and #390 a short term improvement.

PR #403, opened 2021-08-13
Update error message for when resource is not found in gnomAD bucket

### LINKED ISSUE #390: Improve error message if resource is not available from gnomAD buckets
As pointed out in https://github.com/broadinstitute/gnomad_methods/pull/388#discussion_r659829048, the contents of the `gnomad-public`/`gnomad-public-requester-pays` buckets will change over time. Older versions of this package may point to resources that no longer exist. We can improve the error message for those cases and suggest that users update to the latest version.

nawatts:
Currently, if a resource is not available from the user's selected source, we show an error message directing them to use the gnomAD buckets. However, going forward, we won't host all resource files in our buckets once they are copied to cloud providers' buckets (this is already the case with VCFs).

This generalizes the error message to "try a different source" and adds instructions on how to find a list of all available sources.

Resolves #390
Related to #402

### DIFF
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -397,7 +397,7 @@ class GnomadPublicResource(BaseResource, ABC):
                 resource_source = gnomad_public_resource_configuration.source
                 if not self.is_resource_available():
                     if resource_source == GnomadPublicResourceSource.GNOMAD:
-                        message = "This resource is not currently available from the gnomAD project."
+                        message = "This resource is not currently available from the gnomAD project public buckets."
                     elif isinstance(resource_source, GnomadPublicResourceSource):
                         message = f"This resource is not currently available from {resource_source.value}."
                     else:
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -395,17 +395,23 @@ class GnomadPublicResource(BaseResource, ABC):
                 # If one of the known sources is selected, check if the resource is available.
                 # For custom sources, skip the check and attempt to read the resource.
                 resource_source = gnomad_public_resource_configuration.source
-                if (
-                    isinstance(resource_source, GnomadPublicResourceSource)
-                    and resource_source != GnomadPublicResourceSource.GNOMAD
-                ):
-                    if not self.is_resource_available():
-                        raise ResourceNotAvailable(
-                            f"This resource is not currently available from {resource_source.value}.\n\n"
-                            "To load resources directly from gnomAD instead, use:\n\n"
-                            ">>> from gnomad.resources.config import gnomad_public_resource_configuration, GnomadPublicResourceSource\n"
-                            ">>> gnomad_public_resource_configuration.source = GnomadPublicResourceSource.GNOMAD"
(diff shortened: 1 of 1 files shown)


nawatts:
To test:

* Resource not found in gnomAD buckets.

   ```python
   from gnomad.resources.config import gnomad_public_resource_configuration, GnomadPublicResourceSource
   from gnomad.resources.grch37.gnomad import public_pca_loadings

   gnomad_public_resource_configuration.source = GnomadPublicResourceSource.GNOMAD

   ds = public_pca_loadings("").ht()
   ```

* Resource not found in Google's bucket.

   ```python
   from gnomad.resources.config import gnomad_public_resource_configuration, GnomadPublicResourceSource
   from gnomad.resources.grch37.reference_data import dbsnp

   gnomad_public_resource_configuration.source = GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS

   ds = dbsnp.ht() 
   ```

gnomad/resources/resource_utils.py
```diff
@@ -395,17 +395,23 @@ def read_resource(self, *args, **kwargs):
...
-                ):
-                    if not self.is_resource_available():
-                        raise ResourceNotAvailable(
-                            f"This resource is not currently available from {resource_source.value}.\n\n"
-                            "To load resources directly from gnomAD instead, use:\n\n"
-                            ">>> from gnomad.resources.config import gnomad_public_resource_configuration, GnomadPublicResourceSource\n"
-                            ">>> gnomad_public_resource_configuration.source = GnomadPublicResourceSource.GNOMAD"
-                        )
+                if not self.is_resource_available():
+                    if resource_source == GnomadPublicResourceSource.GNOMAD:
```
jkgoodrich, line 399: Just making sure I understand, if the resource_source is 'gnomAD' rather than `Google Cloud Public Datasets` then state that `This resource is not currently available from the gnomAD project.`. Maybe we should say `This resource is not currently available from the gnomAD project public buckets.` Or something that indicates we don't host it, but it is still in other locations that host gnomAD project data

gnomad/resources/resource_utils.py
```diff
@@ -395,17 +395,23 @@ def read_resource(self, *args, **kwargs):
...
+                        message = f"This resource is not currently available from {resource_source.value}."
+                    else:
+                        message = f"This resource is not currently available from {resource_source}."
+
+                    raise ResourceNotAvailable(
+                        f"{message}\n\n"
+                        "To load resources from a different source (for example, Google Cloud Public Datasets) instead, use:\n\n"
+                        ">>> from gnomad.resources.config import gnomad_public_resource_configuration, GnomadPublicResourceSource\n"
+                        ">>> gnomad_public_resource_configuration.source = GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS\n\n"
+                        "To get all available sources for gnomAD resources, use:\n\n"
```
jkgoodrich, line 411: This is great, thank you for adding!

jkgoodrich (approved):
Just a tiny thought on possible wording, but otherwise LGTM.

PR #404, opened 2021-08-18
Remove module imports from __init__.py files

nawatts:
I've been trying to figure out why Pylint reports no errors in gnomad_qc despite several v2 scripts importing a function that no longer exists (`gnomad.utils.slack.try_slack`). These module imports in `__init__.py` files seem to confuse Pylint.

I don't think they serve a purpose. It looks like I [suggested them](https://github.com/broadinstitute/gnomad_methods/pull/207#pullrequestreview-382012887) as an alternative to `from module import *`. But I don't think we have any code that relies on them by importing `*` from one of these packages.

### DIFF
--- a/gnomad/assessment/__init__.py
+++ b/gnomad/assessment/__init__.py
@@ -1,3 +1,1 @@
 # noqa: D104
-
-from gnomad.assessment import validity_checks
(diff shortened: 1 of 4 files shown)


mike-w-wilson (approved):
This makes sense and looks good to me!

PR #405, opened 2021-09-01
Add workflow to publish to PyPI

### LINKED ISSUE #356: Automate PyPI releases
To lower the barrier to releasing an update to PyPI, create a workflow to automate publishing to PyPI when a [GitHub Release](https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository) is created.

nawatts:
When a tag matching `v<VERSION_NUMBER>` is pushed to GitHub, this workflow will publish the tagged code to PyPI. This means that anyone with write access to this repository can publish a new release, as opposed to currently depending on the few team members with PyPI accounts.

Resolves #356

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,10 +11,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Setup Python
+      - name: Setup Python 3.8
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
       - name: Use pip cache
         uses: actions/cache@v2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,10 +41,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Setup Python
+      - name: Setup Python 3.8
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7
+          python-version: 3.8
       - name: Use pip cache
         uses: actions/cache@v1
         with:
@@ -102,10 +102,10 @@ jobs:
         with:
           repository: broadinstitute/gnomad_qc
           path: gnomad_qc
-      - name: Setup Python
+      - name: Setup Python 3.8
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
       - name: Use pip cache
         uses: actions/cache@v2
         with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,53 @@
+name: Publish to PyPI
+on:
+  push:
+    tags:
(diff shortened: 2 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+jobs:
+  publish-release:
+    name: Publish release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name:
```
jkgoodrich, line 13: does this need a name?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+          if [ "$package_version" != "$tag_version" ]; then
+            echo "Tag version (${tag_version}) does not match package version (${package_version})"
+            exit 1
+          fi
+
+          if ! grep -iq "version ${package_version}" CHANGELOG.md; then
+            echo "Missing changelog section for version ${package_version}"
+            exit 1
+          fi
+      - name: Setup Python
```
jkgoodrich, line 26:
```suggestion
      - name: Setup Python 3.7
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
jkgoodrich, line 29: Just a curiosity question, will there become a time that we want to update this python version?

jkgoodrich (approved):
This looks good, thank you so much for adding. Most of my comments are just to clarify things for me, and the addition of a comment at the top of the workflow maybe.

.github/workflows/publish.yml
nawatts: Oops. Step names aren't required, but omitting one here was not intentional.

.github/workflows/publish.yml
nawatts: I copied some of these steps from the PR checks workflow, which uses Python 3.7. Both workflows should probably be updated to use Python 3.8, since that is the version on Dataproc clusters.
https://cloud.google.com/dataproc/docs/tutorials/python-configuration#image_python_environments

.github/workflows/ci.yml
```diff
@@ -11,10 +11,10 @@ jobs:
...
         uses: actions/checkout@v2
-      - name: Setup Python
+      - name: Setup Python 3.8
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
       - name: Use pip cache
         uses: actions/cache@v2
         with:
```
jkgoodrich, line 20: I did notice 2 other 3.7's below this, do they need to be changed as well? under `Build documentation` and `gnomad_qc`

nawatts: Oops. Thanks for catching those.

jkgoodrich (approved):
LGTM!

PR #408, opened 2021-09-02
Document selecting resource source

nawatts:
Currently, if a resource is not available from one source, the error message describes how to select a different source (#388, #403). This provides that same information up front on the documentation website.


mike-w-wilson (approved):
LGTM

Issue #355, opened 2021-04-02
Improve changelog updates


nawatts:
The current process for updating the changelog has some issues.

* Updates in multiple PRs cause merge conflicts.
* Adding links to PRs requires another commit after opening a PR.

It works ok with a low volume of PRs, but merge conflicts become more of a problem with more PRs.

One alternative would be to update the changelog only on releases. However,
* PyPI releases are so infrequent that attempting to gather a list of changes and create a changelog at release time would be difficult.
* We use the latest git revision instead of the latest PyPI release in gnomad_qc so it's useful to have the changelog always up to date instead of only updated on PyPI releases.

More frequent releases could resolve those issues.

Another option would be storing the changelog outside of the repository content. https://github.com/marketplace/actions/release-drafter or something similar looks promising. Each release has its changelog and a draft release has a changelog for the latest development version. The changelog file could also be updated from the release notes on each release.

nawatts:
Perhaps the easiest thing to do is add changelog entries just before merging instead of when creating the PR.

jkgoodrich:
Agree that is a fine way to do it, but I do often forget, so I guess it's good to also encourage reviewers to hold on an approval without the changelog addition when it is needed so there are at least two checks on it.

nawatts:
Good point. We could also add a PR check that blocks merging unless the PR modifies the changelog or has something like "Changelog not needed" in the description or a comment.

PR #410, opened 2021-09-03
Clarify steps for tagging a new release

nawatts:
Clarify that release/version tags should be applied to a commit that is on the master branch and has been pushed to GitHub.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,6 +69,10 @@ https://packaging.python.org/guides/distributing-packages-using-setuptools/#pack
   git log $LAST_RELEASE_TAG..
   ```
 
+  Especially if there are many changes in the release, organize the changelog by type.
+
+  See https://keepachangelog.com/en/1.0.0/#how for more information.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -70,15 +70,22 @@ https://packaging.python.org/guides/distributing-packages-using-setuptools/#pack
   ```
 
 - Update version in setup.py, replace "unreleased" heading in changelog with the version number, and commit.
```
jkgoodrich, line 72: I noticed you also rearrange into, Fixed, Changed and Added, should that be done at this step, or should we encourage all of us to do this when adding to the change log?

jkgoodrich (approved):

CONTRIBUTING.md
nawatts: Good point. I've been going off the advice in https://keepachangelog.com/en/1.0.0/#how. I'll add a note here.

There's also #355 for general improvements to the changelog process.

jkgoodrich (approved):
LGTM!

PR #421, opened 2021-11-17
Modified subset_samples_and_variants()

wlu04:
Added a vds option to function `subset_samples_and_variants()` @jkgoodrich @klaricch

### DIFF
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -251,6 +251,8 @@ def subset_samples_and_variants(
     if is_vds:
         mt = mtds.variant_data
     else:
+        if remove_dead_alleles: 
+            raise ValueError("Removal of alleles observed in no samples is currently only implemented when the input dataset is a VariantDataset.")
         mt = mtds
     missing_ht = sample_ht.anti_join(mt.cols())
     missing_ht_count = missing_ht.count()
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -277,7 +277,7 @@ def subset_samples_and_variants(
 
     logger.info(
         "Finished subsetting samples. Kept %d out of %d samples in MT",
-        mt.count_cols(),
+        n_cols,
         full_count,
     )
     return t
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -231,7 +231,7 @@ def subset_samples_and_variants(
     table_key: str = "s",
     sparse: bool = False,
     gt_expr: str = "GT",
-    remove_dead_alleles: bool = False
+    remove_dead_alleles: bool = False,
 ) -> Union[hl.MatrixTable, hl.vds.VariantDataset]:
     """
     Subset the MatrixTable or VariantDataset to the provided list of samples and their variants.
@@ -265,7 +265,9 @@ def subset_samples_and_variants(
         )
 
     if is_vds:
-        mtds = hl.vds.filter_samples(mtds, sample_ht, keep=True, remove_dead_alleles=remove_dead_alleles)
+        mtds = hl.vds.filter_samples(
+            mtds, sample_ht, keep=True, remove_dead_alleles=remove_dead_alleles
+        )
         n_cols = mtds.variant_data.count_cols()
     else:
         mtds = mtds.semi_join_cols(sample_ht)
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -261,7 +261,7 @@ def subset_samples_and_variants(
             f"IDs that aren't in the MT: {missing_samples}\n"
         )
 
-    if isinstance(t, hl.vds.VariantDataset):
+    if is_vds:
         t = hl.vds.filter_samples(t, sample_ht, keep=True)
         mt = t.variant_data
     else:
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -264,7 +264,7 @@ def subset_samples_and_variants(
 
     if is_vds:
         t = hl.vds.filter_samples(t, sample_ht, keep=True)
-        mt = t.variant_data
+        n_cols = t.variant_data.count_cols()
     else:
         t = t.semi_join_cols(sample_ht)
         if sparse:
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -251,8 +251,10 @@ def subset_samples_and_variants(
     if is_vds:
         mt = mtds.variant_data
     else:
-        if remove_dead_alleles: 
-            raise ValueError("Removal of alleles observed in no samples is currently only implemented when the input dataset is a VariantDataset.")
+        if remove_dead_alleles:
+            raise ValueError(
+                "Removal of alleles observed in no samples is currently only implemented when the input dataset is a VariantDataset."
+            )
         mt = mtds
     missing_ht = sample_ht.anti_join(mt.cols())
     missing_ht_count = missing_ht.count()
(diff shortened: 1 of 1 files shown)


gnomad/utils/filtering.py
```diff
@@ -225,26 +225,30 @@ def add_filters_expr(
...
     table_key: str = "s",
     sparse: bool = False,
     gt_expr: str = "GT",
-) -> hl.MatrixTable:
+) -> Union[hl.MatrixTable, hl.vds.VariantDataset]:
     """
     Subset the MatrixTable to the provided list of samples and their variants.
 
-    :param mt: Input MatrixTable
+    :param t: Input MatrixTable or VariantDataset
```
jkgoodrich, line 238: I guess we should think about the best param name for this since we use `t` to typically mean it will take a MatrixTable or a Table (I know this is what I put in the other file, but I just got thinking about it while reading this). I really don't know what is best, maybe `ds`? @klaricch thoughts?

gnomad/utils/filtering.py
```diff
@@ -225,26 +225,30 @@ def add_filters_expr(
...
+    t: Union[hl.MatrixTable, hl.vds.VariantDataset],
     sample_path: str,
     header: bool = True,
     table_key: str = "s",
     sparse: bool = False,
     gt_expr: str = "GT",
-) -> hl.MatrixTable:
+) -> Union[hl.MatrixTable, hl.vds.VariantDataset]:
     """
     Subset the MatrixTable to the provided list of samples and their variants.
```
jkgoodrich, line 236:
```suggestion
    Subset the MatrixTable or VariantDataset to the provided list of samples and their variants.
```
```suggestion
    Subset the MatrixTable to the provided list of samples and their variants.
```

gnomad/utils/filtering.py
```diff
@@ -225,26 +225,30 @@ def add_filters_expr(
...
     :param header: Whether file with samples has a header. Default is True
     :param table_key: Key to sample Table. Default is "s"
     :param sparse: Whether the MatrixTable is sparse. Default is False
     :param gt_expr: Name of field in MatrixTable containing genotype expression. Default is "GT"
-    :return: MatrixTable subsetted to specified samples and their variants
+    :return: MatrixTable or VariantDataset subsetted to specified samples and their variants
     """
     sample_ht = hl.import_table(sample_path, no_header=not header, key=table_key)
     sample_count = sample_ht.count()
+    if isinstance(t, hl.vds.VariantDataset):
```
jkgoodrich, line 248:
```suggestion
    is_vds =  isinstance(t, hl.vds.VariantDataset)
    if is_vds:
```

gnomad/utils/filtering.py
```diff
@@ -257,20 +261,25 @@ def subset_samples_and_variants(
             f"IDs that aren't in the MT: {missing_samples}\n"
         )
 
-    mt = mt.semi_join_cols(sample_ht)
-    if sparse:
-        mt = mt.filter_rows(
-            hl.agg.any(mt[gt_expr].is_non_ref() | hl.is_defined(mt.END))
-        )
+    if isinstance(t, hl.vds.VariantDataset):
```
jkgoodrich, line 264:
```suggestion
    if is_vds:
```

gnomad/utils/filtering.py
```diff
@@ -257,20 +261,25 @@ def subset_samples_and_variants(
...
+        t = hl.vds.filter_samples(t, sample_ht, keep=True)
+        mt = t.variant_data
```
jkgoodrich, line 266:
```suggestion
        n_cols = t.variant_data.count_cols()
```

gnomad/utils/filtering.py
```diff
@@ -257,20 +261,25 @@ def subset_samples_and_variants(
...
     else:
-        mt = mt.filter_rows(hl.agg.any(mt[gt_expr].is_non_ref()))
+        t = t.semi_join_cols(sample_ht)
+        if sparse:
+            t = t.filter_rows(
+                hl.agg.any(t[gt_expr].is_non_ref() | hl.is_defined(t.END))
+            )
+        else:
+            t = t.filter_rows(hl.agg.any(t[gt_expr].is_non_ref()))
+        mt = t
```
jkgoodrich, line 275:
```suggestion
        n_cols = t.count_cols()
```

gnomad/utils/filtering.py
```diff
@@ -257,20 +261,25 @@ def subset_samples_and_variants(
...
 
     logger.info(
         "Finished subsetting samples. Kept %d out of %d samples in MT",
         mt.count_cols(),
```
jkgoodrich, line 279:
```suggestion
        n_cols,
```

gnomad/utils/filtering.py
```diff
@@ -257,20 +261,25 @@ def subset_samples_and_variants(
             f"IDs that aren't in the MT: {missing_samples}\n"
         )
 
-    mt = mt.semi_join_cols(sample_ht)
-    if sparse:
-        mt = mt.filter_rows(
-            hl.agg.any(mt[gt_expr].is_non_ref() | hl.is_defined(mt.END))
-        )
+    if isinstance(t, hl.vds.VariantDataset):
+        t = hl.vds.filter_samples(t, sample_ht, keep=True)
```
jkgoodrich, line 265: I wonder if we should add an option for `remove_dead_alleles` that can only be used with a VDS?

jkgoodrich (changes requested):
Thanks so much for starting to worth through these!!

gnomad/utils/filtering.py
klaricch: I think that would make sense, and set the default to True?

jkgoodrich: Actually, because of the computation to do this it might actually be better to set it to False. If we don't care then no reason to do the extra computations. I think in most cases we will not care, for release of a subset with samples in the full callset that can't be released, this will need to be set to True

jkgoodrich (commented):
I think this PR just needs the addition of a `remove_dead_alleles` that can only be used if the input is a VDS, default set to False. I think you also need to pull and merge with the master branch so that the tests no longer fail, but after that it will be good to merge

gnomad/utils/filtering.py
```diff
@@ -225,26 +225,33 @@ def add_filters_expr(
...
 
-    :param mt: Input MatrixTable
+    :param mtds: Input MatrixTable or VariantDataset
     :param sample_path: Path to a file with list of samples
     :param header: Whether file with samples has a header. Default is True
     :param table_key: Key to sample Table. Default is "s"
     :param sparse: Whether the MatrixTable is sparse. Default is False
     :param gt_expr: Name of field in MatrixTable containing genotype expression. Default is "GT"
-    :return: MatrixTable subsetted to specified samples and their variants
+    :param remove_dead_alleles: remove alleles observed in no samples for only VariantDataset. Default is False
```
jkgoodrich, line 245:
```suggestion
    :param remove_dead_alleles: Remove alleles observed in no samples. This option is currently only relevant when `mtds` is a VariantDataset. Default is False
```

gnomad/utils/filtering.py
```diff
@@ -225,26 +225,33 @@ def add_filters_expr(
...
+    :return: MatrixTable or VariantDataset subsetted to specified samples and their variants
     """
     sample_ht = hl.import_table(sample_path, no_header=not header, key=table_key)
     sample_count = sample_ht.count()
+    is_vds = isinstance(mtds, hl.vds.VariantDataset)
+    if is_vds:
+        mt = mtds.variant_data
+    else:
```
jkgoodrich, line 253:
```suggestion
    else:
        if remove_dead_alleles: 
            raise ValueError("Removal of alleles observed in no samples is currently only implemented when the input dataset is a VariantDataset.")
```

jkgoodrich (changes requested):
A couple suggestions for the new addition

jkgoodrich (approved):
LGTM!!

jkgoodrich:
Sorry, just need one more run of Black, then good to merge

PR #429, opened 2022-01-06
Use GitHub Releases for change log

### LINKED ISSUE #355: Improve changelog updates

nawatts:
Currently, we store release notes in [CHANGELOG.md](https://github.com/broadinstitute/gnomad_methods/blob/bbb4098b9f67ee35cf5709d8a848102c57048c4d/CHANGELOG.md).

However, this approach has some issues:
- Updates in multiple PRs cause merge conflicts
- Adding links to PRs requires another commit after opening a PR and getting its number

This approach of updating the change log incrementally was chosen because, given our usual long time between releases and messy commit history, it was a task to collect release notes at release time. Now though, GitHub can [automatically generate release notes](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes) from pull requests made since the previous release.

This change moves release notes from CHANGELOG.md to [GitHub Releases](https://github.com/broadinstitute/gnomad_methods/releases). With auto-generated release notes, all that is needed for release notes at PR time is a descriptive PR title. At release time, the generated release notes can be quickly tidied up and organized.

One thing we lose this way is an always up to date change log for the latest development version on the master branch. However, in practice, we don't have that with the current approach either since we don't require updates to CHANGELOG.md in PRs.

Resolves #355


jkgoodrich (commented):
Code looks good to me. Just to confirm what this means for each PR though. You state "With auto-generated release notes, all that is needed for release notes at PR time is a descriptive PR title. At release time, the generated release notes can be quickly tidied up and organized." But I'm not completely clear on what is included in the auto generated release notes. The nice thing about the CHANGELOG was being able to simplify all the commits in a few bullet points that were more descriptive than the PR title (and sometime each PR needs a few bullet points). Though as you point out, if we don't keep up with the change log then it's not very useful anyway. Just want to clarify so we can make sure anyone putting in a PR to `gnomad_methods` goes through the easiest path for good release notes. Also confirming that our current release notes will still have the info in the original  CHANGELOG.md, I wasn't sure about that since all of CHANGELOG.md is being removed in this PR, but I think it makes sense to have then still there since we haven't switched to forcing more descriptive PRs.

nawatts:
> But I'm not completely clear on what is included in the auto generated release notes. The nice thing about the CHANGELOG was being able to simplify all the commits in a few bullet points that were more descriptive than the PR title (and sometime each PR needs a few bullet points).

When I tried it, it looked like it gathered a list of all PRs merged since the last tag. However, the generated release notes can be used as a starting point... they don't have to be used as-is. Also, the release notes can be edited later.

> Though as you point out, if we don't keep up with the change log then it's not very useful anyway. Just want to clarify so we can make sure anyone putting in a PR to gnomad_methods goes through the easiest path for good release notes.

The previous process didn't seem to be working well. It was easy to forget to update the change log. And when it was updated, there were frequently conflicts during times of active development. Though this new process does have downsides (as you mentioned, the PR author doesn't provide detailed notes at PR time), it seemed like a decent compromise.

> Also confirming that our current release notes will still have the info in the original CHANGELOG.md, I wasn't sure about that since all of CHANGELOG.md is being removed in this PR, but I think it makes sense to have them still there since we haven't switched to forcing more descriptive PR titles.

I copied everything that was in CHANGELOG.md into release notes on [GitHub Releases for prior versions of gnomad](https://github.com/broadinstitute/gnomad_methods/releases). CHANGELOG.md now contains a link to that page.

jkgoodrich (approved):
Great, that all sounds good to me. Agree change log in PR vs good release note on a PR is a reasonable compromise. Thanks for putting this in Nick!

PR #431, opened 2022-01-07
Use Google Cloud Public Datasets as default source for public resources

### LINKED ISSUE #264: Update gnomAD resources to reflect free Google Cloud-hosted paths

nawatts:
Now that all files have been synced to Google Cloud Public Datasets, we can use it as the default source for resources. This will be a more pleasant experience for new users, since they won't have to deal with requester pays.

Resolves #264

### DIFF
--- a/docs/examples/vep.rst
+++ b/docs/examples/vep.rst
@@ -7,7 +7,7 @@ tied to a specific reference genome.
 
 .. code-block:: shell
 
-   hailctl dataproc start cluster-name --vep GRCh37 --packages gnomad --requester-pays-allow-buckets gnomad-public-requester-pays
+   hailctl dataproc start cluster-name --vep GRCh37 --packages gnomad
 
 .. note::
 
--- a/docs/resource_sources.rst
+++ b/docs/resource_sources.rst
@@ -5,15 +5,15 @@ gnomAD data is available through `multiple cloud providers' public datasets prog
 
 The functions in the :doc:`gnomad.resources </api_reference/resources/index>` package can be configured to load data from different sources.
 
-By default, resources are loaded from the gnomAD project's public Google Cloud Storage bucket ``gs://gnomad-public-requester-pays``. This bucket is `requester pays <https://cloud.google.com/storage/docs/requester-pays>`_, meaning that charges for data access and transfer will be billed to your Google Cloud project.
+By default, resources are loaded from Google Cloud Public Datasets.
 
-To load resources from a different source (for example, Google Public Datasets), use:
+To load resources from a different source (for example, the gnomAD project's public GCS bucket), use:
 
 .. code-block:: python
 
     from gnomad.resources.config import gnomad_public_resource_configuration, GnomadPublicResourceSource
 
-    gnomad_public_resource_configuration.source = GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS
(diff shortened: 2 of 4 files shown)


docs/examples/vep.rst
```diff
@@ -7,7 +7,7 @@ tied to a specific reference genome.
 
 .. code-block:: shell
 
-   hailctl dataproc start cluster-name --vep GRCh37 --packages gnomad --requester-pays-allow-buckets gnomad-public-requester-pays
+   hailctl dataproc start cluster-name --vep GRCh37 --packages gnomad
```
mike-w-wilson, line 10: Sorry I havent followed which tables have moved but did the VEP data move as well? If it did, I think L14 needs to be updated/removed. If not, I think we should keep the requester pays options here since the examples below reference vep_or_lookup_vep which will need it.

docs/resource_sources.rst
```diff
@@ -23,6 +23,9 @@ To see all available public sources for gnomAD resources, use:
 
     list(GnomadPublicResourceSource)
 
+.. note::
+
+   The gnomAD project's bucket (``gs://gnomad-public-requester-pays``) is `requester pays <https://cloud.google.com/storage/docs/requester-pays>`_, meaning that charges for data access and transfer will be billed to your Google Cloud project.
```
mike-w-wilson, line 28: Should we add in something like 
```suggestion
   The gnomAD project's bucket (``gs://gnomad-public-requester-pays``) is `requester pays <https://cloud.google.com/storage/docs/requester-pays>`_, meaning that charges for data access and transfer will be billed to your Google Cloud project. Clusters must be configured to read requester pays buckets during creation.
``hailctl dataproc start cluster-name --packages gnomad --requester-pays-allow-buckets gnomad-public-requester-pays``
```
since we are removing the example above.

mike-w-wilson (commented):
Just one question on the VEP data and one minor suggestion on the requester pays note in resources.

docs/examples/vep.rst
nawatts: `--requester-pays-allow-buckets gnomad-public-requester-pays` because the resources would be read from that bucket by default.

The VEP context tables were already publicly available, but because the VEP context tables are loaded through the resources system, they'll be read from whichever source is configured. With this change, Google Cloud Public Datasets will be the default source, so ``--requester-pays-allow-buckets gnomad-public-requester-pays` is no longer needed unless someone chooses to change the resource source to `gs://gnomad-public-requester-pays`.

nawatts: Ooh, I misunderstood your comment. The note about VEP data being in a requester pays bucket refers to the data downloaded by Hail's VEP init scripts, not any of our data.

mike-w-wilson: No worries, it was worded poorly. Gotcha, thanks for clarifying!

mike-w-wilson (approved):
LGTM

docs/resource_sources.rst
```diff
@@ -23,6 +23,15 @@ To see all available public sources for gnomAD resources, use:
...
+
+   Clusters must be configured to read requester pays buckets during creation. For example,
+
+   .. code-block::
```
mike-w-wilson, line 32: Thanks for fixing the formatting -- one downside of the suggestions feature. I shouldve done this locally first.

PR #434, opened 2022-01-12
Automatically publish GitHub Release with release notes

nawatts:
The last step in updating the release notes process. Currently, when a version tag (`v#.#.#`) is pushed, the tagged code is automatically published to PyPI. This extends that workflow to also automatically publish a corresponding [GitHub Release](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) with [release notes generated from pull requests](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes). This minimizes the chances of release notes being accidentally forgotten in the release process. Note that release notes can still be edited from the [Releases page](https://github.com/broadinstitute/gnomad_methods/releases) after the release is published.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,7 +47,7 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
-      - name: Create Release Notes
+      - name: Create release notes
         uses: actions/github-script@v5
         with:
           script: |
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,3 +47,11 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -47,3 +47,11 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
+      - name: Create Release Notes
```
jkgoodrich, line 50: Totally not needed, and maybe "Release Notes" is the standard, but otherwise this is more consistent with above
```suggestion
      - name: Create release notes
```

jkgoodrich (approved):
LGTM, one tiny possible style preference thing. Thank you for getting this all so streamlined!

.github/workflows/publish.yml
nawatts: Oops, copied/pasted this and didn't notice the inconsistency.

Issue #433, opened 2022-01-11
Set default resource source based on environment


nawatts:
Currently, resources will alway default to being read from Google Cloud Public Datasets.

https://github.com/broadinstitute/gnomad_methods/blob/748c5e4c1d066bc71507ad9c8017454c43ce8d09/gnomad/resources/config.py#L14-L16

Preferably, the default would be based on the environment, such that jobs running on EMR automatically default to reading resources from the Registry of Open Data on AWS and jobs running on HDInsight default to reading from Azure Open Datasets.

Maybe Hail already has something for identifying the environment (at least in clusters started with `hailctl`)?

gtiao:
I can ask Dan if this is possible

nawatts:
If nothing else, we can read the default source from an environment variable that can be set by either the user or `hailctl`. I was curious if Hail had anything to "guess" which cloud platform code was running on.

gtiao:
Agreed, it would be nice if this could be automatic

danking:
We did not have such a function, but we do have some heuristics as well as some environment variables that we set in all hailctl invocations. I've codified this as a new function in hail.utils: https://github.com/hail-is/hail/pull/11230. If you have an EMR cluster running and can list the environment variables available there, I'm happy to update the heuristics function to include that too. I'm also happy to receive feedback on the returned strings (currently: "dataproc" and "hdinsight").

nawatts:
Added the option to configure the default source with an environment variable in #435. Once hail-is/hail#11230 is merged and that API finalized, we can update `get_default_public_resource_source` to also use Hail's guess for the cloud provider.

PR #435, opened 2022-01-19
Allow setting the default source for public resources with an environment variable

nawatts:
Some users may always prefer to use the AWS or Azure copies of gnomAD resources. This allows them to set a default source in an environment variable instead of having to configure it in every pipeline.

Once hail-is/hail#11230 is merged, `get_default_public_resource_source` can be extended to use Hail's guess for the current cloud provider to set the default resource source.

Related to #433

### DIFF
--- a/gnomad/resources/config.py
+++ b/gnomad/resources/config.py
@@ -1,8 +1,12 @@
 """Configuration for loading resources."""
 
+import logging
 import os
 from enum import Enum
 from typing import Union
+
+
+logger = logging.getLogger(__name__)
 
 
 class GnomadPublicResourceSource(Enum):
@@ -29,8 +33,15 @@ def get_default_public_resource_source() -> Union[GnomadPublicResourceSource, st
         # Convert to a GnomadPublicResourceSource enum if possible
         try:
             default_source = GnomadPublicResourceSource(default_source_from_env)
+            logger.info(
+                "Using configured source for gnomAD resources: %s", default_source.value
+            )
             return default_source
         except ValueError:
+            logger.info(
+                "Using configured custom source for gnomAD resources: %s",
+                default_source_from_env,
+            )
             return default_source_from_env
 
     return GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS
--- a/gnomad/resources/config.py
+++ b/gnomad/resources/config.py
@@ -13,17 +13,19 @@ class GnomadPublicResourceSource(Enum):
     AZURE_OPEN_DATASETS = "Azure Open Datasets"
 
 
-DEFAULT_GNOMAD_PUBLIC_RESOURCE_SOURCE = (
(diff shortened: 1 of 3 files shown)


gnomad/resources/config.py
```diff
@@ -13,17 +14,28 @@ class GnomadPublicResourceSource(Enum):
     AZURE_OPEN_DATASETS = "Azure Open Datasets"
 
 
-DEFAULT_GNOMAD_PUBLIC_RESOURCE_SOURCE = (
-    GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS
-)
+def get_default_public_resource_source() -> Union[GnomadPublicResourceSource, str]:
+    """
+    Get the default source for public gnomAD resources.
```
jkgoodrich, line 19:
```suggestion
    Get the default source for public gnomAD resources.
    
    .. note::
    
        Default is pulled from the `GNOMAD_DEFAULT_PUBLIC_RESOURCE_SOURCE` environment variable if it exists. Otherwise `GOOGLE_CLOUD_PUBLIC_DATASETS` is used.
```

gnomad/resources/config.py
```diff
@@ -13,17 +14,28 @@ class GnomadPublicResourceSource(Enum):
...
+
+    :returns: Default resource source
+    """
+    default_source_from_env = os.getenv("GNOMAD_DEFAULT_PUBLIC_RESOURCE_SOURCE", None)
+    if default_source_from_env:
+        # Convert to a GnomadPublicResourceSource enum if possible
+        try:
+            default_source = GnomadPublicResourceSource(default_source_from_env)
+            return default_source
+        except ValueError:
```
jkgoodrich, line 29: Should we throw a warning that the environment variable defines a source that is not one of those in `GnomadPublicResourceSource`?

jkgoodrich (approved):
Looks good, thank you so much for adding, just a small note, and question. Is there somewhere that we should specify that `GNOMAD_DEFAULT_PUBLIC_RESOURCE_SOURCE` can now be used like in https://github.com/broadinstitute/gnomad_methods/blob/8bffc35b8101a6095276331f7a1fa9edcec4e86f/docs/resource_sources.rst?

gnomad/resources/config.py
nawatts: I debated about this. Users can provide their own source, such as `gs://my-bucket/gnomad-resources`, so this is (potentially) valid. I'll add some logging here.

nawatts:
> Is there somewhere that we should specify that GNOMAD_DEFAULT_PUBLIC_RESOURCE_SOURCE can now be used like in https://github.com/broadinstitute/gnomad_methods/blob/8bffc35b8101a6095276331f7a1fa9edcec4e86f/docs/resource_sources.rst?

Good point. I'll add some docs.

jkgoodrich (approved):
LGTM! Thanks for the doc/logger additions!

PR #436, opened 2022-01-20
Use hl.utils.guess_cloud_spark_provider to set default resources source

### LINKED ISSUE #433: Set default resource source based on environment

nawatts:
Follow on to #435. Use the `hl.utils.guess_cloud_spark_provider` function added in hail-is/hail#11230 to set the default source for resources. With this, pipelines running in Azure HDInsight should default to reading resources from Azure Open Datasets.

Resolves #433

### DIFF
--- a/docs/resource_sources.rst
+++ b/docs/resource_sources.rst
@@ -5,7 +5,10 @@ gnomAD data is available through `multiple cloud providers' public datasets prog
 
 The functions in the :doc:`gnomad.resources </api_reference/resources/index>` package can be configured to load data from different sources.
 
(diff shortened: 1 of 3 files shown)


jkgoodrich (approved):
LGTM! Thank you Nick!

PR #437, opened 2022-01-25
add checkpoint option to get_qc_mt

klaricch:


gnomad/sample_qc/pipeline.py
```diff
@@ -179,7 +180,10 @@ def get_qc_mt(
     )
 
     if ld_r2 is not None:
-        qc_mt = qc_mt.persist()
+        if checkpoint_path:
+            qc_mt = qc_mt.checkpoint(checkpoint_path)
```
jkgoodrich, line 184: Might want to add an `overwrite=True`. Also, let's add a logger that a checkpoint is happening.

gnomad/sample_qc/pipeline.py
```diff
@@ -179,7 +180,10 @@ def get_qc_mt(
...
+        else:
+            qc_mt = qc_mt.persist()
```
jkgoodrich, line 186: I think we should also add a logger to this so users know we have a persist here if there isn't a checkpoint path.

gnomad/sample_qc/pipeline.py
```diff
@@ -121,6 +121,7 @@ def get_qc_mt(
     filter_segdup: bool = True,
     filter_exome_low_coverage_regions: bool = False,
     high_conf_regions: Optional[List[str]] = None,
+    checkpoint_path: Optional[str] = None,
```
jkgoodrich, line 124: Add to params

jkgoodrich (changes requested):
Just requesting a couple loggers, and addition to docstring params. Thanks for adding!

gnomad/sample_qc/pipeline.py
```diff
@@ -147,6 +148,7 @@ def get_qc_mt(
     :param filter_segdup: Filter segmental duplication regions
     :param filter_exome_low_coverage_regions: If set, only high coverage exome regions (computed from gnomAD are kept)
     :param high_conf_regions: If given, the data will be filtered to only include variants in those regions
+    :param checkpoint_path: If given, the QC MT will be checkpointed to the specified path before running ld pruning. If not specified, persist will be used instead.
```
jkgoodrich, line 151:
```suggestion
    :param checkpoint_path: If given, the QC MT will be checkpointed to the specified path before running LD pruning. If not specified, persist will be used instead.
```

gnomad/sample_qc/pipeline.py
```diff
@@ -179,7 +181,12 @@ def get_qc_mt(
     )
 
     if ld_r2 is not None:
-        qc_mt = qc_mt.persist()
+        if checkpoint_path:
+            logger.info("Checkpointing the MT and ld pruning")
```
jkgoodrich, line 185:
```suggestion
            logger.info("Checkpointing the MT and LD pruning")
```

gnomad/sample_qc/pipeline.py
```diff
@@ -179,7 +181,12 @@ def get_qc_mt(
...
+            qc_mt = qc_mt.checkpoint(checkpoint_path, overwrite=True)
+        else:
+            qc_mt = qc_mt.persist()
+            logger.info("Persisting the MT and ld pruning")
```
jkgoodrich, line 189: Might also want to move before the actual persist
```suggestion
            logger.info("Persisting the MT and LD pruning")
```

jkgoodrich (changes requested):
Tiny changes

jkgoodrich (approved):
LGTM!

PR #439, opened 2022-01-27
Fix `vep_or_lookup_vep` to drop `vep_proc_id` if it exists

konradjk:

### DIFF
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -162,7 +162,7 @@ class MatrixTableResource(BaseResource):
 class VariantDatasetResource(BaseResource):
     """
     A Hail VariantDataset resource.
-    
+
     :param path: The VariantDataset path (typically ending in .vds)
     :param import_args: Any sources that are required for the import and need to be kept track of and/or passed to the import_func (e.g. .vcf path for an imported VCF)
     :param import_func: A function used to import the VariantDataset. `import_func` will be passed the `import_args` dictionary as kwargs.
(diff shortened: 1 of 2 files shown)


jkgoodrich (approved):
Ah, yes, this was added to the returned HT for debugging. https://github.com/hail-is/hail/blob/d2fce1913ae56c7e8dbac6d6afe641a90975c41b/hail/python/hail/docs/change_log.md#new-features-11

nawatts:
Looks like this will still break depending on the version of VEP. The VEP 85 and 95 context tables do not include `vep_proc_id`, but the VEP 101 context table does.

Also, this is throwing away information that might be wanted for debugging. It might be better to annotate the context table with missing values if the field is not present.

konradjk:
@nawatts thoughts on that change? We can do the missing thing if you prefer, but this is an easy go at it

nawatts:
I still think that, ideally, users would have the debugging information for the VEP'd variants in case they needed it. The variants annotated from the context table should not have a value for `vep_proc_id`. There's no reason to store it in the context tables, since we assume VEP ran successfully when they were generated. And it would be confusing in debugging if the output contained a mixture of `vep_proc_id`s from the VEP run that generated the context table and the VEP run that VEP'd the variants not found in the context table.

There's also still the issue of the VEP 101 context table containing a `vep_proc_id` field. That could be considered an issue with the context table and the best solution might be to rewrite it with that field dropped. However, as is, this change looks like it will fix the schema mismatch when using VEP 85 or 95 in exchange for surfacing the same issue when using VEP 101. To avoid that, if `vep_proc_id` is going to be dropped from `revep_ht`, it may also be worth dropping it from `vep_ht` as well.

gtiao:
Julia: we should accept Konrad's PR to fix this issue now, but then put in a new PR to address Nick's concerns

klaricch (approved):
lgtm!

PR #441, opened 2022-02-11
Updates public release file paths from requester pays to gcp public bucket

mike-w-wilson:
This updates release file paths from our requestor pays bucket to the GCP public bucket. The majority of the files have been removed from our requester pays bucket and so release file calls are failing.

### DIFF
--- a/gnomad/resources/grch37/gnomad.py
+++ b/gnomad/resources/grch37/gnomad.py
@@ -68,7 +68,7 @@ def _public_release_ht_path(data_type: str, version: str) -> str:
     :param version: One of the release versions of gnomAD on GRCh37
     :return: Path to release Table
     """
-    return f"gs://gnomad-public-requester-pays/release/{version}/ht/{data_type}/gnomad.{data_type}.r{version}.sites.ht"
+    return f"gs://gcp-public-data--gnomad/release/{version}/ht/{data_type}/gnomad.{data_type}.r{version}.sites.ht"
 
 
 def _public_coverage_ht_path(data_type: str, version: str) -> str:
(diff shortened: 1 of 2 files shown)


jkgoodrich:
I think this will take some more changes. 

Currently it expects the path that it is set with to include one of the gnomad_public buckets 

https://github.com/broadinstitute/gnomad_methods/blob/b95578056f8051d00dcbb749a10e585d48888f82/gnomad/resources/resource_utils.py#L503 which is one of https://github.com/broadinstitute/gnomad_methods/blob/b95578056f8051d00dcbb749a10e585d48888f82/gnomad/resources/resource_utils.py#L20

We no longer used `gs://gnomad-public` so we can probably remove that from `GNOMAD_PUBLIC_BUCKETS` . 

We will still write to `gnomad-public-requester-pays` so I'm not sure that we need to remove that. 

As of now this is how the default public resource is defined: https://github.com/broadinstitute/gnomad_methods/blob/b95578056f8051d00dcbb749a10e585d48888f82/gnomad/resources/config.py#L21

So it might actually be fine to leave the code as is and then when I want to write to the public resource I define that in my environment variable `GNOMAD_DEFAULT_PUBLIC_RESOURCE_SOURCE`

If we decide to change the paths in these files, we will need to update `GNOMAD_PUBLIC_BUCKETS` and likely the code for `GnomadPublicResource`, though I haven't thought through it completely.

Thoughts?

mike-w-wilson:
Yeah, I didn't get a chance to dig into this error Friday and had not fully understood the implications in the `GnomadPublicResource` class or the magic that Nick had created. I agree that we can leave the code as is and will just change this to removing the gnomad-public from the config list.

mike-w-wilson:
Closing this as it is not necessary.

PR #445, opened 2022-03-29
Modification to the `annotate_sex` pipeline to allow sex ploidy estimation using only variants instead of ref blocks

jkgoodrich:
During testing of our gnomAD v4 exome dataset we determined that there was better separation of x-ploidy values by using only the variant dataset to determine the mean DP of chromosome X rather than the reference blocks.

This PR is dependent on PR change to the VDS `impute_sex_chromosome_ploidy` method: https://github.com/hail-is/hail/pull/11701


gnomad/sample_qc/pipeline.py
```diff
@@ -270,23 +274,71 @@ def annotate_sex(
...
+                    use_only_variants=variants_only_y_ploidy,
+                )
+                y_ploidy_ht.select(
+                    "chrY_ploidy",
+                    "chrY_mean_dp",
+                    f"var_data_{normalization_contig}_mean_dp"
+                    if variants_only_y_ploidy
+                    else f"{normalization_contig}_mean_dp",
+                )
+                if variants_only_x_ploidy:
```
klaricch, line 334: should this be moved up and unindented?

klaricch, line 334: should this be moved up and unindented?

klaricch (changes requested):

gnomad/sample_qc/pipeline.py
```diff
@@ -270,6 +270,8 @@ def annotate_sex(
             raise NotImplementedError(
                 "The use of the parameter 'excluded_intervals' is currently not implemented for imputing sex chromosome ploidy on a VDS!"
             )
+        # ploidy_ht = hl.vds.impute_sex_chromosome_ploidy(
```
klaricch, line 273:
```suggestion
```

gnomad/sample_qc/pipeline.py
```diff
@@ -331,13 +342,16 @@ def annotate_sex(
                     if variants_only_y_ploidy
                     else f"{normalization_contig}_mean_dp",
                 )
-                if variants_only_x_ploidy:
+                ploidy_ht = ploidy_ht.annotate(**y_ploidy_ht[ploidy_ht.key])
+                # If the `variants_only_y_ploidy' is True modify the name of the normalization contig mean DP to indicate
+                # that this is the variant dataset only mean DP (this will have already been added if
+                # 'variants_only_x_ploidy' was also True.
+                if variants_only_y_ploidy:
```
klaricch, line 349: i think this renaming was already taken care of in line 342

klaricch (changes requested):

gnomad/sample_qc/pipeline.py
jkgoodrich: great point, I think that above is actually wrong because `impute_sex_ploidy` will return it as `{normalization_contig}_mean_dp` so I think above needs to be changed to only select that and then here we rename it

klaricch: ah ok, and then should there also be a renaming for if variants_only_x_ploidy was true after line 326?

jkgoodrich: Very right, I added the change

klaricch (approved):
I think after black will be good to go

PR #446, opened 2022-04-01
Add update to Black version in requirements and pre-commit

jkgoodrich:
I got a pre-commit hook error when trying to commit.

The error was: `ImportError: cannot import name ‘_unicodefun’ from ‘click’ `

[This](https://github.com/psf/black/issues/2964) indicated that I needed to update the version of Black, and after that I was able to add a commit.


klaricch (approved):
lgtm!

PR #447, opened 2022-04-20
Added a fix to the requirements-dev.txt so that GitHub Actions will work with Black update

jkgoodrich:


requirements-dev.txt
```diff
@@ -60,23 +59,21 @@ pytest==6.2.5
     # via -r requirements-dev.in
 pyyaml==6.0
     # via pre-commit
-regex==2021.11.10
```
klaricch, line 63: why does this (and typed-ast/appdirs) need to be removed?

klaricch (approved):

requirements-dev.txt
jkgoodrich: I'm not actually sure, but assuming that some other requirement needed it, but it's no longer needed, the `requirements-dev.txt` is automatically generated from the `requirements-dev.in` using `pip-compile`

jkgoodrich: @nawatts if you have time, can you confirm this is the reason?

nawatts: > assuming that some other requirement needed it, but it's no longer needed

Yes, this is the reason. `pip-compile` adds `# via <package>` comments to this file to indicate where transitive dependencies originate from. `regex` and `typed-ast` both had `# via black` comments, so it looks like the newer version doesn't require them anymore.

jkgoodrich: Thank you @nawatts!

PR #448, opened 2022-05-04
Change type of "pc_cols" param in ancestry function from hl.expr.ArrayExpression to List[int] to help track PCs that were used in RF model

klaricch:


gnomad/sample_qc/ancestry.py
```diff
@@ -230,8 +230,14 @@ def assign_population_pcs(
 
     if hail_input:
         pops_ht = hl.Table.from_pandas(pop_pc_pd, key=list(pop_pca_scores.key))
-        pops_ht.annotate_globals(
-            assign_pops_from_pc_params=hl.struct(min_assignment_prob=min_prob)
+        pops_ht = pops_ht.annotate_globals(
```
jkgoodrich, line 233: Nice catch

gnomad/sample_qc/ancestry.py
```diff
@@ -149,6 +150,7 @@ def assign_population_pcs(
 
     :param pop_pca_scores: Input Hail Table or Pandas Dataframe
     :param pc_cols: Columns storing the PCs to use
+    :param pc_names: List of integers to use for naming the selected PCs (i.e. an input of [1, 3] will result in the first two PCs of pc_cols being named PC1 and PC3)
```
jkgoodrich, line 153: I think this certainly works, but I'm wondering if there is a better way to do this.  `pop_pca_scores` has all the PCA scores, and is filtered for use with `pc_cols`. I think there is likely a way that we can use this info to give the needed PC names rather than pass in names as another parameter?

klaricch: Do you mean change the pc_cols parameter? 
assign_population_pcs is currently called with 
```
pop_ht, pops_rf_model = assign_population_pcs(
            pop_pca_scores_ht,
            pc_cols=[pop_pca_scores_ht.scores[i - 1] for i in pcs],
            ....
```
so some information is being lost by just subsetting to these PCs for the Hail Table option

jkgoodrich: Yeah, if we are already subsetting to these specific PCs, it seems like we can do it within the function. What I worry about is that we will add incorrect names due to this extra parameter, when we can modify this function so that we extract that info from the list, but I'm completely open to other thoughts

klaricch: I put in one way to modify this, let me know if it seems like a good option

gnomad/sample_qc/ancestry.py
```diff
@@ -148,7 +148,7 @@ def assign_population_pcs(
         can be used to expand this column into multiple `PC` columns.
 
     :param pop_pca_scores: Input Hail Table or Pandas Dataframe
-    :param pc_cols: Columns storing the PCs to use
+    :param pc_cols: List of which PCS to use/columns storing the PCs to use (i.e. [1,2,4,5])
```
jkgoodrich, line 151:
```suggestion
    :param pc_cols: List of which PCs to use/columns storing the PCs to use (i.e. [1,2,4,5])
```

jkgoodrich: Should also give an example if using a pandas dataframe? Also, we should note if this indexing is 0-based or 1-based (I think from the code it looks one based)

gnomad/sample_qc/ancestry.py
```diff
@@ -163,19 +163,17 @@ def assign_population_pcs(
 
     hail_input = isinstance(pop_pca_scores, hl.Table)
     if hail_input:
```
jkgoodrich, line 165: I'm thinking we might want to add a check that this is a list of integers with an error message detailing the expectation

jkgoodrich (changes requested):
I think this works and is a bit cleaner, but again I'm open to suggestions if you think it's better to pass in names as you were doing originally.

One thing to note is that this needs to have a label change of `breaking change` instead of `ignore` with the new changes

jkgoodrich:
@klaricch I think this one is back to you

jkgoodrich (approved):
LGTM

PR #451, opened 2022-05-19
Handle tags created through GitHub in publish release workflow

nawatts:
Currently, the GitHub Actions workflow for publishing a release assumes that the version is tag is created locally and pushed to GitHub. It automatically creates a corresponding GitHub Release with automatically generated release notes.

https://github.com/broadinstitute/gnomad_methods/blob/00365eb7b11ae1f93ad865f1c56917b04c19006f/CONTRIBUTING.md#publishing-a-release

https://github.com/broadinstitute/gnomad_methods/blob/00365eb7b11ae1f93ad865f1c56917b04c19006f/.github/workflows/publish.yml#L50-L57

However, it's also possible to create a tag through GitHub by manually [creating a GitHub Release](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository#creating-a-release).

In that case, the last step of the release workflow fails because a release already exists for the tag. This happened for the [0.6.2 release](https://github.com/broadinstitute/gnomad_methods/runs/6375842345?check_suite_focus=true) and caused confusion because, while everything ended in the desired state (version tagged, code published to PyPI, and release notes in a GitHub Release), the publish release workflow showed as failed.

This changes the publish release workflow to handled the case where a tag is created through GitHub by first checking if a GitHub Release already exists for the version tag before attempting to create one.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,11 +47,29 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
-      - name: Create release notes
+      - name: Create GitHub Release
         uses: actions/github-script@v5
         with:
           script: |
-            await github.request(`POST /repos/${{ github.repository }}/releases`, {
-              tag_name: "${{ github.ref }}",
-              generate_release_notes: true
-            });
+            // Get repository and tag from workflow context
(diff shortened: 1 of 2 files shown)


ch-kr (approved):
one small (optional) suggestion, but LGTM. thank you for adding this!!

PR #453, opened 2022-05-20
Fix to only add the `error_rate` annotation if `fit` is not supplied to `assign_population_pcs`

klaricch:
Only add error_rate, evaluation_sample, and training_sample annotations if "fit" not supplied to assign_population_pcs


jkgoodrich (approved):
lgtm

PR #454, opened 2022-05-25
Modify `merge_sample_qc_expr` to work with the additional VDS sample QC metrics: n_singleton_ti, n_singleton_tv, and r_ti_tv_singleton

jkgoodrich:
We added three new metrics to the VDS sample QC output. This will modify `merge_sample_qc_expr` so that it handles those new annotations appropriately.


klaricch (approved):
lgtm

PR #468, opened 2022-07-18
Add an `n_pcs` option to `run_platform_pca`

jkgoodrich:
`run_platform_pca` currently has no option to change the number of PCs to compute


mike-w-wilson (approved):
LGTM

PR #471, opened 2022-07-25
Add `bi_allelic_only` and `snv_only` options to `get_qc_mt`

jkgoodrich:
Allows for some additional QC MT options so that multi-allelic variants and indels can be included if wanted


mike-w-wilson (commented):
Two things: one is a small suggestion for the doctring and the other I cannot add in because the lines were not changed but  `bi_allelic_only` and `snv_only` should also be added to the global annotations in L212

mike-w-wilson (approved):
LGTM

PR #473, opened 2022-08-19
Add block_size option to get_qc_mt for LD pruning

klaricch:
Add option for setting the block size for LD pruning within  get_qc_mt


jkgoodrich (approved):
Just needs a change log label, but LGTM

PR #482, opened 2022-09-16
update the paths within import_args to use the gcp public bucket for grch38

klaricch:
updates the paths within import_args for certain resources to use the gcp public bucket


jkgoodrich (approved):
Thanks for fixing! LGTM for grch38. Can you modify the grch37 resource import args too?

PR #483, opened 2022-09-19
Update the paths within import_args to use the gcp public bucket for …

klaricch:
updates the paths within import_args for certain resources to use the gcp public bucket


jkgoodrich (approved):
LGTM

PR #484, opened 2022-09-22
Add `coverage_mt` option to `annotate_sex` which takes an optional precomputed coverage MT to use for ploidy imputation instead of remaking it.

jkgoodrich:
Will currently fail tests because I am waiting on a PR in Hail: https://github.com/hail-is/hail/pull/12195


mike-w-wilson (approved):
This looks good to me as is so approving but I'm guessing we'd wait to merge until the hail PR is in?

PR #485, opened 2022-09-23
add generic constraint functions: `build_models()`, `build_plateau_models_pop()`, `build_plateau_models_total()`, `build_coverage_model()`, `get_all_pop_lengths()`

averywpx:
Add generic constraint functions called in building model part of the constraint pipeline.

### DIFF
--- a/gnomad/utils/constraint.py
+++ b/gnomad/utils/constraint.py
@@ -1,7 +1,7 @@
 # noqa: D100
 
 import logging
-from typing import Union
+from typing import List, Optional, Tuple, Union
 
 import hail as hl
 
@@ -11,6 +11,8 @@ logging.basicConfig(
 )
 logger = logging.getLogger("constraint_utils")
 logger.setLevel(logging.INFO)
+
+COVERAGE_CUTOFF = 40
 
 
 def annotate_mutation_type(
@@ -125,3 +127,274 @@ def collapse_strand(
         if isinstance(t, hl.Table)
         else t.annotate_rows(**collapse_expr)
     )
+
+
+def build_models(
+    coverage_ht: hl.Table,
+    weighted: bool = False,
+    pops: Optional[Tuple[str]] = (),
+    keys: Tuple[str] = (
+        "context",
+        "ref",
+        "alt",
+        "methylation_level",
+        "mu_snp",
+    ),
+    cov_cutoff: int = COVERAGE_CUTOFF,
+) -> Tuple[Tuple[float, float], hl.Struct]:
+    """
+    Build coverage and plateau models.
+
+    Coverage model: corrects proportion of expected variation at low coverage sites. Low coverage sites are defined as sites with median coverage < `HIGH_COVERAGE_CUTOFF`.
+        The x and y of the plateau models:
+            x: `mu_snp` - mutation rate
+            y: proportion observed ('observed_variants' or 'observed_{pop}' / 'possible_variants')
+
+    This function builds plateau models to calibrate mutation rate estimates against the proportion observed
+    of each substitution, context, and methylation level in `coverage_ht` considering only high coverage sites,
+    or sites above a median coverage of `HIGH_COVERAGE_CUTOFF`. Since the output of `get_proportion_observed_by_coverage`
+    is used as `coverage_ht`, the proportion observed will be high-quality variants below 0.1% frequency at synonymous
+    sites. Two plateau models are fit, one for CpG transitions and one for the remainder of sites (transversions and non-CpG transitions).
+
+    Plateau model: adjusts proportion of expected variation based on location in the genome and CpG status.
+        The x and y of the coverage model:
+            x: log10('exome_coverage') at low coverage site
+            y: sum('observed_variants')/ (`high_coverage_scale_factor` * sum('possible_variants' * 'mu_snp') at low coverage site
+                where `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage site
+
+    For low coverage sites, or sites below `HIGH_COVERAGE_CUTOFF`, this function performs a base-level resolution rather than exon-level to compute a coverage correction factor
+    to reduce the inaccuracy of expected variant counts caused by low coverage on each base. The coverage models are built
+    by first defining a metric that is derived by dividing the number of observed variants with the total number of
+    possible variants times the mutation rate summed across all substitutions, contexts, and methylation level. Since
+    the output of `get_proportion_observed_by_coverage` is used as `coverage_ht`, the number of observed variants and possible
+    variants will be at synonymous sites. The function computes this metric for high coverage sites as a global scaling
+    factor, and divides this metric at low coverage sites by this scaling factor to create an observed:expected ratio. Then the
+    coverage model is built of log10(coverage) to this scaled ratio as a correction factor for low coverage sites.
+
+    .. note::
+        This function expects that the input `coverage_ht` is the output of `get_proportion_observed_by_coverage`, and
+        therefore the following fields should be present in `coverage_ht`:
+            - context - trinucleotide genomic context
+            - ref - the reference allele
+            - alt - the alternate base
+            - methylation_level - methylation level
+            - exome_coverage - median exome coverage at integer values between 1-100
+            - observed_variants - the number of observed variants in the dataset for each variant. Note that the term "variant" here refers to a specific substitution, context, methylation level, and coverage combination.
+            - downsampling_counts_{pop} (pop defaults to ()) - array of observed variant counts per population after downsampling
+            - mu_snp - mutation rate
+            - possible_variants - the number of possible variants in the dataset for each variant. Note that the term "variant" here refers to a specific substitution, context, methylation level, and coverage combination.
+
+    :param coverage_ht: Input coverage Table.
+    :param weighted: Whether to weight the high coverage model (a linear regression model) by 'possible_variants'. Defaults to False.
+    :param pops: List of populations used to build coverage and plateau models. Defaults to ().
+    :param keys: Keys to group observed and possible variant counts.
+    :param cov_cutoff: Median coverage cutoff. Sites with coverage above this cutoff are considered well covered and will be used to build plateau models. Sites below this cutoff have low coverage and will be used to build coverage models. Defaults to `HIGH_COVERAGE_CUTOFF`.
+    :return: Coverage model and plateau models.
+    """
+    # Filter to sites with coverage above `cov_cutoff`
+    all_high_coverage_ht = coverage_ht.filter(coverage_ht.exome_coverage >= cov_cutoff)
+    agg_expr = {
+        "observed_variants": hl.agg.sum(all_high_coverage_ht.observed_variants),
+        "possible_variants": hl.agg.sum(all_high_coverage_ht.possible_variants),
+    }
+    for pop in pops:
+        agg_expr[f"observed_{pop}"] = hl.agg.array_sum(
+            all_high_coverage_ht[f"downsampling_counts_{pop}"]
+        )
+    high_coverage_ht = all_high_coverage_ht.group_by(*keys).aggregate(**agg_expr)
+    high_coverage_ht = annotate_mutation_type(high_coverage_ht)
+
+    # Build plateau models
+    plateau_models = build_plateau_models_total(
+        high_coverage_ht,
+        cpg="cpg",
+        mu_snp="mu_snp",
+        observed_variants="observed_variants",
+        possible_variants="possible_variants",
+        weighted=weighted,
+    )
+    # Build plateau models for populations if `pops` is given
+    if pops:
+        pop_lengths = get_all_pop_lengths(high_coverage_ht, pops)
+        for length, pop in pop_lengths:
+            pop_plateau_model = build_plateau_models_pop(
+                high_coverage_ht,
+                cpg="cpg",
+                mu_snp="mu_snp",
+                observed_variants=f"observed_{pop}",
+                possible_variants="possible_variants",
+                pop=pop,
+                length=length,
+                weighted=weighted,
+            )
+            plateau_models = plateau_models.annotate(**pop_plateau_model[pop])
+
+    # Build coverage model
+    all_low_coverage_ht = coverage_ht.filter(
+        (coverage_ht.exome_coverage < cov_cutoff) & (coverage_ht.exome_coverage > 0)
+    )
+    # a metric that represents the relative mutability of the exome for high coverage sites as a global scaling factor
+    high_coverage_scale_factor = all_high_coverage_ht.aggregate(
+        hl.agg.sum(all_high_coverage_ht.observed_variants)
+        / hl.agg.sum(
+            all_high_coverage_ht.possible_variants * all_high_coverage_ht.mu_snp
+        )
+    )
+    low_coverage_ht = all_low_coverage_ht.group_by(
+        log_coverage=hl.log10(all_low_coverage_ht.exome_coverage)
+    ).aggregate(
+        low_coverage_obs_exp=hl.agg.sum(all_low_coverage_ht.observed_variants)
+        / (
+            high_coverage_scale_factor
+            * hl.agg.sum(
+                all_low_coverage_ht.possible_variants * all_low_coverage_ht.mu_snp
+            )
+        )
+    )
+    coverage_model = build_coverage_model(
+        low_coverage_ht,
+        low_coverage_obs_exp="low_coverage_obs_exp",
+        log_coverage="log_coverage",
+    )
+    # TODO: consider weighting here as well
+
+    return coverage_model, plateau_models
+
+
+def build_plateau_models_pop(
+    ht: hl.Table,
+    cpg: str,
+    mu_snp: str,
+    observed_variants: str,
+    possible_variants: str,
+    pop: str,
+    length: int,
+    weighted: bool = False,
+) -> hl.Struct:
+    """
+    Build plateau model for each `pop` population to calibrate mutation rate to compute predicted proportion observed value.
+
+    The x and y of the plateau models:
+        x: `mu_snp` - mutation rate
+        y: proportion observed (observed_{pop}' / 'possible_variants')
+
+    :param ht: High coverage Table.
+    :param cpg: The annotation name of booleans that determine whether the site is CPG site or not.
+    :param mu_snp: The annotation name of the mutation rate.
+    :param observed_variants: The annotation name of the observed variant counts for each combination of keys in `ht`.
+    :param possible_variants: The annotation name of the possible variant counts for each combination of keys in `ht`.
+    :param pop: List of populations. Defaults to None.
+    :param length: The minimum array length of variant counts array in downsampings for `pop`.
+    :param weighted: Whether to generalize the model to weighted least squares using 'possible_variants', defaults to False.
+    :return: A Dictionary of intercepts and slopes for plateau models of each population.
+    """
+    agg_expr = {
+        pop: [
+            hl.agg.group_by(
+                ht[cpg],
+                hl.agg.linreg(
+                    ht[observed_variants][i] / ht[possible_variants],
+                    [1, ht[mu_snp]],
+                    weight=ht[possible_variants] if weighted else None,
+                ).beta,
+            )
+            for i in range(length)
+        ]
+    }
+    return ht.aggregate(hl.struct(**agg_expr))
+
+
+def build_plateau_models_total(
+    ht: hl.Table,
+    cpg: str,
+    mu_snp: str,
+    observed_variants: str,
+    possible_variants: str,
+    weighted: bool = False,
+) -> hl.Struct:
+    """
+    Build plateau model for all release samples without downsamplings to calibrate mutation rate to compute predicted proportion observed value.
+
+    The x and y of the plateau models:
+        x: `mu_snp` - mutation rate
+        y: proportion observed ('observed_variants'/ 'possible_variants')
+
+    :param ht: High coverage Table.
+    :param cpg: The annotation name of booleans that determine whether the site is CPG site or not.
+    :param mu_snp: The annotation name of the mutation rate.
+    :param observed_variants: The annotation name of the observed variant counts for each combination of keys in `ht`.
+    :param possible_variants: The annotation name of the possible variant counts for each combination of keys in `ht`.
+    :param weighted: Whether to generalize the model to weighted least squares using 'possible_variants', defaults to False.
+    :return: A Dictionary of intercepts and slopes for a plateau model.
+    """
+    agg_expr = {
+        "total": hl.agg.group_by(
+            ht[cpg],
+            hl.agg.linreg(
+                ht[observed_variants] / ht[possible_variants],
+                [1, ht[mu_snp]],
+                weight=ht[possible_variants] if weighted else None,
+            ).beta,
+        )
+    }
+    return ht.aggregate(hl.struct(**agg_expr))
+
+
+def build_coverage_model(
+    ht: hl.Table,
+    low_coverage_obs_exp: str,
+    log_coverage: str,
+) -> Tuple[float, float]:
+    """
+    Build coverage model.
+
+    This function uses linear regression to build a model of log10(coverage) to this scaled ratio as a correction
+    factor for predicted proportion observed/expected variant counts at low coverage sites.
+
+    The x and y of the coverage model:
+        x: log10('exome_coverage') at low coverage site
+        y: sum('observed_variants')/ (`high_coverage_scale_factor` * sum('possible_variants' * 'mu_snp') at low coverage site
+            where `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage site
+
+    :param ht: Low coverage Table.
+    :param low_cov_oe_expr: The annotation name of observed:expected ratio for a given coverage level.
+    :param log_cov_expr: The annotation name of log10 coverage.
+    :return: Tuple with intercept and slope of the model.
+    """
+    return tuple(
+        ht.aggregate(
+            hl.agg.linreg(ht[low_coverage_obs_exp], [1, ht[log_coverage]])
+        ).beta
+    )
+
+
+def get_all_pop_lengths(
+    ht, pops: Tuple[str], prefix: str = "observed_"
+) -> List[Tuple[str, str]]:
+    """
+    Get the minimum array length of variant counts array in downsampings for specific per population annotations in `ht`.
+
+    The minimum array length will be used to determine the number of plateau models built for each population.
+
+    The annotations are specified by the combination of `prefix` and each population in `pops`.
+
+    :param ht: Input Table used to build population plateau models.
+    :param pops: List of populations.
+    :param prefix: Prefix of population variant count. Defaults to 'observed_'.
+    :return: A Dictionary with the minimum array length for each population.
+    """
+    pop_downsampling_lengths = ht.aggregate(
+        [hl.agg.min(hl.len(ht[f"{prefix}{pop}"])) for pop in pops]
+    )
+
+    pop_lengths = list(zip(pop_downsampling_lengths, pops))
+    logger.info(f"Found: {pop_lengths}")
+
+    # check all the arrays of variant counts within a population have the same length
+    assert ht.all(
+        hl.all(
+            lambda f: f,
+            [hl.len(ht[f"{prefix}{pop}"]) == length for length, pop in pop_lengths],
+        )
+    )
+    return pop_lengths


gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+    """
+    pop_downsampling_lengths = ht.aggregate(
+        [hl.agg.min(hl.len(ht[f"{prefix}{pop}"])) for pop in pops]
+    )
+
+    pop_lengths = list(zip(pop_downsampling_lengths, pops))
+    logger.info("Found: %s", "".join(map(str, pop_lengths)))
+
+    # check all the arrays of variant counts within a population have the same length
+    assert ht.all(
```
averywpx, line 394: I found no circumstance the `skip_assertion` is False, so I delete it. Does the downsampling process have such check as well? If so, we can simplify it.

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+    """
+    Build coverage model.
+
+    This function uses linear regression to build a model of log10(coverage) to this scaled ratio as a correction
+    factor for predicted proportion observed/expected variant counts at low coverage sites.
+
+    The x and y of the coverage model:
+    - x: log10('exome_coverage') at low coverage site
+    - y: sum('observed_variants')/ (`high_coverage_scale_factor` * sum('possible_variants' * 'mu_snp') at low coverage site
+    where `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage site
```
ch-kr, line 357:
```suggestion
    The x and y of the coverage model:
        - x: log10 groupings of exome coverage at low coverage sites
        - y: sum('observed_variants')/ (`high_coverage_scale_factor` * sum('possible_variants' * 'mu_snp') at low coverage sites
        
        `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage sites
```
(spacing it out to make it easier to read)

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+                weight=ht[possible_variants] if weighted else None,
+            ).beta,
+        )
+    }
+    return ht.aggregate(hl.struct(**agg_expr))
+
+
+def build_coverage_model(
+    ht: hl.Table,
+    low_coverage_obs_exp: str,
```
ch-kr, line 345: should this be `low_cov_oe_expr` (or should the docstring be changed to `low_coverage_obs_exp`?

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+    :param ht: Low coverage Table.
+    :param low_cov_oe_expr: The annotation name of observed:expected ratio for a given coverage level.
+    :param log_cov_expr: The annotation name of log10 coverage.
+    :return: Tuple with intercept and slope of the model.
+    """
+    return tuple(
+        ht.aggregate(
+            hl.agg.linreg(ht[low_coverage_obs_exp], [1, ht[log_coverage]])
+        ).beta
+    )
```
ch-kr, line 368: maybe this function doesn't actually need to accept a Table and could instead accept expressions, something like:
```suggestion
    return tuple(
        ht.aggregate(
            hl.agg.linreg(low_cov_oe_expr, [1, log_cov_expr)
        ).beta
    )
```

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+        "total": hl.agg.group_by(
+            ht[cpg],
+            hl.agg.linreg(
+                ht[observed_variants] / ht[possible_variants],
+                [1, ht[mu_snp]],
+                weight=ht[possible_variants] if weighted else None,
+            ).beta,
+        )
+    }
+    return ht.aggregate(hl.struct(**agg_expr))
```
ch-kr, line 340: I feel it would be cleaner to combine the two plateau model functions into something like: 
```suggestion
    agg_expr = {
        "total": hl.agg.group_by(
            ht[cpg],
            hl.agg.linreg(
                ht[observed_variants] / ht[possible_variants],
                [1, ht[mu_snp]],
                weight=ht[possible_variants] if weighted else None,
            ).beta,
        )
    }
    if pops:
        for pop in pops:
            agg_expr[pop]: [
                hl.agg.group_by(
                        ht[cpg],
                        hl.agg.linreg(
                            ht[observed_variants][i] / ht[possible_variants],
                            [1, ht[mu_snp]],
                            weight=ht[possible_variants] if weighted else None,
                        ).beta,
                    )
                    for i in range(length)
                ]
    return ht.aggregate(hl.struct(**agg_expr))
```

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+) -> hl.Struct:
+    """
+    Build plateau model for all release samples without downsamplings to calibrate mutation rate to compute predicted proportion observed value.
+
+    The x and y of the plateau models:
+    - x: `mu_snp` - mutation rate
+    - y: proportion observed ('observed_variants'/ 'possible_variants')
+
+    :param ht: High coverage Table.
+    :param cpg: The annotation name of booleans that determine whether the site is CPG site or not.
```
ch-kr, line 323:
```suggestion
    :param cpg: The name of the Boolean noting whether a site is a CPG site.
```

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+    :param mu_snp: The annotation name of the mutation rate.
```
ch-kr, line 324:
```suggestion
    :param mu_snp: The name of the mutation rate annotation.
```

gnomad/utils/constraint.py
ch-kr: also, if this is the annotation name, should this be `cpg_str` or something similar?

gnomad/utils/constraint.py
ch-kr: and same comment about variable name for this and the two below

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+        low_coverage_ht,
+        low_coverage_obs_exp="low_coverage_obs_exp",
+        log_coverage="log_coverage",
+    )
+    # TODO: consider weighting here as well
+
+    return coverage_model, plateau_models
+
+
+def build_plateau_models_pop(
```
ch-kr, line 264: it seems redundant to have two plateau models functions

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+                all_low_coverage_ht.possible_variants * all_low_coverage_ht.mu_snp
+            )
+        )
+    )
+    coverage_model = build_coverage_model(
+        low_coverage_ht,
+        low_coverage_obs_exp="low_coverage_obs_exp",
+        log_coverage="log_coverage",
+    )
+    # TODO: consider weighting here as well
```
ch-kr, line 259:
```suggestion
    # TODO: consider weighting here as well
    coverage_model = build_coverage_model(
        low_coverage_ht,
        low_coverage_obs_exp="low_coverage_obs_exp",
        log_coverage="log_coverage",
    )
```
moving the comment to make it clearer

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+def build_plateau_models_pop(
+    ht: hl.Table,
+    cpg: str,
+    mu_snp: str,
+    observed_variants: str,
+    possible_variants: str,
+    pop: str,
+    length: int,
+    weighted: bool = False,
+) -> hl.Struct:
```
ch-kr, line 273:
```suggestion
) -> hl.StructExpression:
```

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+    pops: Optional[Tuple[str]] = (),
+    keys: Tuple[str] = (
+        "context",
+        "ref",
+        "alt",
+        "methylation_level",
+        "mu_snp",
+    ),
+    cov_cutoff: int = COVERAGE_CUTOFF,
+) -> Tuple[Tuple[float, float], hl.Struct]:
```
ch-kr, line 144:
```suggestion
) -> Tuple[Tuple[float, float], hl.StructExpression]:
```

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+
+
+def build_plateau_models_total(
+    ht: hl.Table,
+    cpg: str,
+    mu_snp: str,
+    observed_variants: str,
+    possible_variants: str,
+    weighted: bool = False,
+) -> hl.Struct:
```
ch-kr, line 314:
```suggestion
) -> hl.StructExpression:
```

gnomad/utils/constraint.py
ch-kr: could also change this function to return the agg_expr and do the actual HT aggregation elsewhere (so that you pass in HailExpressions into this function rather than the full Table)

ch-kr (changes requested):
thanks for moving these into gnomad methods! I have some edits and suggestions

gnomad/utils/constraint.py
```diff
@@ -125,3 +127,274 @@ def collapse_strand(
...
+    """
+    return tuple(
+        ht.aggregate(
+            hl.agg.linreg(ht[low_coverage_obs_exp], [1, ht[log_coverage]])
+        ).beta
+    )
+
+
+def get_all_pop_lengths(
+    ht, pops: Tuple[str], prefix: str = "observed_"
```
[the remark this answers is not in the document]
averywpx: `,`is not necessary

gnomad/utils/constraint.py
```diff
@@ -125,3 +135,246 @@ def collapse_strand(
...
+            )
+        )
+    )
+    # Build the coverage model
+    # TODO: consider weighting here as well
+    coverage_model_expr = build_coverage_model(
+        low_coverage_obs_exp=low_coverage_ht.low_coverage_obs_exp,
+        log_coverage=low_coverage_ht.log_coverage,
+    )
+    coverage_model = tuple(low_coverage_ht.aggregate(coverage_model_expr).beta)
```
averywpx, line 269: Tbh I don't really like the idea to take the `ht` param out of the function. The function is not a complete function without this line of code. If the users only use the function without this line of code, they get nothing.

gnomad/utils/constraint.py
```diff
@@ -125,3 +135,246 @@ def collapse_strand(
...
+    :return: A Dictionary with the minimum array length for each population.
+    """
+    pop_downsampling_lengths = ht.aggregate(
+        [hl.agg.min(hl.len(ht[f"{prefix}{pop}"])) for pop in pops]
+    )
+
+    pop_lengths = list(zip(pop_downsampling_lengths, pops))
+    logger.info("Found: %s", "".join(map(str, pop_lengths)))
+
+    for length, pop in pop_lengths:
```
averywpx, line 375: I reformat the code so that I can raise errors. Just double check the logic is fine.

gnomad/utils/constraint.py
```diff
@@ -125,3 +135,246 @@ def collapse_strand(
...
+    keys: Tuple[str] = (
+        "context",
+        "ref",
+        "alt",
+        "methylation_level",
+        "mu_snp",
+    ),
+    cov_cutoff: int = COVERAGE_CUTOFF,
+) -> Tuple[Tuple[float, float], hl.StructExpression]:
+    """
```
averywpx, line 153: I reformat the docstring a little bit. I didn't delete some sentence because it make more sense to keep it under this format

gnomad/utils/constraint.py
ch-kr: this seems fine, but I tested it, and it's doing more work; slight preference for reverting and adding a generic error message
image

gnomad/utils/constraint.py
```diff
@@ -251,150 +261,123 @@ def build_models(
...
+def build_plateau_models(
     ht: hl.Table,
-    cpg: str,
-    mu_snp: str,
-    observed_variants: str,
-    possible_variants: str,
+    cpg: hl.BooleanExpression,
+    mu_snp: hl.Float64Expression,
+    observed_variants: hl.Int64Expression,
+    possible_variants: hl.Int64Expression,
```
ch-kr, line 279: in general, nice to add `_expr` to these variable names to make it extra clear that they are expressions

gnomad/utils/constraint.py
```diff
@@ -132,7 +141,7 @@ def collapse_strand(
 def build_models(
```
ch-kr, line 141: sorry for the late comment on this also, but after re-reading this a few times, I feel like this function belongs more in a pipeline script (something like `gnomad/constraint/pipeline.py` possibly, similar to how we have some more pipeline sample QC functions within https://github.com/broadinstitute/gnomad_methods/blob/main/gnomad/sample_qc/pipeline.py)

gnomad/utils/constraint.py
```diff
@@ -251,150 +261,123 @@ def build_models(
...
-    possible_variants: str,
+    cpg: hl.BooleanExpression,
+    mu_snp: hl.Float64Expression,
+    observed_variants: hl.Int64Expression,
+    possible_variants: hl.Int64Expression,
+    prefix: str = "observed_",
     weighted: bool = False,
-) -> hl.Struct:
+    pops: Tuple[str] = (),
+) -> dict:
```
ch-kr, line 283:
```suggestion
) -> Dict[str, Dict[bool, hl.expr.ArrayExpression]]
```

gnomad/utils/constraint.py
```diff
@@ -1,7 +1,7 @@
 # noqa: D100
 
 import logging
-from typing import List, Optional, Tuple, Union
+from typing import List, Tuple, Union
```
ch-kr, line 4:
```suggestion
from typing import Dict, List, Tuple, Union
```

gnomad/utils/constraint.py
```diff
@@ -141,54 +150,66 @@ def build_models(
         "mu_snp",
     ),
     cov_cutoff: int = COVERAGE_CUTOFF,
-) -> Tuple[Tuple[float, float], hl.Struct]:
+) -> Tuple[Tuple[float, float], hl.StructExpression]:
```
ch-kr, line 153:
```suggestion
) -> Tuple[Tuple[float, float], hl.expr.StructExpression]:
```

gnomad/utils/constraint.py
ch-kr, line 279:
```suggestion
    cpg: hl.expr.BooleanExpression,
    mu_snp: hl.expr.Float64Expression,
    observed_variants: hl.expr.Int64Expression,
    possible_variants: hl.expr.Int64Expression,
```

gnomad/utils/constraint.py
```diff
@@ -251,150 +261,123 @@ def build_models(
...
 
 
 def build_coverage_model(
-    ht: hl.Table,
-    low_coverage_obs_exp: str,
-    log_coverage: str,
-) -> Tuple[float, float]:
+    low_coverage_obs_exp: hl.Float64Expression,
+    log_coverage: hl.Float64Expression,
+) -> hl.StructExpression:
```
ch-kr, line 334:
```suggestion
    low_coverage_obs_exp: hl.expr.Float64Expression,
    log_coverage: hl.expr.Float64Expression,
) -> hl.expr.StructExpression:
```

ch-kr (changes requested):
thanks for your hard work! I have more suggestions

gnomad/utils/constraint.py
averywpx: Maybe we can move the `build_models()` to `gnomad-constraint/gnomad_constraint/utils/constraint.py`? I can definitely create a new file at `gnomad/constraint/pipeline.py`. My concern is that I can't think about other functions to put into this file, and the purpose of this file is similar to the `constraint.py`.

ch-kr: open to suggestions here. what do you think @jkgoodrich?

gnomad/utils/constraint.py
```diff
@@ -125,3 +136,248 @@ def collapse_strand(
...
+    )
+    plateau_models = high_coverage_ht.aggregate(hl.struct(**plateau_models_agg_expr))
+
+    # Filter to sites with coverage below `cov_cutoff and larger than 0
+    all_low_coverage_ht = coverage_ht.filter(
+        (coverage_ht.exome_coverage < cov_cutoff) & (coverage_ht.exome_coverage > 0)
+    )
+    # Metric that represents the relative mutability of the exome calculated on high coverage sites
+    # This metric is used as scaling factor when building the coverage model
+    high_coverage_scale_factor = all_high_coverage_ht.aggregate(
```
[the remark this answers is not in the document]
averywpx: My understanding is `all_high_coverage_ht` has been filtered to high coverage at line [216](https://github.com/broadinstitute/gnomad_methods/blob/417ff02fcc1c95b8be2f7780d55c9cb67d2e262d/gnomad/utils/constraint.py#L216). Please correct me if I misunderstood it!

gnomad/utils/constraint.py
averywpx: Sorry I didn't say I idea clearly. I felt if the users call `build_coverage_model()` they won't get a complete coverage model, because the coverage model haven't been aggregated on this line. I just worry that the users may be confused on the uncompleted model.

gnomad/utils/constraint.py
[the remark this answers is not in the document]
ch-kr: nope that's right, thanks!

gnomad/utils/constraint.py
```diff
@@ -125,3 +136,248 @@ def collapse_strand(
...
+    agg_expr = {
+        "observed_variants": hl.agg.sum(all_high_coverage_ht.observed_variants),
+        "possible_variants": hl.agg.sum(all_high_coverage_ht.possible_variants),
+    }
+    for pop in pops:
+        agg_expr[f"observed_{pop}"] = hl.agg.array_sum(
+            all_high_coverage_ht[f"downsampling_counts_{pop}"]
+        )
+    # Generate a Table with all necessary annotations (x and y listed above) for the plateau models
+    high_coverage_ht = all_high_coverage_ht.group_by(*keys).aggregate(**agg_expr)
```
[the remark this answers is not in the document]
ch-kr: ^this comment is not relevant since `all_high_coverage_ht` does get used again below

gnomad/utils/constraint.py
ch-kr: oh, I see! I don't think this is an issue given that many of our functions are similar conceptually (they return HailExpressions that aren't meaningful until actually evaluated elsewhere)

gnomad/utils/constraint.py
```diff
@@ -125,3 +136,251 @@ def collapse_strand(
...
+        where `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage site
+
+    :param low_coverage_obs_exp_expr: The Float64Expression of observed:expected ratio for a given coverage level.
+    :param log_coverage_expr: The Float64Expression of log10 coverage.
+    :return: Tuple with intercept and slope of the model.
+    """
+    return hl.agg.linreg(low_coverage_obs_exp_expr, [1, log_coverage_expr])
+
+
+def get_all_pop_lengths(
```
averywpx, line 353: This function wasn't used in the scope of PR but it will be used when finalizing dataset, so I didn't delete the function.
One of the situation it may be used is that we may want to use this function to check the length of observed variant counts array for each population in `build_models()`.

gnomad/utils/constraint.py
```diff
@@ -125,3 +136,251 @@ def collapse_strand(
...
+    for pop in pops:
+        agg_expr[f"observed_{pop}"] = hl.agg.array_sum(
+            all_high_coverage_ht[f"downsampling_counts_{pop}"]
+        )
+
+    # Generate a Table with all necessary annotations (x and y listed above) for the plateau models
+    high_coverage_ht = all_high_coverage_ht.group_by(*keys).aggregate(**agg_expr)
+    high_coverage_ht = annotate_mutation_type(high_coverage_ht)
+
+    # Build plateau models
```
averywpx, line 232: Do we want to check the length of observed variant counts array for each population here? The check can't be placed check to `build_plateau_models()` because it needs `ht` for aggregation. The only concern is calling `get_all_pop_lengths()` may be redundant, because the dictionary it returns won't be used. Please let me know if you have any ideas to simplify the code!

gnomad/utils/constraint.py
```diff
@@ -125,3 +136,251 @@ def collapse_strand(
...
+
+
+def build_plateau_models(
+    cpg_expr: hl.expr.BooleanExpression,
+    mu_snp_expr: hl.expr.Float64Expression,
+    observed_variants_expr: hl.expr.Int64Expression,
+    possible_variants_expr: hl.expr.Int64Expression,
+    pop_observed_variants_exprs: List[hl.ArrayNumericExpression] = [],
+    weighted: bool = False,
+) -> Dict[str, Dict[bool, hl.expr.ArrayExpression]]:
```
averywpx, line 289: I found that the return type could be either `Dict[bool, hl.expr.ArrayExpression]` or `<ArrayExpression of type array<dict<bool, array<float64>>>>` when `pop_observed_variants_exprs` is specified. I explained it at the return type of the docstring. Should we specify this here as well?

gnomad/utils/constraint.py
```diff
@@ -360,3 +370,254 @@ def collapse_strand(
...
+            hl.agg.linreg(
+                observed_variants_expr / possible_variants_expr,
+                [1, mu_snp_expr],
+                weight=possible_variants_expr if weighted else None,
+            ).beta,
+        )
+    }
+    # Build plateau models using sites in population downsamplings if population is specified
+    for pop, pop_observed_variants_expr in pop_observed_variants_exprs.items():
+        plateau_models_agg_expr[pop] = hl.agg.array_agg(
```
averywpx, line 552: Konrad [suggested](https://the-tgg.slack.com/archives/C03PP8Y98LA/p1665262483906149?thread_ts=1665177762.851659&cid=C03PP8Y98LA) that we can used `array_agg()` so I add it here. I tested it and it works as expected!!! Just want to double check if it can be simplified.

gnomad/utils/constraint.py
jkgoodrich: Hmm, I see what you mean @ch-kr. I think I'm OK with it being here because to me it makes sense to have it in `gnomad_methods` and I think if there are not other files in `gnomad/constraint/` other than `constraint.py` and `pipeline.py` there isn't really a need for the `pipeline.py` since I think of those as pulling together the more generalized functions in multiple `sample_qc` or `variant_qc` modules. However, I am also fine with whatever the two of you think is best.

ch-kr (commented):
thanks for the hard work on this! a few more comments, but this is close

gnomad/utils/constraint.py
ch-kr: works for me, let's keep this function in this script then. we can always revisit in the future if needed

gnomad/utils/constraint.py
averywpx: I tried to use `map_value()` and got this error: `dynamic variables created by 'hl.bind' or lambda methods like 'hl.map' may not be aggregated`. I guess it's saying the `array_agg` can't aggregate the dict values given by lambda in `map_value()`, so I kept the old version.

gnomad/utils/constraint.py
```diff
@@ -23,12 +30,13 @@ def annotate_with_mu(
 
     .. note::
 
-        `ht` is expected to include annotations that `mutation_ht` is keyed by, but these annotations don't need to be
-        the keys of `ht`.
+        `ht` is expected to include annotations that `mutation_ht` is keyed by, but
```
averywpx, line 33: I wrapped the docstrings for all the functions in this file. Please ignore the functions that are out of the scope of this PR. Their contents weren't changed at all!

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,275 @@ def collapse_strand(
...
+
+    The minimum array length will be used to determine the number of plateau models
+    built for each population.
+
+    The annotations are specified by the combination of `prefix` and each population in
+    `pops`.
+
+    :param ht: Input Table used to build population plateau models.
+    :param pops: Populations used to categorize observed variant counts in downsampings.
+    :param prefix: Prefix of population observed variant counts. Default is "observed".
```
averywpx, line 664: Should we just delete the `observed` since it failed the test?

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+            hl.agg.sum(high_cov_ht.observed_variants)
+            / hl.agg.sum(high_cov_ht.possible_variants * high_cov_ht.mu_snp)
+        )
+
+        # Generate a Table with all necessary annotations (x and y listed above)
+        # for the coverage model.
+        low_cov_group_ht = low_cov_ht.group_by(
+            log_coverage=hl.log10(low_cov_ht.exome_coverage)
+        ).aggregate(
+            low_coverage_obs_exp=hl.agg.sum(low_cov_ht.observed_variants)
```
ch-kr, line 554:
```suggestion
            low_coverage_oe=hl.agg.sum(low_cov_ht.observed_variants)
```

ch-kr: (for consistency, since `build_coverage_model` now uses `oe` rather than `obs_exp`)

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+            / (
+                high_coverage_scale_factor
+                * hl.agg.sum(low_cov_ht.possible_variants * low_cov_ht.mu_snp)
+            )
+        )
+
+        # Build the coverage model.
+        # TODO: consider weighting here as well
+        coverage_model_expr = build_coverage_model(
+            low_coverage_oe_expr=low_cov_group_ht.low_coverage_obs_exp,
```
ch-kr, line 564:
```suggestion
            low_coverage_oe_expr=low_cov_group_ht.low_coverage_oe,
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+
+def get_all_pop_lengths(
+    ht,
+    pops: Tuple[str],
+    prefix: str = "observed_",
+) -> List[Tuple[str, str]]:
+    """
+    Get the minimum length of observed variant counts array for each population downsamping.
+
+    The minimum array length will be used to determine the number of plateau models
```
ch-kr, line 660: this isn't true anymore, right? since this function doesn't get called when building the plateau models

gnomad/utils/constraint.py
ch-kr: let's keep it; looks like this has been resolved

gnomad/utils/constraint.py
ch-kr: thanks for the heads up! as a quick poinrt of feedback: these docstring changes made the PR a lot harder to read through

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+        `coverage_ht`:
+        - context - trinucleotide genomic context
+        - ref - the reference allele
+        - alt - the alternate base
+        - methylation_level - methylation level
+        - cpg - whether the site is CpG site
+        - exome_coverage - median exome coverage at integer values between 1-100
+        - observed_variants - the number of observed variants in the dataset for each
+        variant. Note that the term "variant" here refers to a specific substitution,
+        context, methylation level, and coverage combination.
```
ch-kr, line 484:
```suggestion
        context, methylation level, and coverage combination
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+        - downsampling_counts_{pop} (optional) - array of observed variant counts per
+        population after downsampling. Used only when `pops` is specified.
+        - mu_snp - mutation rate
+        - possible_variants - the number of possible variants in the dataset for each
+        variant.
```
ch-kr, line 489:
```suggestion
        variant
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+
+    :param coverage_ht: Input coverage Table.
+    :param weighted: Whether to weight the high coverage model (a linear regression
```
ch-kr, line 492:
```suggestion
    :param weighted: Whether to weight the plateau models (a linear regression
```
this doesn't actually weight the coverage models at the moment

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+      model) by 'possible_variants'. Default is False.
+    :param pops: List of populations used to build coverage and plateau models.
```
ch-kr, line 494:
```suggestion
    :param pops: List of populations used to build plateau models.
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+      Default is ().
+    :param keys: Keys to group observed and possible variant counts.
```
ch-kr, line 496:
```suggestion
    :param keys: Fields used group observed and possible variant counts.
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+      Default is ("context","ref", "alt", "methylation_level", "mu_snp").
+    :param cov_cutoff: Median coverage cutoff. Sites with coverage above this cutoff
+      are considered well covered and will be used to build plateau models. Sites below
+      this cutoff have low coverage and will be used to build coverage models. Defaults
+      to `COVERAGE_CUTOFF`.
+    :param coverage_model: Whether to build a coverage model. Default is False.
```
ch-kr, line 502:
```suggestion
    :param coverage_model: Whether to build coverage model. Default is False.
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+    :return: Coverage model and plateau models.
```
ch-kr, line 503:
```suggestion
    :return: Coverage model (if specified, otherwise None) and plateau models.
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+def build_plateau_models(
+    cpg_expr: hl.expr.BooleanExpression,
+    mu_snp_expr: hl.expr.Float64Expression,
+    observed_variants_expr: hl.expr.Int64Expression,
+    possible_variants_expr: hl.expr.Int64Expression,
+    pop_observed_variants_exprs: Dict[str, hl.ArrayNumericExpression] = {},
+    weighted: bool = False,
+) -> Dict[str, Union[Dict[bool, hl.expr.ArrayExpression], hl.ArrayExpression]]:
+    """
+    Build plateau models for all the sites in `ht` and sites in each `pop` population downsampling to calibrate mutation rate to compute predicted proportion observed value.
```
ch-kr, line 580:
```suggestion
    Build plateau models to calibrate mutation rate to compute predicted proportion observed value.
```
the original description is a bit long, and this function doesn't always compute models per pop

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+    The x and y of the plateau models:
+    - x: `mu_snp` - mutation rate
+    - y: proportion observed ('observed_variants' or 'observed_{pop}' / 'possible_variants')
+
+    :param cpg_expr: The BooleanExpression noting whether a site is a CPG site.
+    :param mu_snp_expr: The Float64Expression of the mutation rate.
+    :param observed_variants_expr: The Int64Expression of the observed variant counts
+      for each combination of keys in `ht`.
+    :param possible_variants_expr: The Int64Expression of the possible variant counts
+      for each combination of keys in `ht`.
```
ch-kr, line 591:
```suggestion
    :param cpg_expr: BooleanExpression noting whether a site is a CPG site.
    :param mu_snp_expr: Float64Expression of the mutation rate.
    :param observed_variants_expr: Int64Expression of the observed variant counts
      for each combination of keys in `ht`.
    :param possible_variants_expr: Int64Expression of the possible variant counts
      for each combination of keys in `ht`.
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,279 @@ def collapse_strand(
...
+    :param pop_observed_variants_exprs: A dictionary where its key is a population name
+      and its value is a ArrayNumericExpression of observed variant counts for
```
ch-kr, line 593:
```suggestion
    :param pop_observed_variants_exprs: Dictionary with population names (keys)
      and observed variant counts ArrayNumericExpressions (values) for
```

gnomad/utils/constraint.py
ch-kr: ah, works for me. I don't love the for loop + hail expression combination though, maybe you could post to zulip to see if the hail team has any suggestions for how to streamline this?

ch-kr (commented):
some more suggestions! thanks for updating `build_models`

gnomad/utils/constraint.py
averywpx: So sorry for the inconvenience! Maybe I should separate it to another PR next time.

gnomad/utils/constraint.py
```diff
@@ -360,3 +414,280 @@ def collapse_strand(
...
+        agg_expr[f"observed_{pop}"] = hl.agg.array_sum(
+            high_cov_ht[f"downsampling_counts_{pop}"]
+        )
+
+    # Generate a Table with all necessary annotations (x and y listed above)
+    # for the plateau models.
+    high_cov_group_ht = high_cov_ht.group_by(*keys).aggregate(**agg_expr)
+    high_cov_group_ht = annotate_mutation_type(high_cov_group_ht)
+
+    # Build plateau models.
```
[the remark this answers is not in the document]
averywpx: It's called in `finalize_dataset()` part when calculating metrics for LoF variants (see [here](https://github.com/broadinstitute/gnomad_lof/blob/e09d9fdc13d43c8ac996c93735ed206ba41ed473/constraint_utils/constraint_basics.py#L826)). Maybe we can call this function but not assigning any value to its result? It still looks weird though...

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,274 @@ def collapse_strand(
...
+                cpg_expr,
+                hl.agg.linreg(
+                    pop_observed_variants / possible_variants_expr,
+                    [1, mu_snp_expr],
+                    weight=possible_variants_expr if weighted else None,
+                ).beta,
+            ),
+            pop_observed_variants_expr,
+        )
+        for pop, pop_observed_variants_expr in pop_observed_variants_exprs.items()
```
averywpx, line 623: I ended up using dictionary comprehension. There is a hail discussion [here](https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/Iterate.20through.20DictExpression).

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+        high_cov_group_ht.aggregate(hl.struct(**plateau_models_agg_expr))
+    )
+
+    if pops:
+        # Map the models to their corresponding populations.
+        pop_models_arr = plateau_models["pop"]
+        for idx, pop in enumerate(pops):
+            plateau_models[pop] = pop_models_arr[idx]
+        plateau_models.pop("pop")
+    plateau_models = hl.struct(**plateau_models)
```
ch-kr, line 540: not that much less redundant, but what about:
```
_plateau_models = dict(
    high_cov_group_ht.aggregate(hl.struct(**plateau_models_agg_expr))
)

plateau_models = {
    pop: hl.literal(pop_models_expr[idx]) for idx, pop in enumerate(pops),
}
plateau_models["total"] =  _plateau_models["total"],
plateau_models = hl.struct(**plateau_models)
```
I've also renamed `pop_models_arr` to `pop_models_expr` to make it clearer

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+    coverage_model = tuple(low_cov_group_ht.aggregate(coverage_model_expr).beta)
+    return coverage_model, plateau_models
+
+
+def build_plateau_models(
+    cpg_expr: hl.expr.BooleanExpression,
+    mu_snp_expr: hl.expr.Float64Expression,
+    observed_variants_expr: hl.expr.Int64Expression,
+    possible_variants_expr: hl.expr.Int64Expression,
+    pop_observed_variants_exprs_arr: Dict[str, hl.ArrayNumericExpression] = None,
```
ch-kr, line 582:
```suggestion
    pop_observed_variants_array_expr: hl.expr.ArrayExpression = None,
```
updated name and type

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+    - y: proportion observed ('observed_variants' or 'observed_{pop}' / 'possible_variants')
+
+    :param cpg_expr: BooleanExpression noting whether a site is a CPG site.
+    :param mu_snp_expr: Float64Expression of the mutation rate.
+    :param observed_variants_expr: Int64Expression of the observed variant counts
+      for each combination of keys in `ht`.
+    :param possible_variants_expr: Int64Expression of the possible variant counts
+      for each combination of keys in `ht`.
+    :param pop_observed_variants_exprs_arr: ArrayExpression includes all observed
+      variant counts ArrayNumericExpressions for specified populations. Default is None.
```
ch-kr, line 599:
```suggestion
    :param pop_observed_variants_exprs_arr: Nested ArrayExpression with all observed
      variant counts ArrayNumericExpressions for specified populations. Default is None.
```
slightly clearer. maybe it would be clearer to just show an example, like `[[0, 0, 1], [1, 1, 2]]`

ch-kr: looking at this again, this might be a very wide array. I guess we'll cross that bridge if we get to it

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+    :param weighted: Whether to generalize the model to weighted least squares using
+      'possible_variants'. Default is False.
+    :return: A dictionary of intercepts and slopes for plateau models of each
+      population. The key of the dictionary is either 'total' or 'pop', and the value
+      is a dictionary (or a list of list of dictionaries if
+      `pop_observed_variants_exprs` is specified) mapping cpg BooleanExpression to a
+      intercept and a slope.
+    """
```
ch-kr, line 607:
```suggestion
    :return: A dictionary of intercepts and slopes for plateau models of each
      population. The possible keys for this dictionary are 'total' and 'pop', and the values
      are a dictionary (for 'total') or a list of list of dictionaries (for 'pop').
      The key of the value dictionary is CpG status (BooleanExpression), and the value is an ArrayExpression containing intercept and slope values.
    """
```

ch-kr (commented):
a few more suggestions! thanks for making the array agg change

gnomad/utils/constraint.py
averywpx: Just to confirm. Is it ok to have such wide array?

ch-kr: we won't know until it causes issues, though you could ask Tim?

averywpx: I see. It works fine on v2 but may raise error on v4. It's Tim's suggestion to pass in an ArrayExpression instead of dictionary. I guess we will see when testing the pipeline on v4!

gnomad/utils/constraint.py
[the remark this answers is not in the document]
averywpx: I will add the check if I can remove the `pop_lengths` downstream. I will add a note here!

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,287 @@ def collapse_strand(
...
+    :param possible_variants_expr: Int64Expression of the possible variant counts
+        for each combination of keys in `ht`.
+    :param pop_observed_variants_array_expr: Nested ArrayExpression with all observed
+        variant counts ArrayNumericExpressions for specified populations. e.g. [[1,1,1]
+        [1,1,1]] Default is None.
+    :param weighted: Whether to generalize the model to weighted least squares using
+        'possible_variants'. Default is False.
+    :return: A dictionary of intercepts and slopes for plateau models of each
+        population. The possible keys for this dictionary are 'total' and 'pop', and
+        the values are a dictionary (for 'total') or a list of list of dictionaries
```
ch-kr, line 606: maybe "nested list of dictionaries" would also be easier to read here than "list of list of", with an example, like `'afr': <ArrayExpression of type array<dict<bool, array<float64>>>>`

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,287 @@ def collapse_strand(
...
+
+    The x and y of the coverage model:
+    - x: log10('exome_coverage') at low coverage site
+    - y: sum('observed_variants')/ (`high_coverage_scale_factor` * sum('possible_variants' * 'mu_snp') at low coverage site
+    where `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage site
+
+    :param low_coverage_oe_expr: The Float64Expression of observed:expected ratio
+        for a given coverage level.
+    :param log_coverage_expr: The Float64Expression of log10 coverage.
+    :return: Tuple with intercept and slope of the model.
```
ch-kr, line 658:
```suggestion
    :return: StructExpression with intercept and slope of the model.
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,287 @@ def collapse_strand(
...
+    """
+    return hl.agg.linreg(low_coverage_oe_expr, [1, log_coverage_expr])
+
+
+def get_all_pop_lengths(
+    ht,
```
ch-kr, line 664:
```suggestion
    ht: hl.Table,
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,287 @@ def collapse_strand(
...
+
+    :param low_coverage_oe_expr: The Float64Expression of observed:expected ratio
+        for a given coverage level.
+    :param log_coverage_expr: The Float64Expression of log10 coverage.
+    :return: Tuple with intercept and slope of the model.
+    """
+    return hl.agg.linreg(low_coverage_oe_expr, [1, log_coverage_expr])
+
+
+def get_all_pop_lengths(
```
ch-kr, line 663: maybe add a TODO in this function that you are going to revisit it

ch-kr (commented):
a few more small suggestions (some on code that I suggested, sorry!). almost there!

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,287 @@ def collapse_strand(
...
+
+    # Build plateau models.
+    plateau_models_agg_expr = build_plateau_models(
+        cpg_expr=high_cov_group_ht.cpg,
+        mu_snp_expr=high_cov_group_ht.mu_snp,
+        observed_variants_expr=high_cov_group_ht.observed_variants,
+        possible_variants_expr=high_cov_group_ht.possible_variants,
+        pop_observed_variants_array_expr=hl.array(
+            [high_cov_group_ht[f"observed_{pop}"] for pop in pops]
+        ),
```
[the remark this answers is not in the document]
averywpx: I added `hl.array` since the hail documentation said the type of input parameter is `ArrayExpression`. I just tested the code without `hl.array` and it works as well. I will get rid of it!

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+    """
+    Build coverage model.
+
+    This function uses linear regression to build a model of log10(coverage) to correct
+    proportion of expected variation at low coverage sites.
+
+    The x and y of the coverage model:
+    - x: log10('exome_coverage') at low coverage site
+    - y: sum('observed_variants')/ (`high_coverage_scale_factor` * sum('possible_variants' * 'mu_snp') at low coverage site
+    where `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage site
```
ch-kr, line 649:
```suggestion
    - y: sum('observed_variants')/ (`high_coverage_scale_factor` * sum('possible_variants' * 'mu_snp') at low coverage sites
    where `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage sites
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+    The x and y of the plateau models:
+    - x: `mu_snp` - mutation rate
+    - y: proportion observed ('observed_variants' or 'observed_{pop}' / 'possible_variants')
+
+    :param cpg_expr: BooleanExpression noting whether a site is a CPG site.
+    :param mu_snp_expr: Float64Expression of the mutation rate.
+    :param observed_variants_expr: Int64Expression of the observed variant counts
+        for each combination of keys in `ht`.
+    :param possible_variants_expr: Int64Expression of the possible variant counts
+        for each combination of keys in `ht`.
```
ch-kr, line 591: probably best to remove references to `ht` here since this function no longer takes a table as input

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+        'possible_variants'. Default is False.
+    :return: A dictionary of intercepts and slopes for plateau models of each
+    population. The keys for this dictionary are 'total' and 'pop'. The values for
+    'total' is a dictionary (e.g., <DictExpression of type dict<bool,
+    array<float64>>>), and the value for 'pop' is a nested list of dictionaries (e.g.,
+    <ArrayExpression of type array<array<dict<bool, array<float64>>>>>). The key of the
+    dictionary in the nested list is CpG status (BooleanExpression), and the value is
+    an ArrayExpression containing intercept and slope values.
+    """
+    # Build a plateau model using all the sites in the Table.
```
ch-kr, line 605:
```suggestion
    # Build plateau models for all sites
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+            lambda pop_observed_variants_expr: hl.agg.array_agg(
+                lambda pop_observed_variants: hl.agg.group_by(
+                    cpg_expr,
+                    hl.agg.linreg(
+                        pop_observed_variants / possible_variants_expr,
+                        [1, mu_snp_expr],
+                        weight=possible_variants_expr,
+                    ).beta,
+                ),
+                pop_observed_variants_expr,
```
ch-kr, line 629: technically this is also an array. I wonder if there are more descriptive variable names you could use in this agg?

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+        observed_variants_expr=high_cov_group_ht.observed_variants,
+        possible_variants_expr=high_cov_group_ht.possible_variants,
+        pop_observed_variants_array_expr=[
+            high_cov_group_ht[f"observed_{pop}"] for pop in pops
+        ],
+        weighted=weighted,
+    )
+    _plateau_models = dict(
+        high_cov_group_ht.aggregate(hl.struct(**plateau_models_agg_expr))
+    )
```
ch-kr, line 529:
```suggestion
    )
    
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+    # Map the models to their corresponding populations if pops is specified.
```
ch-kr, line 530: this (and lines 527-529 above) will need to be updated, since 531-534 should get nested within an if statement. maybe something like this
```
if pops:
    _plateau_models = dict(
        high_cov_group_ht.aggregate(hl.struct(**plateau_models_agg_expr))
    )
    pop_models = _plateau_models["pop"]
    plateau_models = {pop: hl.literal(pop_models[idx]) for idx, pop in enumerate(pops)}
    plateau_models["total"] = _plateau_models["total"]
    plateau_models = hl.struct(**plateau_models)
else:
    plateau_models = high_cov_group_ht.aggregate(hl.struct(**plateau_models_agg_expr))
```

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,285 @@ def collapse_strand(
...
+
+    # Build plateau models.
+    plateau_models_agg_expr = build_plateau_models(
+        cpg_expr=high_cov_group_ht.cpg,
+        mu_snp_expr=high_cov_group_ht.mu_snp,
+        observed_variants_expr=high_cov_group_ht.observed_variants,
+        possible_variants_expr=high_cov_group_ht.possible_variants,
+        pop_observed_variants_array_expr=[
+            high_cov_group_ht[f"observed_{pop}"] for pop in pops
+        ],
```
ch-kr, line 524: it just occurred to me that this line will not work with the function as written (pops defaults to ()). how about something like this
```suggestion
        pop_observed_variants_array_expr=[
            high_cov_group_ht[f"observed_{pop}"] for pop in pops
        ] if pops else None,
```

ch-kr (changes requested):
a few cosmetic changes and a bug I previously missed (sorry!)

gnomad/utils/constraint.py
averywpx: Sorry it's my bad. It worked before I got rid of the redundancy and I forgot to double check. I will fix it!

gnomad/utils/constraint.py
```diff
@@ -519,17 +519,22 @@ def build_models(
...
+
+    pop_models = {}
+    if pops:
+        # Map the models to their corresponding populations if pops is specified.
+        pop_models = _plateau_models["pop"]
+        plateau_models = {
+            pop: hl.literal(pop_models[idx]) for idx, pop in enumerate(pops)
+        }
     plateau_models["total"] = _plateau_models["total"]
     plateau_models = hl.struct(**plateau_models)
```
ch-kr, line 539: these two lines (with 527-529 above) aggregate the plateau models into a struct, convert them into a dictionary, and then reformats them again into a struct, which is a bit redundant. using an if/else statement here (while not as clean) would avoid these unnecessary reformats

gnomad/utils/constraint.py
```diff
@@ -585,24 +590,22 @@ def build_plateau_models(
...
     :param weighted: Whether to generalize the model to weighted least squares using
         'possible_variants'. Default is False.
     :return: A dictionary of intercepts and slopes for plateau models of each
-    population. The keys for this dictionary are 'total' and 'pop'. The values for
-    'total' is a dictionary (e.g., <DictExpression of type dict<bool,
-    array<float64>>>), and the value for 'pop' is a nested list of dictionaries (e.g.,
-    <ArrayExpression of type array<array<dict<bool, array<float64>>>>>). The key of the
-    dictionary in the nested list is CpG status (BooleanExpression), and the value is
-    an ArrayExpression containing intercept and slope values.
+        population. The keys for this dictionary are 'total' and 'pop'. The values for
```
ch-kr, line 601: maybe "group" here is better than "population"?

gnomad/utils/constraint.py
```diff
@@ -613,11 +616,11 @@ def build_plateau_models(
             ).beta,
         )
     }
-    if pop_observed_variants_array_expr is not None:
+    if pops_observed_variants_array_expr:
         # Build plateau models using sites in population downsamplings if
         # population is specified.
         plateau_models_agg_expr["pop"] = hl.agg.array_agg(
-            lambda pop_observed_variants_expr: hl.agg.array_agg(
+            lambda pop_observed_variants_array_expr: hl.agg.array_agg(
```
ch-kr, line 623: how about `pop_obs_var_array_expr` to distinguish these variants a bit more?

ch-kr (commented):
a few more small comments

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,290 @@ def collapse_strand(
...
+    :param mu_snp_expr: Float64Expression of the mutation rate.
+    :param observed_variants_expr: Int64Expression of the observed variant counts.
+    :param possible_variants_expr: Int64Expression of the possible variant counts.
+    :param pops_observed_variants_array_expr: Nested ArrayExpression with all observed
+        variant counts ArrayNumericExpressions for specified populations. e.g., `[[1,1,
+        1],[1,1,1]]`. Default is None.
+    :param weighted: Whether to generalize the model to weighted least squares using
+        'possible_variants'. Default is False.
+    :return: A dictionary of intercepts and slopes of plateau models for all set and
+        for populations. The keys for this dictionary are 'total' and 'pop'. The values
```
averywpx, line 603: I guess your purpose is to emphasize the dictionary won't return separate aggregators for each population. I rephrased it a bit but didn't use "group" because I used it in `count_variants_by_group` and `downsampling_counts_expr` when grouping tables. We can definitely use "group" if you think it won't confuse people!

> these two lines (with 527-529 above) aggregate the plateau models into a struct, convert them into a dictionary, and then reformats them again into a struct, which is a bit redundant. using an if/else statement here (while not as clean) would avoid these unnecessary reformats

ch-kr: I'm a little confused; I think your comment is linked to another comment (not the one included below). This edit looks good, though maybe this wording is a bit clearer:
```
A dictionary of intercepts and slopes of plateau models for all sites and (optionally) per population.
```

gnomad/utils/constraint.py
```diff
@@ -597,9 +599,9 @@ def build_plateau_models(
         1],[1,1,1]]`. Default is None.
     :param weighted: Whether to generalize the model to weighted least squares using
         'possible_variants'. Default is False.
-    :return: A dictionary of intercepts and slopes for plateau models of each
-        population. The keys for this dictionary are 'total' and 'pop'. The values for
-        'total' is a dictionary (e.g., <DictExpression of type dict<bool,
+    :return: A dictionary of intercepts and slopes of plateau models for all set and
+        for populations. The keys for this dictionary are 'total' and 'pop'. The values
```
ch-kr, line 603:
```suggestion
        for populations. The possible keys for this dictionary are 'total' and 'pop'. The values
```
adding "possible" since this function won't add the 'pop' key by default

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,290 @@ def collapse_strand(
...
+            high_cov_group_ht[f"observed_{pop}"] for pop in pops
+        ],
+        weighted=weighted,
+    )
+    if pops:
+        _plateau_models = dict(
+            high_cov_group_ht.aggregate(hl.struct(**plateau_models_agg_expr))
+        )
+        # Map the models to their corresponding populations if pops is specified.
+        pop_models = _plateau_models["pop"]
```
ch-kr, line 532:
```suggestion
         # Map the models to their corresponding populations if pops is specified.
        _plateau_models = dict(
            high_cov_group_ht.aggregate(hl.struct(**plateau_models_agg_expr))
        )
        pop_models = _plateau_models["pop"]
```
the comment might make a bit more sense at the top of this code block

ch-kr (commented):
a few small docstring suggestions

gnomad/utils/constraint.py
averywpx: Sorry I linked the wrong comment :( This is the one I wanted to link: 

> maybe "group" here is better than "population"?

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,291 @@ def collapse_strand(
...
+    """
+    Build coverage model.
+
+    This function uses linear regression to build a model of log10(coverage) to correct
+    proportion of expected variation at low coverage sites.
+
+    The x and y of the coverage model:
+    - x: log10('exome_coverage') at low coverage site
+    - y: sum('observed_variants')/ (`high_coverage_scale_factor` * sum('possible_variants' * 'mu_snp') at low coverage sites
+    where `high_coverage_scale_factor` = sum('observed_variants') / sum('possible_variants' * 'mu_snp') at high coverage sites
```
ch-kr, line 655:
```suggestion
    - x: `log_coverage_expr`
    - y: low_coverage_oe_expr`
```
(editing this formula for this function)

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,291 @@ def collapse_strand(
...
+    possible_variants_expr: hl.expr.Int64Expression,
+    pops_observed_variants_array_expr: List[hl.expr.ArrayExpression] = [],
+    weighted: bool = False,
+) -> Dict[str, Union[Dict[bool, hl.expr.ArrayExpression], hl.ArrayExpression]]:
+    """
+    Build plateau models to calibrate mutation rate to compute predicted proportion observed value.
+
+    The x and y of the plateau models:
+    - x: `mu_snp` - mutation rate
+    - y: proportion observed ('observed_variants' or 'observed_{pop}' / 'possible_variants')
```
ch-kr, line 591:
```suggestion
    - x: `mu_snp_expr`
    - y: `observed_variants_expr` / `possible_variants_expr`
```
(editing this formula for this function)

ch-kr (commented):
two more docstring edits

gnomad/utils/constraint.py
```diff
@@ -599,14 +599,13 @@ def build_plateau_models(
...
         'possible_variants'. Default is False.
-    :return: A dictionary of intercepts and slopes of plateau models for all sites and
-        populations (optional). The keys for this dictionary are 'total' and 'pop'
-        (optional). The values for 'total' is a dictionary (e.g., <DictExpression of
-        type dict<bool,array<float64>>>), and the value for 'pop' is a nested list of
-        dictionaries (e.g., <ArrayExpression of type array<array<dict<bool,
-        array<float64>>>>>). The key of the dictionary in the nested list is CpG status
-        (BooleanExpression), and the value is an ArrayExpression containing intercept
-        and slope values.
+    :return: A dictionary of intercepts and slopes of plateau models. The keys are
```
averywpx, line 602: We may still want to specify `total` and `pop` are keys in this dict. I rephrase it a bit but feel free to change!

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,289 @@ def collapse_strand(
...
+    possible_variants_expr: hl.expr.Int64Expression,
+    pops_observed_variants_array_expr: List[hl.expr.ArrayExpression] = [],
+    weighted: bool = False,
+) -> Dict[str, Union[Dict[bool, hl.expr.ArrayExpression], hl.ArrayExpression]]:
+    """
+    Build plateau models to calibrate mutation rate to compute predicted proportion observed value.
+
+    The x and y of the plateau models:
+    - x: `mu_snp_expr`
+    - y: `observed_variants_expr` (or `pops_observed_variants_array_expr`[index] if `pops` is specified) / `possible_variants_expr`
```
ch-kr, line 591:
```suggestion
    - y: `observed_variants_expr` / `possible_variants_expr`
    or `pops_observed_variants_array_expr`[index] / `possible_variants_expr`
    if `pops` is specified
```
optional

gnomad/utils/constraint.py
ch-kr: this looks good to me

gnomad/utils/constraint.py
```diff
@@ -360,3 +412,289 @@ def collapse_strand(
...
+) -> hl.expr.StructExpression:
+    """
+    Build coverage model.
+
+    This function uses linear regression to build a model of log10(coverage) to correct
+    proportion of expected variation at low coverage sites.
+
+    The x and y of the coverage model:
+    - x: `log_coverage_expr`
+    - y: low_coverage_oe_expr`
```
ch-kr, line 653:
```suggestion
    - y: `low_coverage_oe_expr`
```

ch-kr (approved):
one small docstring fix, one optional docstring change, and one comment, but LGTM!

gnomad/utils/constraint.py
```diff
@@ -53,35 +61,45 @@ def count_variants_by_group(
     """
     Count number of observed or possible variants by context, ref, alt, and optionally methylation_level.
 
-    Performs variant count aggregations based on specified criteria (`count_singletons`, `count_downsamplings`, and
-    `max_af`), and grouped by: 'context', 'ref', 'alt', 'methylation_level' (optional), and all annotations provided
-    in `additional_grouping`.
+    Performs variant count aggregations based on specified criteria
+    (`count_singletons`, `count_downsamplings`, and `max_af`), and grouped by:
+    'context', 'ref', 'alt', 'methylation_level' (optional), and all annotations
```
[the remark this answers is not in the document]
averywpx: Yes! Maybe it will be useful for v4 so we kept this.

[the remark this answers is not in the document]
ch-kr: wouldn't we still want to separate variants by methylation level? otherwise agree it's good to keep the code flexible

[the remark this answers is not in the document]
averywpx: I remember we decided to keep this because we want to it to be generic in `gnomad_method`. It will also be used in building mutation rate table as well (see [here](https://github.com/broadinstitute/gnomad_lof/blob/e09d9fdc13d43c8ac996c93735ed206ba41ed473/constraint_utils/constraint_basics.py#L371)).

PR #486, opened 2022-09-28
Add more file types to `file_exists` for checking '_SUCCESS'

jkgoodrich:
Add `check_file_exists_raise_error` to use `file_exists` to check the existence of all files in a list of file paths with options to raise an error.

### DIFF
--- a/gnomad/utils/file_utils.py
+++ b/gnomad/utils/file_utils.py
@@ -102,7 +102,7 @@ def parallel_file_exists(fpaths: List[str], parallelism: int = 750) -> Dict[str,
     )
 
 
-def file_exists(fname: str, overwrite: Optional[bool] = None) -> bool:
+def file_exists(fname: str) -> bool:
     """
     Check whether a file exists.
 
@@ -110,7 +110,6 @@ def file_exists(fname: str, overwrite: Optional[bool] = None) -> bool:
     If the file is a Hail file (.ht, .mt, .bm, .parquet, and .vds extensions), it checks that _SUCCESS is present.
 
     :param fname: File name.
-    :param overwrite: Optional boolean that will raise an error if set to False and file exists.
     :return: Whether the file exists.
     """
     fext = os.path.splitext(fname)[1]
@@ -127,37 +126,40 @@ def file_exists(fname: str, overwrite: Optional[bool] = None) -> bool:
 
     exists = all([exists_func(p) for p in paths])
 
-    if exists and overwrite is False:
-        raise DataException(
-            f"{fname} already exists and the overwrite option was not used!"
-        )
-
     return exists
 
 
-def file_list_exists(
-    fname_list: List[str],
-    overwrite: bool = None,
+def check_file_exists_no_overwrite(
+    fname: Union[str, List[str]],
+    overwrite: bool = True,
 ) -> bool:
     """
     Check whether all files in the list exist and raise an exception if any of them exist and overwrite is False.
 
-    :param fname_list: List of file paths to check the existence of.
-    :param overwrite: Optional boolean that will raise an exception if set to False and any of the resources exist.
-    :return: Boolean indicating if all files in `fname_list` exist.
+    Useful when writing out to files at the end of a pipeline, but want to first check if the file already exists and
+    requires the file to be removed or overwrite specified so the pipeline doesn't fail.
+
+    :param fname: File path, or list of file paths to check the existence of.
+    :param overwrite: Boolean that will raise an exception if set to False and any of the files exist.
+    :return: Boolean indicating if `fname` or all files in `fname` exist.
     """
-    exists = []
-    exceptions_raised = []
-    for fname in fname_list:
-        try:
-            exists.append(file_exists(fname, overwrite))
-        except DataException as e:
-            exceptions_raised.append(str(e))
+    if isinstance(fname, str):
+        fname = [fname]
 
-    if exceptions_raised:
-        raise DataException("\n".join(exceptions_raised))
+    all_exist = True
+    data_exceptions = []
+    for f in fname:
+        exists = file_exists(f)
+        all_exist &= exists
+        if exists and overwrite is False:
+            data_exceptions.append(f)
 
-    return all(exists)
+    if data_exceptions:
+        raise DataException(
+            f"The following files already exist and the overwrite option was not used! {','.join(data_exceptions)}"
+        )
+
+    return all_exist
 
 
 def write_temp_gcs(
--- a/gnomad/utils/file_utils.py
+++ b/gnomad/utils/file_utils.py
@@ -16,6 +16,8 @@ from hailtop.aiotools.router_fs import RouterAsyncFS
 from hailtop.aiogoogle import GoogleStorageAsyncFS
 from hailtop.utils import bounded_gather, tqdm
 
+from gnomad.resources.resource_utils import DataException
+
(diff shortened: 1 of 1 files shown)


gnomad/utils/file_utils.py
```diff
@@ -100,23 +102,62 @@ def parallel_file_exists(fpaths: List[str], parallelism: int = 750) -> Dict[str,
...
+
+def file_list_exists(
+    fname_list: List[str],
+    overwrite: bool = None,
+) -> bool:
+    """
+    Check whether all files in the list exist and raise an exception if any of them exist and overwrite is False.
+
+    :param fname_list: List of file paths to check the existence of.
+    :param overwrite: Optional boolean that will raise an exception if set to False and any of the resources exist.
```
averywpx, line 146:
```suggestion
    :param overwrite: Optional boolean that will raise an exception if set False and any of the resources doesn't exist.
```

gnomad/utils/file_utils.py
```diff
@@ -100,23 +102,62 @@ def parallel_file_exists(fpaths: List[str], parallelism: int = 750) -> Dict[str,
...
     Check whether a file exists.
 
     Supports either local or Google cloud (gs://) paths.
-    If the file is a Hail file (.ht, .mt extensions), it checks that _SUCCESS is present.
+    If the file is a Hail file (.ht, .mt, .bm, .parquet, and .vds extensions), it checks that _SUCCESS is present.
 
-    :param fname: File name
-    :return: Whether the file exists
+    :param fname: File name.
+    :param overwrite: Optional boolean that will raise an error if set to False and file exists.
```
averywpx, line 113:
```suggestion
    :param overwrite: Optional boolean that will raise an error if set to False and file doesn't exist.
```

averywpx (changes requested):

gnomad/utils/file_utils.py
jkgoodrich: Actually in this case I want to raise the error if it does exist. Use case is when the file exists and we are trying to overwrite it, but we forgot to specify overwrite=True

gnomad/utils/file_utils.py
jkgoodrich: Same comment as above

gnomad/utils/file_utils.py
averywpx: What if I just want to check if any file in the list doesn't exist and which are they? Setting `overwrite` to False to figure out make thing complicated. This question may be asked more, because the function doesn't overwrite anything.

jkgoodrich: This function isn't for a list, it's just for a single file. By default the `overwrite` parameter is None and will not be used at all, this will then raise no error, it will simply do what it did before and return True or False. I will change to make another function that calls this one and has a clearer name that includes overwrite, I was just trying to put it into a single function. For the other function you again don't need to pass in overwrite.

gnomad/utils/file_utils.py
```diff
@@ -100,23 +102,62 @@ def parallel_file_exists(fpaths: List[str], parallelism: int = 750) -> Dict[str,
     )
 
 
-def file_exists(fname: str) -> bool:
+def file_exists(fname: str, overwrite: Optional[bool] = None) -> bool:
     """
     Check whether a file exists.
```
averywpx, line 108: I feel that the `overwrite` parameter needs use case explanation, since overwrite is kind of out of the functions scope (just to check if file exists) and it causes some confusion. Maybe just me, so it's up to you :)
```suggestion
The function will optionally raise an error if `overwrite` is set to False and file exists, when the the function is called to check if the new created file path exist.
```

gnomad/utils/file_utils.py
```diff
@@ -100,23 +102,62 @@ def parallel_file_exists(fpaths: List[str], parallelism: int = 750) -> Dict[str,
...
+    return exists
+
+
+def file_list_exists(
+    fname_list: List[str],
+    overwrite: bool = None,
+) -> bool:
+    """
+    Check whether all files in the list exist and raise an exception if any of them exist and overwrite is False.
+
```
averywpx, line 144: I feel that the `overwrite` parameter needs use case explanation, since overwrite is kind of out of the functions scope (just to check if file in list exists) and it causes some confusion. Maybe just me, so it's up to you :)
```suggestion
The function will optionally raise an error if `overwrite` is set to False and any file exists, when the the function is called to check if the new created file paths exist.
```

gnomad/utils/file_utils.py
jkgoodrich: Let me know if the changes I made works. They will work for my need

gnomad/utils/file_utils.py
```diff
@@ -105,18 +107,59 @@ def file_exists(fname: str) -> bool:
...
+    if isinstance(fname, str):
+        fname = [fname]
+
+    all_exist = True
+    data_exceptions = []
+    for f in fname:
+        exists = file_exists(f)
+        all_exist &= exists
+        if exists and overwrite is False:
+            data_exceptions.append(f)
```
averywpx, line 155:
```suggestion
    for f in fname:
        exists = file_exists(f)
        all_exist &= exists
        if exists and raise_error == "exist":
            data_exceptions.append(f)
        elif not exist and raise_error == "not exist":
            data_exceptions.append(f)
```

gnomad/utils/file_utils.py
```diff
@@ -105,18 +107,59 @@ def file_exists(fname: str) -> bool:
...
+
+    if data_exceptions:
+        raise DataException(
+            f"The following files already exist and the overwrite option was not used! {','.join(data_exceptions)}"
+        )
```
averywpx, line 160:
```suggestion
    if data_exceptions:
        if raise_error == "exist":
            raise DataException(
                f"The following files already exist and the overwrite option was not used! {','.join(data_exceptions)}"
            )
        elif raise_error == "not exist":
            raise DataException(
                f"The following files doesn't exist! {','.join(data_exceptions)}"
            )
```

averywpx (commented):
I add some suggestions so that the functions can be used in other cases as well. It's up to you if you want to change it :)

averywpx (approved):

gnomad/utils/file_utils.py
```diff
@@ -105,18 +107,59 @@ def file_exists(fname: str) -> bool:
...
+        exists_func = os.path.isfile
+
+    exists = all([exists_func(p) for p in paths])
+
+    return exists
+
+
+def check_file_exists_no_overwrite(
+    fname: Union[str, List[str]],
+    overwrite: bool = True,
```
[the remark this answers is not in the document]
jkgoodrich: I tried adding in something with more flexibility, let me know if that works

gnomad/utils/file_utils.py
```diff
@@ -105,18 +107,71 @@ def file_exists(fname: str) -> bool:
...
+    exists = all([exists_func(p) for p in paths])
+
+    return exists
+
+
+def check_file_exists_raise_error(
+    fname: Union[str, List[str]],
+    error_if_exists: bool = False,
+    error_if_not_exists: bool = False,
+    error_if_exists_msg: str = "The following files already exist and the overwrite option was not used: ",
```
averywpx, line 136:
```suggestion
    error_if_exists_msg: str = "The following files already exist: ",
```

averywpx (commented):
Just one single comment!

PR #489, opened 2022-10-05
Add additional_samples_to_drop option to `run_pca_with_relateds`

klaricch:
-Resolves issue where the `autosomes_only` filter was ignored if `related_samples_to_drop` was provided to `run_pca_with_relateds`
-Adds an additional_samples_to_drop option to `run_pca_with_relateds`, which will remove additional provided samples when generating the PCs and then project them in the PC space

### DIFF
--- a/gnomad/sample_qc/ancestry.py
+++ b/gnomad/sample_qc/ancestry.py
@@ -259,52 +259,60 @@ def assign_population_pcs(
 def run_pca_with_relateds(
     qc_mt: hl.MatrixTable,
     related_samples_to_drop: Optional[hl.Table],
+    outlier_samples_to_drop: Optional[hl.Table],
     n_pcs: int = 10,
     autosomes_only: bool = True,
 ) -> Tuple[List[float], hl.Table, hl.Table]:
     """
-    Run PCA excluding the given related samples, and project those samples in the PC space to return scores for all samples.
+    Run PCA excluding the given related or outlier samples, and project those samples in the PC space to return scores for all samples.
 
-    The `related_samples_to_drop` Table has to be keyed by the sample ID and all samples present in this
+    The `related_samples_to_drop` and `outlier_samples_to_drop` Tables have to be keyed by the sample ID and all samples present in these
     table will be excluded from the PCA.
 
     The loadings Table returned also contains a `pca_af` annotation which is the allele frequency
     used for PCA. This is useful to project other samples in the PC space.
 
     :param qc_mt: Input QC MT
-    :param related_samples_to_drop: Optional table of related samples to drop
+    :param related_samples_to_drop: Optional table of related samples to drop when generating the PCs, these samples will be projected in the PC space
+    :param outlier_samples_to_drop: Optional table of outlier samples to drop when generating the PCs, these samples will be projected in the PC space
     :param n_pcs: Number of PCs to compute
     :param autosomes_only: Whether to run the analysis on autosomes only
     :return: eigenvalues, scores and loadings
     """
-    unrelated_mt = qc_mt.persist()
+    qc_mt = qc_mt.persist()
 
     if autosomes_only:
-        unrelated_mt = filter_to_autosomes(unrelated_mt)
+        qc_mt = filter_to_autosomes(qc_mt)
+
+    # 'pca_mt' is the MatrixTable to use for generating the PCs
+    # If samples to drop are provided in related_samples_to_drop or outlier_samples_to_drop, 'project_pca_mt' will also be generated and will contain the samples to project in the PC space
+    pca_mt = qc_mt
 
     if related_samples_to_drop:
-        unrelated_mt = qc_mt.filter_cols(
-            hl.is_missing(related_samples_to_drop[qc_mt.col_key])
+        pca_mt = pca_mt.filter_cols(
+            hl.is_missing(related_samples_to_drop[pca_mt.col_key])
+        )
+    if outlier_samples_to_drop:
+        pca_mt = pca_mt.filter_cols(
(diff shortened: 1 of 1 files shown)


gnomad/sample_qc/ancestry.py
```diff
@@ -259,52 +259,60 @@ def assign_population_pcs(
 def run_pca_with_relateds(
     qc_mt: hl.MatrixTable,
     related_samples_to_drop: Optional[hl.Table],
+    outlier_samples_to_drop: Optional[hl.Table],
```
jkgoodrich, line 262: One thought is to modify this to additional samples to drop or something like that. Right now for removing unreleasable samples we just combine them with the related sample HT that we pass to this function, but we could just pass that table here too.

gnomad/sample_qc/ancestry.py
```diff
@@ -259,52 +259,60 @@ def assign_population_pcs(
...
     :param qc_mt: Input QC MT
-    :param related_samples_to_drop: Optional table of related samples to drop
+    :param related_samples_to_drop: Optional table of related samples to drop when generating the PCs, these samples will be projected in the PC space
+    :param outlier_samples_to_drop: Optional table of outlier samples to drop when generating the PCs, these samples will be projected in the PC space
     :param n_pcs: Number of PCs to compute
     :param autosomes_only: Whether to run the analysis on autosomes only
     :return: eigenvalues, scores and loadings
     """
-    unrelated_mt = qc_mt.persist()
+    qc_mt = qc_mt.persist()
```
jkgoodrich, line 282: I think this is very old, lets remove the persist. The user can checkpoint before this if they need to

jkgoodrich (commented):
Mostly one comment that will change a few names and comments/docstrings if you agree.

gnomad/sample_qc/ancestry.py
klaricch: I like additional samples to drop. Don't we drop the unreleasable samples and non-high-quality samples completely before running this step?

jkgoodrich: Nope https://github.com/broadinstitute/gnomad_qc/blob/a828b1529a439fc41ecc73647a442c7e63897701/gnomad_qc/v3/sample_qc/sample_qc.py#L457

```
    if not include_unreleasable_samples:
        logger.info("Excluding unreleasable samples for PCA.")
        samples_to_drop = samples_to_drop.union(
            qc_mt.filter_cols(
                ~project_meta.ht()[qc_mt.col_key].releasable
                | project_meta.ht()[qc_mt.col_key].exclude
            )
            .cols()
            .select()
        )
    else:
        logger.info("Including unreleasable samples for PCA")

    return run_pca_with_relateds(qc_mt, samples_to_drop, n_pcs=n_pcs)
```

We do drop samples that fail hard filtering

klaricch: ah ok, for the subpop analysis the unreleasable samples are dropped

jkgoodrich: Yeah, I think that's fine

gnomad/sample_qc/ancestry.py
```diff
@@ -259,52 +259,59 @@ def assign_population_pcs(
...
     :return: eigenvalues, scores and loadings
     """
-    unrelated_mt = qc_mt.persist()
 
     if autosomes_only:
-        unrelated_mt = filter_to_autosomes(unrelated_mt)
+        qc_mt = filter_to_autosomes(qc_mt)
+
+    # 'pca_mt' is the MatrixTable to use for generating the PCs
+    # If samples to drop are provided in related_samples_to_drop or additional_samples_to_drop, 'project_pca_mt' will also be generated and will contain the samples to project in the PC space
```
jkgoodrich, line 287:
```suggestion
    # If samples to drop are provided in 'related_samples_to_drop' or 'additional_samples_to_drop', 'project_pca_mt' will also be generated and will contain the samples to project in the PC space
```

jkgoodrich (approved):
looks good to me, only thing is that we might want to add a breaking change label because if anyone used `run_pca_with_relateds(qc_mt, related_samples_to_drop_table, n_pcs)` it won't work

gnomad/sample_qc/ancestry.py
```diff
@@ -259,52 +259,59 @@ def assign_population_pcs(
...
 
     :param qc_mt: Input QC MT
-    :param related_samples_to_drop: Optional table of related samples to drop
+    :param related_samples_to_drop: Optional table of related samples to drop when generating the PCs, these samples will be projected in the PC space
+    :param additional_samples_to_drop: Optional table of additional samples to drop when generating the PCs, these samples will be projected in the PC space
     :param n_pcs: Number of PCs to compute
     :param autosomes_only: Whether to run the analysis on autosomes only
     :return: eigenvalues, scores and loadings
     """
-    unrelated_mt = qc_mt.persist()
```
jkgoodrich, line 282:
```suggestion
```

gnomad/sample_qc/ancestry.py
```diff
@@ -259,52 +259,59 @@ def assign_population_pcs(
 def run_pca_with_relateds(
     qc_mt: hl.MatrixTable,
     related_samples_to_drop: Optional[hl.Table],
+    additional_samples_to_drop: Optional[hl.Table],
     n_pcs: int = 10,
```
jkgoodrich, line 263:
```suggestion
    related_samples_to_drop: Optional[hl.Table] = None,
    additional_samples_to_drop: Optional[hl.Table] = None,
    n_pcs: int = 10,
```

PR #490, opened 2022-10-07
Run isort on the repo to sort all imports and add isort to requirements-dev.in and commit hooks .pre-commit-config.yaml

jkgoodrich:
This will help us easily standardize import order without needing to think about it, or add in PR reviews. 

Let me know if you think this is a good idea for us to add to our other repos too


averywpx (approved):
I really like this idea! It reduces conflicts and really save our time on keeping the import functions in order. I will apply isort to the constraint repo as well. You will see the update no later than next week!

.github/workflows/ci.yml
```diff
@@ -29,6 +29,8 @@ jobs:
           pip install -r requirements-dev.txt
       - name: Check formatting
         run: black --check gnomad tests
+      - name: Check imports
+        run: isort --check-only gnomad tests
```
ch-kr, line 33: basic question, what is the difference between `--check` (used above for black) and `--check-only`?

gnomad/resources/import_resources.py
```diff
@@ -4,15 +4,15 @@
 import itertools
 import textwrap
 from inspect import getmembers
-from typing import Dict, Tuple, Optional
+from typing import Dict, Optional, Tuple
```
ch-kr, line 7: so it IS convention to use alphabetical imports!

gnomad/utils/release.py
```diff
@@ -2,12 +2,7 @@
 
 from typing import Dict, List, Optional
 
-from gnomad.resources.grch38.gnomad import (
-    GROUPS,
-    POPS,
-    SEXES,
-    SUBSETS,
-)
+from gnomad.resources.grch38.gnomad import GROUPS, POPS, SEXES, SUBSETS
```
ch-kr, line 5: this edit I'm a bit confused by -- the previous formatting seemed like it was from black. maybe something changed in the newest black version?

ch-kr (approved):
one small question and comment, but LGTM. thanks for finding and adding this functionality!!

.github/workflows/ci.yml
jkgoodrich: it's the same thing, they check but don't reformat, just named differently for black and isort

gnomad/resources/import_resources.py
jkgoodrich: Yup!!

gnomad/utils/release.py
jkgoodrich: yeah I'm not sure, I just tried running black on it again, but it didn't change it back

PR #493, opened 2022-10-08
Change to Black options to format long strings and add autopep8 to automatically wrap long comments

jkgoodrich:


.pre-commit-config.yaml
```diff
@@ -1,19 +1,24 @@
...
       - id: trailing-whitespace
   - repo: https://github.com/psf/black
-    rev: 22.3.0 # This should be kept in sync with the version in requirements-dev.in
+    rev: 22.10.0 # This should be kept in sync with the version in requirements-dev.in
     hooks:
       - id: black
         language_version: python3
         files: gnomad
+  - repo: https://github.com/pre-commit/mirrors-autopep8
+    rev: v1.7.0 # This should be kept in sync with the version in requirements-dev.in
```
mike-w-wilson, line 16: Slightly different than the reqiurements-dev.in which omits the `v`  in `v1.7.0`, not sure if this is needed or if it can be removed here as well?

requirements-dev.txt
```diff
@@ -22,13 +24,6 @@ filelock==3.4.0
     # via virtualenv
 identify==2.4.0
     # via pre-commit
-importlib-metadata==5.0.0
```
mike-w-wilson, line 25: Just curious why this is being removed?

mike-w-wilson (approved):
Just two questions but then LGTM!

requirements-dev.txt
jkgoodrich: requirements-dev.txt is automatically created by running `pip-compile requirements-dev.in` Looks like it's no longer needed by these. Might be because of this: https://realpython.com/lessons/importlibmetadata/

.pre-commit-config.yaml
jkgoodrich: I think it needs to match the tag in the github repo https://github.com/hhatto/autopep8/releases/tag/v1.7.0

PR #509, opened 2023-01-09
Add `determine_nearest_neighbors` to find nearest neighbors for each sample. Modify `compute_stratified_metrics_filter` to work with a `comparison_sample_expr` that specifies what samples to compare to for filtering, this works well with the output of `determine_nearest_neighbor`.

jkgoodrich:


gnomad/sample_qc/filtering.py
```diff
@@ -362,3 +442,196 @@ def merge_sample_qc_expr(
...
+            nbrs_ht = indexes_ht
+
+        # Add nearest_neighbors annotation to use instead of indexes.
+        nbrs_ht = nbrs_ht.add_index()
+        explode_nbrs_ht = nbrs_ht.key_by("idx").explode("nearest_neighbor_idxs")
+        explode_nbrs_ht = explode_nbrs_ht.transmute(
+            nbr=explode_nbrs_ht[hl.int64(explode_nbrs_ht.nearest_neighbor_idxs)].s
+        )
+        explode_nbrs_ht = explode_nbrs_ht.group_by("s").aggregate(
+            nearest_neighbors=hl.agg.collect_as_set(explode_nbrs_ht.nbr)
```
klaricch, line 617: does maintaining the order when collecting matter?

gnomad/sample_qc/filtering.py
```diff
@@ -111,75 +148,118 @@ def compute_stratified_metrics_filter(
...
-    :param ht: HT containing relevant sample QC metric annotations
-    :param qc_metrics: list of metrics (name and expr) for which to compute the critical values for filtering outliers
-    :param strata: List of annotations used for stratification. These metrics should be discrete types!
-    :param lower_threshold: Lower MAD threshold
-    :param upper_threshold: Upper MAD threshold
-    :param metric_threshold: Can be used to specify different (lower, upper) thresholds for one or more metrics
-    :param filter_name: Name of resulting filters annotation
-    :return: Table grouped by strata, with upper and lower threshold values computed for each sample QC metric
+    :param ht: HT containing relevant sample QC metric annotations.
+    :param qc_metrics: list of metrics (name and expr) for which to compute the
```
klaricch, line 157:
```suggestion
    :param qc_metrics: List of metrics (name and expr) for which to compute the
```

klaricch (changes requested):

gnomad/sample_qc/filtering.py
jkgoodrich: Oh good point, if we ever want to map to the `nearest_neighbor_dists` array it needs to be in the same order

gnomad/sample_qc/filtering.py
```diff
@@ -614,7 +617,7 @@ def determine_nearest_neighbors(
             nbr=explode_nbrs_ht[hl.int64(explode_nbrs_ht.nearest_neighbor_idxs)].s
         )
         explode_nbrs_ht = explode_nbrs_ht.group_by("s").aggregate(
-            nearest_neighbors=hl.agg.collect_as_set(explode_nbrs_ht.nbr)
+            nearest_neighbors=hl.agg.collect(explode_nbrs_ht.nbr)
```
klaricch, line 620: does collect vs collect_as_set function differently regarding maintaining order?

klaricch (approved):
lgtm (just need to add a label)

PR #511, opened 2023-02-08
Add `get_slope_int_relationship_expr` to get relationship between a pair of samples given slope and intercepts of lines to use as cutoffs.

jkgoodrich:
Fix `compute_related_samples_to_drop` when `min_related_hard_filter` or `filtered_samples` are used to build a list of additional samples to drop.

### DIFF
--- a/gnomad/sample_qc/relatedness.py
+++ b/gnomad/sample_qc/relatedness.py
@@ -265,6 +265,118 @@ def get_relationship_expr(  # TODO: The threshold detection could be easily auto
     )
 
 
+def get_slope_int_relationship_expr(
+    kin_expr: hl.expr.NumericExpression,
+    y_expr: hl.expr.NumericExpression,
+    parent_child_max_y: float,
+    second_degree_sibling_lower_cutoff_slope: float,
+    second_degree_sibling_lower_cutoff_intercept: float,
+    second_degree_upper_sibling_lower_cutoff_slope: float,
+    second_degree_upper_sibling_lower_cutoff_intercept: float,
+    duplicate_twin_min_kin: float = 0.42,
+    second_degree_min_kin: float = 0.1,
+    duplicate_twin_ibd1_min: float = -0.15,
+    duplicate_twin_ibd1_max: float = 0.1,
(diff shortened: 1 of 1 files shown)


gnomad/sample_qc/relatedness.py
```diff
@@ -681,7 +793,7 @@ def compute_related_samples_to_drop(
             "Computing samples related to too many individuals (>%d) for exclusion",
             min_related_hard_filter,
         )
-        gbi = relatedness_ht.annotate(s=list(relatedness_ht.key))
+        gbi = relatedness_ht.annotate(s=[relatedness_ht.key[0], relatedness_ht.key[1]])
```
mike-w-wilson, line 796: Is this change for readability or does casting no longer work?

mike-w-wilson (approved):
LGTM! -- two questions for my knowledge but not critical for merging

gnomad/sample_qc/relatedness.py
mike-w-wilson: Or I guess speed?

jkgoodrich: `list(relatedness_ht.key)` returns ['i', 'j'] for each `s`, I'm assuming this was something that changed and was not always the case, but I couldn't confirm.

gnomad/sample_qc/relatedness.py
```diff
@@ -265,6 +265,118 @@ def get_relationship_expr(  # TODO: The threshold detection could be easily auto
...
+    else:
+        dup_twin_ibd1_expr = True
+
+    return (
+        hl.case()
+        .when(kin_expr < second_degree_min_kin, UNRELATED)
+        .when(
+            kin_expr > duplicate_twin_min_kin,
+            hl.if_else(
+                dup_twin_ibd1_expr & (y_expr < parent_child_max_y),
```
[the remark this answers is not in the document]
jkgoodrich: Yeah, it's not really important for cuKING, but I added it for support with our PC-Relate IBD0, and yes I think it probably has to do with that long investigation
Screen Shot 2023-02-10 at 9 16 23 AM

PR #512, opened 2023-02-16
Added utility function to repartition HTs prior to join

ch-kr:
This PR adds a small utility function (`repartition_for_join`) that repartitions a Table to make a downstream join more efficient

### DIFF
--- a/gnomad/utils/file_utils.py
+++ b/gnomad/utils/file_utils.py
@@ -280,3 +280,44 @@ def read_list_data(input_file_path: str) -> List[str]:
         output.append(line.strip())
     f.close()
     return output
+
+
+def repartition_for_join(
+    ht_path: str,
+    new_partition_percent: Optional[float] = None,
+    n_partitions: Optional[int] = None,
+) -> hl.Table:
+    """
+    Repartition a Table prior to joining with another Table(s).
+
+    This repartitioning makes the join(s) much more efficient.
+
+    :param ht_path: Path to Table to repartition.
+    :param new_partition_percent: Percent of initial dataset partitions to use.
+        Value should be greater than 1.
+        Default is None.
+        Should not be specified if `n_partitions` is not None.
+    :param n_partitions: Number of partitions desired after repartition.
+        Default is None.
+        Should not be specified if `new_partition_percent` is not None.
+    :return: hl.Table
+    """
+    if not new_partition_percent and not n_partitions:
+        raise DataException(
+            "Must specify either new_partition_percent or n_partitions!"
+        )
+
+    ht = hl.read_table(ht_path)
+    if new_partition_percent:
+        if new_partition_percent < 1:
+            logger.warning(
+                "new_partition_percent value is less than 1! The new HT will have fewer"
+                " partitions than the original HT!"
+            )
+        partition_intervals = ht._calculate_new_partitions(
+            ht.n_partitions() * new_partition_percent
+        )
+    if n_partitions:
+        partition_intervals = ht._calculate_new_partitions(n_partitions)
+
+    return hl.read_table(ht_path, _intervals=partition_intervals)
--- a/gnomad/utils/file_utils.py
+++ b/gnomad/utils/file_utils.py
@@ -296,7 +296,7 @@ def repartition_for_join(
     :param ht_path: Path to Table to repartition.
     :param new_partition_percent: Percent of initial dataset partitions to use.
         Value should be greater than 1 so that input HT will have more
-        partitions for the join.
+        partitions for the join. Defaults to 1.1.
     :return: List of IntervalExpressions calculated over new set of partitions
         (number of partitions in HT * desired percent increase).
     """
--- a/gnomad/utils/file_utils.py
+++ b/gnomad/utils/file_utils.py
@@ -284,9 +284,8 @@ def read_list_data(input_file_path: str) -> List[str]:
 
 def repartition_for_join(
     ht_path: str,
-    new_partition_percent: Optional[float] = None,
-    n_partitions: Optional[int] = None,
-) -> hl.Table:
+    new_partition_percent: float,
+) -> List[hl.expr.IntervalExpression]:
     """
     Repartition a Table prior to joining with another Table(s).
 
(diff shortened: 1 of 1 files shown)


gnomad/utils/file_utils.py
```diff
@@ -280,3 +280,44 @@ def read_list_data(input_file_path: str) -> List[str]:
...
+                "new_partition_percent value is less than 1! The new HT will have fewer"
+                " partitions than the original HT!"
+            )
+        partition_intervals = ht._calculate_new_partitions(
+            ht.n_partitions() * new_partition_percent
+        )
+    if n_partitions:
+        partition_intervals = ht._calculate_new_partitions(n_partitions)
+
+    return hl.read_table(ht_path, _intervals=partition_intervals)
```
mike-w-wilson, line 323: So I think the script should return the partition_intervals(an array of intervals, one entry per partition) and the  original table re-read or just the partition_intervals. The performance gain we see in a join is because all the tables involved are partitioned using the same intervals. Right now, I don't know of a way to access the partition intervals cleanly from the returned table.

ch-kr: just to make sure I'm following, there are two thoughts here:

- the return type should be both the Table and the newly calculated intervals (hl.Table, hl.expr.ArrayExpression) or just the intervals?
- `n_partitions` isn't helpful for the join here?

I'm happy to change, though when I was doing quick joins for MPC, I actually just set the new partitions using `n_partitions` (I think I set them all to 15k with `_calculate_new_partitions`), and that sped up the join more than me simply repartitioning each HT on read to 15k partitions. Not sure if that wasn't best practices

mike-w-wilson: > the return type should be both the Table and the newly calculated intervals (hl.Table, hl.expr.ArrayExpression) or just the intervals?

Yeah so my understanding to what matters for the join efficiency is the intervals covered within each partition so returning the intervals would be most useful for joining because you then just use that in 
```
ht=hl.read_table(<path>, _intervals=partition_intervals)
```
The _intervals size is equal to n_partitions in this case so youre getting that for free... just getting extra info for where the table should be diced up.

>n_partitions isn't helpful for the join here?

It is helpful but not as helpful as `_intervals`. When you pass 15k to `_calculate_new_partitions` for table A you're saying I want 15k partitions on table A but that doesnt mean you'll get the same intervals in each partition for when you call it on table B. Since using `hl.read_table(<path>, _n_partitions=n_partitions)` calls `_calculcate_new_partitions` in hail's `read_table`

```
    if _intervals is not None and _n_partitions is not None:
        raise ValueError("'read_table' does not support both _intervals and _n_partitions")
    tr = ir.TableNativeReader(path, _intervals, _filter_intervals)
    ht = Table(ir.TableRead(tr, False, drop_row_uids=not _create_row_uids, _assert_type=_assert_type))

    if _n_partitions:
        intervals = ht._calculate_new_partitions(_n_partitions)
        return read_table(path, _intervals=intervals, _assert_type=_assert_type, _load_refs=_load_refs, _create_row_uids=_create_row_uids)
    return ht
```
 I'm guessing the tables are similar sizes for MPC? So `n_parititons` is probably making the intervals  close and speeding things up. The _intervals just gives us  the added benefit of partitions matching their start and end locus when the tables are mismatched. Sorry if I went extra basic here, maybe lets have a meeting if I'm not being clear?

ch-kr: No this is great, thanks! Will update.

One quick note:
> and I'm guessing the tables are similar sizes for MPC

They actually aren't, but this might be an instance where spark partitioning is a black box, and I'm not sure why/how using only `_n_partitions` helped speed up the joins so much

mike-w-wilson: Huh, my guess is it is a parallelization boost in spark? but yeah -- black box

gnomad/utils/file_utils.py
```diff
@@ -280,3 +280,30 @@ def read_list_data(input_file_path: str) -> List[str]:
         output.append(line.strip())
     f.close()
     return output
+
+
+def repartition_for_join(
+    ht_path: str,
+    new_partition_percent: float,
```
mike-w-wilson, line 287:
```suggestion
    new_partition_percent: float = 1.1,
```

gnomad/utils/file_utils.py
```diff
@@ -280,3 +280,30 @@ def read_list_data(input_file_path: str) -> List[str]:
...
+    Repartition a Table prior to joining with another Table(s).
+
+    This repartitioning makes the join(s) much more efficient.
+    For more information, see:
+    https://discuss.hail.is/t/room-for-improvement-when-joining-multiple-hts/2278/8
+
+    :param ht_path: Path to Table to repartition.
+    :param new_partition_percent: Percent of initial dataset partitions to use.
+        Value should be greater than 1 so that input HT will have more
+        partitions for the join.
```
mike-w-wilson, line 299:
```suggestion
        partitions for the join. Defaults to 1.1.
```

gnomad/utils/file_utils.py
```diff
@@ -280,3 +280,30 @@ def read_list_data(input_file_path: str) -> List[str]:
...
+    new_partition_percent: float,
+) -> List[hl.expr.IntervalExpression]:
+    """
+    Repartition a Table prior to joining with another Table(s).
+
+    This repartitioning makes the join(s) much more efficient.
+    For more information, see:
+    https://discuss.hail.is/t/room-for-improvement-when-joining-multiple-hts/2278/8
+
+    :param ht_path: Path to Table to repartition.
```
mike-w-wilson, line 296:
```suggestion
    :param ht_path: Path to Table to use for interval partition calculation. 
```

gnomad/utils/file_utils.py
```diff
@@ -280,3 +280,30 @@ def read_list_data(input_file_path: str) -> List[str]:
...
+
+
+def repartition_for_join(
+    ht_path: str,
+    new_partition_percent: float,
+) -> List[hl.expr.IntervalExpression]:
+    """
+    Repartition a Table prior to joining with another Table(s).
+
+    This repartitioning makes the join(s) much more efficient.
```
mike-w-wilson, line 292:
```suggestion
    Calculate  partitions' intervals using the passed Table. 

    Reading in all Tables using these partition intervals using  `_intervals` 
    before they are joined, makes the join(s) much more efficient.
```

gnomad/utils/file_utils.py
```diff
@@ -280,3 +280,30 @@ def read_list_data(input_file_path: str) -> List[str]:
...
+    For more information, see:
+    https://discuss.hail.is/t/room-for-improvement-when-joining-multiple-hts/2278/8
+
+    :param ht_path: Path to Table to repartition.
+    :param new_partition_percent: Percent of initial dataset partitions to use.
+        Value should be greater than 1 so that input HT will have more
```
mike-w-wilson, line 298:
```suggestion
        Value should be greater than 1 so that input Table will have more
```

mike-w-wilson (commented):
I added a 10% default because I  figured it wouldnt hurt  and also removed the repartition language since this function isnt doing any repartitioning. The passed Table needs to be re-read with the intervals to repartition/for the join to be maximally efficient.

gnomad/utils/file_utils.py
mike-w-wilson: This needs to be workshopped...

mike-w-wilson: friday brain = ☁️

mike-w-wilson (approved):
LGTM

PR #513, opened 2023-02-16
Added pylint to pre-commit

ch-kr:
This PR adds pylint to the pre-commit config yaml. It relies on a local pylint installation per https://pylint.pycqa.org/en/latest/user_guide/installation/pre-commit-integration.html (which is why I've tagged this as a breaking change)


ch-kr:
production team discussed this and decided not to add this change

PR #522, opened 2023-03-27
Update some gnomAD resources from lists to version dictionaries

mike-w-wilson:
These resources were previously v3 specific. This update changes them to dictionaries where the key is the version and the value is a list.

### DIFF
--- a/gnomad/resources/grch38/gnomad.py
+++ b/gnomad/resources/grch38/gnomad.py
@@ -19,17 +19,20 @@ GENOME_COVERAGE_RELEASES = GENOME_RELEASES + ["3.0.1"]
 DATA_TYPES = ["genomes"]
 
 GENOME_POPS = ["AFR", "AMI", "AMR", "ASJ", "EAS", "FIN", "NFE", "SAS", "OTH"]
-SUBSETS = [
-    "non_v2",
-    "non_topmed",
-    "non_cancer",
-    "controls_and_biobanks",
-    "non_neuro",
-    "tgp",
-    "hgdp",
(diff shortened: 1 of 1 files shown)


gnomad/resources/grch38/gnomad.py
```diff
@@ -19,17 +19,20 @@
...
+    "v3": [
+        "non_v2",
+        "non_topmed",
+        "non_cancer",
+        "controls_and_biobanks",
+        "non_neuro",
+        "tgp",
+        "hgdp",
+    ],
+    "v4": ["ukb", "non-ukb", "non-topmed"],
```
jkgoodrich, line 32: Just confirming there is an intentional switch to `-` from `_` I don't remember all thee fine details associated with that.

jkgoodrich (approved):
Ugh, I hate that we need to add a breaking change here, but I see no reasonable way around it. 

Just one confirmation comment and then it's good to go. Thanks for adding.

gnomad/resources/grch38/gnomad.py
mike-w-wilson: Ah no good catch. The 4.3 vcf spec does not allow hyphens so underscores are better here to be consistent with the vcf. I took this straight from the v4 release HT schema but I'll update it there as well.

mike-w-wilson: https://atgu.slack.com/archives/CNL7NA6H2/p1604673884093000?thread_ts=1604652743.092300&cid=CNL7NA6H2

PR #523, opened 2023-03-28
Add additional groupings to optional stratified allele frequencies

KoalaQin:
Add additional groupings to optional stratified allele frequencies to allow stratified frequencies by multiple annotations: e.g. platform and population, or GATK version and population.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -391,12 +391,19 @@ def annotate_freq(
     :param pop_expr: When specified, frequencies are stratified by population. If `sex_expr` is also specified, then a pop/sex stratifiction is added.
     :param subpop_expr: When specified, frequencies are stratified by sub-continental population. Note that `pop_expr` is required as well when using this option.
     :param additional_strata_expr: When specified, frequencies are stratified by the given additional strata found in the dict. This can e.g. be used to stratify by platform.
+    :param additional_strata_grouping_expr: When specified, frequencies are further stratified by groups within the additional_strata_expr. This can e.g. be used to stratify by platform-population.
     :param downsamplings: When specified, frequencies are computed by downsampling the data to the number of samples given in the list. Note that if `pop_expr` is specified, downsamplings by population is also computed.
     :return: MatrixTable with `freq` annotation
     """
     if subpop_expr is not None and pop_expr is None:
         raise NotImplementedError(
             "annotate_freq requires pop_expr when using subpop_expr"
+        )
+
+    if additional_strata_grouping_expr is not None and additional_strata_expr is None:
+        raise NotImplementedError(
+            "annotate_freq requires additional_strata_expr when using"
+            " additional_strata_grouping_expr"
         )
 
     if additional_strata_expr is None:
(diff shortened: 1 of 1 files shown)


gnomad/utils/annotations.py
```diff
@@ -334,6 +334,9 @@ def annotate_freq(
     pop_expr: Optional[hl.expr.StringExpression] = None,
     subpop_expr: Optional[hl.expr.StringExpression] = None,
     additional_strata_expr: Optional[Dict[str, hl.expr.StringExpression]] = None,
+    additional_strata_grouping_expr: Optional[
+        Dict[str, hl.expr.StringExpression]
+    ] = None,
     downsamplings: Optional[List[int]] = None,
```
jkgoodrich, line 340: Will this break any of our other code that assumes ordering?

gnomad/utils/annotations.py
```diff
@@ -410,7 +422,10 @@ def annotate_freq(
     # Annotate cols with provided cuts
     mt = mt.annotate_cols(_freq_meta=_freq_meta_expr)
 
-    # Get counters for sex, pop and subpop if set
+    if additional_strata_grouping_expr is None:
+        additional_strata_grouping_expr = {}
```
jkgoodrich, line 426: could move up to if statement above so make above:

```
    if additional_strata_grouping_expr is None:
        additional_strata_grouping_expr = {}
    else:
        _freq_meta_expr = _freq_meta_expr.annotate(**additional_strata_grouping_expr)

```

gnomad/utils/annotations.py
```diff
@@ -509,6 +524,28 @@ def annotate_freq(
...
     )
 
+    # Add additional groupings to strata, e.g. strata-pop, strata-sex, strata-pop-sex
+    if additional_strata_grouping_expr is not None:
+        for strata in additional_strata_grouping_expr:
+            if strata not in cut_data.keys():
+                raise KeyError(
+                    "%s is not an existing annotation and thus cannot be combined with"
+                    " additional strata"
+                )
```
jkgoodrich, line 534: Will this ever happen? strata should be in cut_data right because it's added to `_freq_meta_expr` which is annotated as `mt._freq_meta` and that is used to build the cut_dict for the cut_data aggregation

jkgoodrich (approved):
Just a couple small questions, and this will work for now, but I agree with you that we should probably refactor the frequency code after this release to make this more flexible.

A couple small things that wont change functionality

gnomad/utils/annotations.py
mike-w-wilson: I checked gnomad_methods and gnomad_qc. gnomad_methods does not called annotate_freq. gnomad_qc/v3 does but only assumes ordering of `additional_strata_expr` when passing subsets, so this change will not impact that code. All other calls use keyword assignment.

gnomad/utils/annotations.py
mike-w-wilson: You're right it should not! This is an artifact of a previous approach, before we added `additional_strata_groupings_expr` to _freq_meta_expr- will remove this check.

PR #529, opened 2023-04-05
Fix access to version's SUBSETS and POPS within repo

mike-w-wilson:
SUBSETS, POPS, and DOWNSAMPLINGS were updated to dictionaries using the major version as a key and version-specific lists as values. The methods within the repo where not updated to use them properly. This fixes that. DOWNSAMPLINGS is not used within the repo. This was the simplest and clearest way I could think of using the dictionaries but open to other approaches!


jkgoodrich (approved):
LGTM, thanks for the fix

PR #530, opened 2023-04-13
Modify `annotate_mutation_type` to take optional context length as a parameter.

jkgoodrich:
The way context_length is currently determined, there needs to be a full scan through the HT/MT. This allows for a parameter to be passed in instead.

It also throws an error if it finds a 'context' value that is not the expected length. And if no parameter is passed it will detect it using the number of specified values in `num_scan_context_length`.


klaricch (approved):
lgtm

PR #531, opened 2023-04-19
Add `pab_max_expr` function and modify `default_compute_info` to add 'AS_pab_max' annotation

jkgoodrich:


mike-w-wilson (approved):
Just a couple small things then LGTM

PR #536, opened 2023-05-04
Add global annotations for 'vep_version', 'vep_help', and 'vep_config ' to the returned Table in `vep_or_lookup_vep`

jkgoodrich:


mike-w-wilson (approved):
👍

PR #541, opened 2023-05-11
Add VRS annotations fields dictionary

KoalaQin:
add a dictionary of VRS annotations, to be used in "gnomad_qc/v3/create_release/prepare_vcf_data_release.py"

### DIFF
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -328,6 +328,42 @@ IN_SILICO_ANNOTATIONS_INFO_DICT = {
 Dictionary with in silico score descriptions to include in the VCF INFO header.
 """
 
+
+VRS_FIELDS_DICT = {
+    "VRS_Allele_IDs": {
+        "Number": "1",
+        "Description": (
+            "The computed identifiers for the GA4GH VRS Alleles corresponding to the"
+            " values in the REF and ALT fields"
+        ),
+    },
+    "VRS_Starts": {
+        "Number": "1",
+        "Description": (
+            "Interresidue coordinates used as the location starts for the GA4GH VRS"
+            " Alleles corresponding to the values in the REF and ALT fields"
+        ),
+    },
+    "VRS_Ends": {
+        "Number": "1",
+        "Description": (
+            "Interresidue coordinates used as the location ends for the GA4GH VRS"
+            " Alleles corresponding to the values in the REF and ALT fields"
+        ),
+    },
+    "VRS_States": {
+        "Number": "1",
+        "Description": (
+            "The literal sequence states used for the GA4GH VRS Alleles corresponding"
+            " to the values in the REF and ALT fields"
+        ),
+    },
+}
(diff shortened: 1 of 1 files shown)


gnomad/utils/vcf.py
```diff
@@ -328,6 +328,42 @@
 Dictionary with in silico score descriptions to include in the VCF INFO header.
 """
 
+
+VRS_FIELDS_DICT = {
+    "VRS_Allele_IDs": {
+        "Number": "1",
```
klaricch, line 334:
```suggestion
        "Number": "R",
```

gnomad/utils/vcf.py
```diff
@@ -328,6 +328,42 @@
...
+        "Description": (
+            "The computed identifiers for the GA4GH VRS Alleles corresponding to the"
+            " values in the REF and ALT fields"
+        ),
+    },
+    "VRS_Starts": {
+        "Number": "1",
```
klaricch, line 341:
```suggestion
        "Number": "R",
```

gnomad/utils/vcf.py
```diff
@@ -328,6 +328,42 @@
...
+        "Description": (
+            "Interresidue coordinates used as the location starts for the GA4GH VRS"
+            " Alleles corresponding to the values in the REF and ALT fields"
+        ),
+    },
+    "VRS_Ends": {
+        "Number": "1",
```
klaricch, line 348:
```suggestion
        "Number": "R",
```

gnomad/utils/vcf.py
```diff
@@ -328,6 +328,42 @@
...
+        "Description": (
+            "Interresidue coordinates used as the location ends for the GA4GH VRS"
+            " Alleles corresponding to the values in the REF and ALT fields"
+        ),
+    },
+    "VRS_States": {
+        "Number": "1",
```
klaricch, line 355:
```suggestion
        "Number": "R",
```

klaricch (changes requested):

gnomad/utils/vcf.py
KoalaQin: Are you referring to this? https://samtools.github.io/hts-specs/VCFv4.2.pdf
For example, if the INFO field contains a single number, then this value should be 1; if the INFO field describes a pair of numbers, then this value should be 2 and so on. There are also certain special characters used to define special cases:
• If the field has one value per alternate allele then this value should be ‘A’.
• If the field has one value for each possible allele (including the reference), then this value should be ‘R’.
• If the field has one value for each possible genotype (more relevant to the FORMAT tags) then this value
should be ‘G’.
• If the number of possible values varies, is unknown, or is unbounded, then this value should be ‘.’.

KoalaQin: I got the point now, R = "respectively"?

klaricch (approved):
lgtm

PR #542, opened 2023-05-17
Add additional strata to freq dict labels

mike-w-wilson:
We built functionality to add additional strata to the freq array but did not include the ability to build those strata into the freq index dictionary. This seemed like the simplest way to do this. Additional strata are simply added to the end of the label `amr_xx_adj_4.1.0.1`. We historicaly have liked to see adj last so I did put a note that we could insert before the last elemnt in sort order. We could also add the functionality to insert the strata into the SORT_ORDER list at a certain index but this requires knowledge of the order and the order could change based on other groupings. For these reasons, I felt it was best (and easiest) to tack it onto the end of the order or before the last element.

### DIFF
--- a/gnomad/utils/release.py
+++ b/gnomad/utils/release.py
@@ -103,7 +103,9 @@ def make_freq_index_dict(
                 # Note: Tack the new strata onto the end of the sort order so the labels
                 # can be made
                 if k not in sort_order:
-                    sort_order.append(k)
+                    sort_order.append(
+                        k
+                    )  # TODO: Should we add the additional strata before group, aka adj? sort_order.insert(-1,k)
             index_dict.update({**_get_index(dict(group=groups, **strata))})
 
     return index_dict
--- a/gnomad/utils/release.py
+++ b/gnomad/utils/release.py
@@ -1,5 +1,5 @@
 # noqa: D100
-
+import copy
 from typing import Dict, List, Optional
 
 from gnomad.resources.grch38.gnomad import (
@@ -9,7 +9,7 @@ from gnomad.resources.grch38.gnomad import (
     SEXES,
     SUBSETS,
 )
-from gnomad.utils.vcf import index_globals
+from gnomad.utils.vcf import SORT_ORDER, index_globals
 
 
 def make_faf_index_dict(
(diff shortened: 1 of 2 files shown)


jkgoodrich (commented):
maybe this is a really stupid question, but can we change to just something like this: 

```
def make_freq_index_dict(
    freq_meta: List[Dict[str, str]],
    label_delimiter: str = "_",
    sort_order: Optional[List[str]] = SORT_ORDER,
) -> Dict[str, int]:
    index_dict = {}
    for i, f in enumerate(freq_meta):
        if sort_order and len(set(f.keys()) - set(sort_order)) < 1:
            index_dict[label_delimiter.join([f[g] for g in sorted(f.keys(), key=(lambda x: sort_order.index(x)) if sort_order else None)])] = i

    return index_dict
```

We could rename it something so that the existing function still exists as is, but I think this is way over complicated for what we need

gnomad/utils/release.py
```diff
@@ -91,4 +97,15 @@ def _get_index(label_groups):
...
 
+    if additional_strata:
+        for strata in additional_strata:
+            for k, v in strata.items():
+                # Note: Tack the new strata onto the end of the sort order so the labels
+                # can be made
+                if k not in sort_order:
+                    sort_order.append(
+                        k
+                    )  # TODO: Should we add the additional strata before group, aka adj? sort_order.insert(-1,k)
```
KoalaQin, line 108: I think the GATK version looks better after adj, number like 4.1.0.1 in the middle seems a bit weird.

KoalaQin (commented):
All the other parts look good, do you want to keep the second solution for now?

KoalaQin (changes requested):
I tested Julia's suggestion, but I think you need to put the addtional_strata in the SORT_ORDER now. Julia prefers adj/raw in the end of all the combos. 
Also, this code need a minor change: 
```
def make_freq_index_dict(
    freq_meta: List[Dict[str, str]],
    label_delimiter: str = "_",
    sort_order: Optional[List[str]] = SORT_ORDER,
) -> Dict[str, int]:
    index_dict = {}
    for i, f in enumerate(freq_meta):
        if sort_order and (len(f.keys()) - len(sort_order)) < 1:
            index_dict[label_delimiter.join([f[g] for g in sorted(f.keys(), key=(lambda x: sort_order.index(x)) if sort_order else None)])] = i

    return index_dict
```

gnomad/utils/vcf.py
```diff
@@ -517,7 +517,7 @@ def make_label_combos(
     combos = []
     for x, y in itertools.product(
         anchor_val,
-        make_label_combos(copy_label_groups, label_delimiter=label_delimiter),
+        make_label_combos(copy_label_groups, sort_order, label_delimiter),
```
KoalaQin, line 520: Do you need to add additional strata like "gatk_version" in SORT_ORDER

KoalaQin (changes requested):

PR #545, opened 2023-05-25
Allow any frequency groupings

mike-w-wilson:
This PR expands on the utility of `additional_strata_expr` by allowing the user to pass a list of dictionaries where each dictionary will be grouped together. This relies on the user to define the specific groupings for freq calculations, e.g. [{'pop': pop_expr}, {'platform': platform_expr, 'pop': pop_expr}, {'sample_age_bin': sample_age_bin_expr}]. I think this could eventually be used for all aggregations and clean this function up quite a bit.  I kept the ability to pass a single dictionary but have removed the additonal_strata_groupings_expr as this is a cleaner and more flexible implementation.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -397,8 +397,8 @@ def annotate_freq(
 
     If the `additional_strata_expr` parameter is used, frequencies will be computed for each of the strata dictionaries across all
     values. For example, if `additional_strata_expr` is set to `[{'platform': mt.platform}, {'platform':mt.platform, 'pop': mt.pop},
-     {'age_bin': mt.age_bin}]`, then frequencies will be computed for each of the values of `mt.platform`, each of the combined values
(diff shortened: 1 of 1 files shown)


KoalaQin (changes requested):

mike-w-wilson:
@KoalaQin , wI see you marked the review as changes requested but what changes would you like to see?

KoalaQin:
@mike-w-wilson Hi, Mike, sorry, I wasn't clear, I was thinking maybe you can change the doc for "additional_strata_expr", because you used `pop` and `platform` in your example, but we are going to do this on `gatk_version` and `sample_age_bin` this time, just for my testing convenience.

mike-w-wilson:
Since this is the gnomad_methods repo, we want the functions to be as general as possible. Pop and platform are something we mention in the repo so I figured it makes sense to use this as an example where gatk_version is not. All of that logic, around gatk_version will live in the gnomad_qc repo.

KoalaQin (approved):

PR #546, opened 2023-05-26
add VEP 105 CSQ FIELDs

KoalaQin:
make VEP_CSQ_FIELDS a dictionary to contain both 101 & 105 fields, but make VEP_CSV_HEADER default to 105. Double checked vep_struct_to_csq(), no exceptions need to be added. 
remove the missing fields: MINIMISED, SWISSPROT|TREMBL|UNIPARC from VEP 105 CSQ fields. ancestal & context were not in the CSQ fields of VEP 101.


KoalaQin:
@mike-w-wilson I wasn't sure if I should also change [line 391](https://github.com/broadinstitute/gnomad_methods/blob/f9c2f0ffb10332002730d271a842ec8dbec65fb6/gnomad/utils/vep.py#L391). My intension was to keep both 101 & 105 fields.

mike-w-wilson (changes requested):

mike-w-wilson:
Yes, please change the linked line 391 VEP_CSQ_FIELDS. You can have it default to `CURRENT_VEP_VERSION`

KoalaQin:
@mike-w-wilson, I added VEP_VERSIONS & CURRENT_VEP_VERSION.

gnomad/utils/vep.py
```diff
@@ -388,7 +400,8 @@ def filter_vep_to_synonymous_variants(
 
 
 def vep_struct_to_csq(
-    vep_expr: hl.expr.StructExpression, csq_fields: str = VEP_CSQ_FIELDS
+    vep_expr: hl.expr.StructExpression,
+    csq_fields: str = VEP_CSQ_FIELDS[CURRENT_VEP_VERSION],
```
mike-w-wilson, line 404: I can't add it on L419, but can you add something like `Default is the CSQ fields of the CURRENT_VEP_VERSION.`

mike-w-wilson (commented):
One more thing and then this looks good!

gnomad/utils/vep.py
KoalaQin: I did, sorry, I will keep in mind changing the docs every time I change the function.

mike-w-wilson (approved):
LGTM!

PR #547, opened 2023-05-26
Add callstats option to inbreeding coeff

mike-w-wilson:
Adds the option to pass a callstats struct expression with AC, AN, and the number of homozygote alternates to the inbreeding coefficient expr, allowing someone without access to the GT calls to still calculate inbreeding coefficient.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -125,7 +125,7 @@ def project_max_expr(
                     > 0
                 ),
                 # order the callstats computed by AF in decreasing order
-                lambda x: -x[1].AF[ai]
+                lambda x: -x[1].AF[ai],
                 # take the n_projects projects with largest AF
             )[:n_projects].map(
                 # add the project in the callstats struct
@@ -219,8 +219,10 @@ def faf_expr(
                 ~locus.in_autosome_or_par(),
                 hl.struct(
                     **{
-                        f"faf{str(threshold)[2:]}": hl.experimental.filtering_allele_frequency(
-                            freq[i].AC, freq[i].AN, threshold
+                        f"faf{str(threshold)[2:]}": (
+                            hl.experimental.filtering_allele_frequency(
+                                freq[i].AC, freq[i].AN, threshold
+                            )
                         )
                         for threshold in faf_thresholds
                     }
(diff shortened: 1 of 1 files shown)


KoalaQin (commented):
I tried to see if it's a switch from het to homo-non-ref, but we only change GT if the variant AF>0.1, right? Could you show me your test run?

mike-w-wilson:
You can test both options on this small dense MT from v4 here: "gs://gnomad-mwilson/v4/frequencies/test_on_gene.mt". The results should match up. 
```
mt = hl.read_matrix_table("gs://gnomad-mwilson/v4/frequencies/test_on_gene.mt")
logger.info("Calculating InbreedingCoeff...")
        # NOTE: This is not the ideal location to calculate this, but added here to avoid another densify # noqa
mt = mt.annotate_rows(
    InbreedingCoeff=bi_allelic_site_inbreeding_expr(mt.GT)
)
ht = mt.rows()
ht = freq_ht.annotate(callstats_InbreedingCoefficient=bi_allelic_site_inbreeding_expr(callstats_expr=freq_ht.freq[1]))

File: gnomad/utils/annotations.py [hunk 1]
```diff
@@ -906,7 +913,24 @@ def inbreeding_coeff(
         q = (2 * gt_counts.get(2, 0) + gt_counts.get(1, 0)) / (2 * n)
         return 1 - (gt_counts.get(1, 0) / (2 * p * q * n))
 
-    return hl.bind(inbreeding_coeff, hl.agg.counter(call.n_alt_alleles()))
+    if callstats_expr is not None:
```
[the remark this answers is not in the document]
mike-w-wilson: Decided test should be on v4 GT vs v4 calsltats as we are not certain what went  into v3's inbreeding coeff generation

mike-w-wilson:
@KoalaQin , this is back to you -- I've updated the function to what was suggested at code review.

gnomad/utils/annotations.py
```diff
@@ -906,7 +912,34 @@ def inbreeding_coeff(
...
+                (callstats_expr.AN.dtype == hl.tint32)
+                | (callstats_expr.AN.dtype == hl.tint64)
+            )
+            & (
+                (callstats_expr.homozygote_count.dtype == hl.tint32)
+                | (callstats_expr.homozygote_count.dtype == hl.tint64)
+            )
+        ):
+            raise ValueError(
+                "callstats_expr must be a StructExpression containing int32 fields AC,"
```
KoalaQin, line 932: you may change to "containing int32 or int64"

KoalaQin (changes requested):

gnomad/utils/annotations.py
mike-w-wilson: Good catch!

mike-w-wilson:
Back to you @KoalaQin !

PR #564, opened 2023-07-18
Add `sample_list_stratification` option to `create_fake_pedigree` function

jkgoodrich:
`sample_list_stratification` is an optional dictionary with samples as keys and a value that should be used to stratify samples in `sample_list` into groups that the trio should be picked from. This ensures that each fake trio will contain samples from only the same stratification. For example, if all samples within a fake trio should be chosen from the same platform, this can be a dictionary of sample: platform.


klaricch (approved):
lgtm

PR #570, opened 2023-08-02
Function to annotate downsamplings onto HT/MT

mike-w-wilson:
This adds a function to annotate downsamplings that can be used in the frequency calculation.


jkgoodrich (approved):
LGTM!

PR #537, opened 2023-05-05
Clone and modify annotate_freq for faster frequency and het annotation

tpoterba:
Directly returns a table with freq and het info, used like this:
```
ht = annotate_freq_and_high_ab_hets(
    mt,
    sex_expr=mt.meta.sex_imputation.sex_karyotype,
    pop_expr=mt.meta.population_inference.pop,
    downsamplings=DOWNSAMPLINGS["v4"],
    additional_strata_expr={"gatk_version": mt.meta.project_meta.gatk_version},
    additional_strata_grouping_expr={"pop": mt.meta.population_inference.pop},
    ab_cutoff=0.9
)
ht.write('/tmp/freq.ht', overwrite=True)
```

On my laptop I'm seeing about 11 variants/s/core. This should go pretty quickly on the full dataset.

PR #571, opened 2023-08-03
Update annotate_freq and qual_hists, add split_vds and compute_freq_by_strata

mike-w-wilson:
This updates annotate_freq to use the added array aggregation functionality in hail, originally added by Tim in #537, adds the ability to add entry aggregation annotations, and generally cleans up the function by splitting the existing annotate_freq into two functions: annotate_freq which calls compute_freq_by_strata. 

This also adds split_vds_by_strata which returns a lists of VDSs with as many VDSs as there are unique values for the passed expression. 

We also add the ability to return a single struct of qual histograms, containing separate structs of raw and adj qual histograms.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -359,8 +359,8 @@ def annotate_freq(
         ]
     ] = None,
     downsamplings: Optional[List[int]] = None,
-    ds_pop_counts: Optional[Dict[str, int]] = None,
     downsampling_expr: Optional[hl.expr.StructExpression] = None,
+    ds_pop_counts: Optional[Dict[str, int]] = None,
     entry_agg_funcs: Optional[Dict[str, Tuple[Callable, Callable]]] = None,
     annotate_mt: bool = True,
 ) -> Union[hl.Table, hl.MatrixTable]:
@@ -410,14 +410,17 @@ def annotate_freq(
 
     .. rubric:: The `downsamplings` parameter
 
-    If the `downsamplings` parameter is used, frequencies will be computed for all
-    samples and by population (if `pop_expr` is specified) by downsampling the number
-    of samples without replacement to each of the numbers specified in the
-    `downsamplings` array, provided that there are enough samples in the dataset. In
-    addition, if `pop_expr` is specified, a downsampling to each of the exact number
-    of samples present in each population is added. Note that samples are randomly
-    sampled only once, meaning that the lower downsamplings are subsets of the higher
-    ones.
+    If the `downsamplings` parameter is used without the `downsampling_expr`,
+    frequencies will be computed for all samples and by population (if `pop_expr` is
+    specified) by downsampling the number of samples without replacement to each of the
+    numbers specified in the `downsamplings` array, provided that there are enough
+    samples in the dataset. In addition, if `pop_expr` is specified, a downsampling to
+    each of the exact number of samples present in each population is added. Note that
+    samples are randomly sampled only once, meaning that the lower downsamplings are
+    subsets of the higher ones. If the `downsampling_expr` parameter is used with the
+    `downsamplings` parameter, the `downsamplings` parameter informs the function which
+    downsampling groups were already created and are to be used in the frequency
+    calculation.
 
     .. rubric:: The `additional_strata_expr` parameter
 
@@ -428,6 +431,25 @@ def annotate_freq(
     frequencies will be computed for each of the values of `mt.platform`, each of the
     combined values of `mt.platform` and `mt.pop`, and each of the values of
     `mt.age_bin`.
+
+    .. rubric:: The `downsampling_expr` and `ds_pop_counts` parameters
+
+    If the `downsampling_expr` parameter is used, `downsamplings` must also be set
+    and frequencies will be computed for all samples and by population (if `pop_expr`
+    is specified) using the downsampling indices to each of the numbers specified in
+    the `downsamplings` array. The function `annotate_downsamplings` can be used to to
+    create the `downsampling_expr`, `downsamplings`, and `ds_pop_counts` expressions.
+
+    .. rubric:: The `entry_agg_funcs` parameter
+
+    If the `entry_agg_funcs` parameter is used, the output MatrixTable will also
+    contain the annotations specified in the `entry_agg_funcs` parameter. The keys of
+    the dict are the names of the annotations and the values are tuples of functions.
+    The first function is used to transform the `mt` entries in some way, and the
+    second function is used to aggregate the output from the first function. For
+    example, if `entry_agg_funcs` is set to {'adj_samples': (get_adj_expr, hl.agg.sum)}`,
+    then the output MatrixTable will contain an annotation `adj_samples` which is an
+    array the of the number of adj samples per strata in each row.
 
     :param mt: Input MatrixTable
     :param sex_expr: When specified, frequencies are stratified by sex. If `pop_expr`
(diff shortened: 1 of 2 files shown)


gnomad/utils/annotations.py
```diff
@@ -1446,3 +1373,166 @@ def annotate_downsamplings(
...
+    strata_expr: List[Dict[str, hl.expr.StringExpression]],
+    downsamplings: Optional[List[int]] = None,
+    ds_pop_counts: Optional[Dict[str, int]] = None,
+    entry_agg_funcs: Optional[Dict[str, Tuple[Callable, Callable]]] = None,
+) -> hl.Table:
+    """
+    Compute call statistics and, when passed, entry aggregation function(s) by strata.
+
+    The computed call statistics are AC, AF, AN, and homozygote_count. Downsamplings are
+    added to the strata when downsamplings when passed. The entry aggregation functions
```
jkgoodrich, line 1389: I think somewhere, either here or in the other function we need to address the expectations for `pop_idx` and `global_idx` in `downsampling_expr` or any value in `strata_expr` with 'downsampling'  and it should reference `annotate_downsamplings`

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1373,166 @@ def annotate_downsamplings(
...
+    :param downsamplings: Optional list of downsampling groups.
+    :param ds_pop_counts: Optional dict of population counts for downsampling groups.
+    :param entry_agg_funcs: Optional dict of entry aggregation functions. When
+        specified, additional annotations are added to the output Table/MatrixTable.
+        The keys of the dict are the names of the annotations and the values are tuples
+        of functions. The first function is used to transform the `mt` entries in some
+        way, and the second function is used to aggregate the output from the first
+        function.
+    :return: Table or MatrixTable with allele frequencies by strata.
+    """
```
jkgoodrich, line 1407: Do you think we need to add errors like we have in the other function? We could also consider adding a `_` to this function name indicating that it's mostly intended to be used internally

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1373,166 @@ def annotate_downsamplings(
...
+            }
+        )
+    )
+
+    # Add all desired strata to sample group filters.
+    sample_group_filters = [({}, True)]
+    for strata in strata_expr:
+        downsampling_expr = strata.get("downsampling")
+        strata_values = []
+        for s in strata:
```
jkgoodrich, line 1426: We probably want a comment above here

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1373,166 @@ def annotate_downsamplings(
...
+                v = [("downsampling", d) for d in downsamplings]
+            else:
+                v = [(s, k[s]) for k in strata_counts.get(s, {})]
+                if s == "pop" and downsampling_expr is not None:
+                    v.append(("pop", "global"))
+            strata_values.append(v)
+
+        # Get all combinations of strata values.
+        strata_combinations = itertools.product(*strata_values)
+        for combo in strata_combinations:
```
jkgoodrich, line 1437: Probably want a comment above this for loop

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1373,166 @@ def annotate_downsamplings(
...
+
+    # Annotate columns with group_membership.
+    mt = mt.annotate_cols(group_membership=[x[1] for x in sample_group_filters])
+
+    # Get sample count per strata group.
+    freq_sample_count = mt.aggregate_cols(
+        hl.agg.array_agg(lambda x: hl.agg.count_where(x), mt.group_membership)
+    )
+
+    # Create and annotate global expression with meta and sample count information
```
jkgoodrich, line 1474:
```suggestion
    # Create and annotate global expression with meta and sample count information.
```

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1373,166 @@ def annotate_downsamplings(
...
+                lambda s_i: ht.cols[s_i].group_membership[g_i]
+            )
+        )
+    )
+    ht = ht.annotate(
+        adj_array=ht.entries.map(lambda e: e.adj),
+        gt_array=ht.entries.map(lambda e: e.GT),
+    )
+
+    def _agg_by_group(ht, agg_func, agg_expr, *args):
```
jkgoodrich, line 1500: Add types and docstring

jkgoodrich (commented):
Just some areas that documentation needs to be added

gnomad/utils/annotations.py
mike-w-wilson: I added errors but if you think they are not enough coverage, we can mark it as a private function but I would like to keep it open I think

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1401,214 @@ def annotate_downsamplings(
...
+        )
+    )
+
+    # Add all desired strata to sample group filters.
+    sample_group_filters = [({}, True)]
+    for strata in strata_expr:
+        downsampling_expr = strata.get("downsampling")
+        strata_values = []
+        # Add to all downsampling groups, both global and population-specific, to
+        # strata
```
jkgoodrich, line 1497:
```suggestion
        # strata.
```

gnomad/utils/annotations.py
jkgoodrich: pinging in case this comment was missed from the last round

jkgoodrich (commented):
Just a small request

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1401,214 @@ def annotate_downsamplings(
...
+    if downsamplings is not None and not ds_in_strata:
+        errors.append(
+            "Strata must contain a downsampling expression when downsamplings"
+            "are provided."
+        )
+    if downsamplings is not None and not ds_in_strata:
+        errors.append(
+            "Strata must contain a downsampling expression when downsamplings"
+            "are provided."
+        )
```
jkgoodrich, line 1458: I think this is a duplicate

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1401,220 @@ def annotate_downsamplings(
...
+    ht = ht.annotate(
+        adj_array=ht.entries.map(lambda e: e.adj),
+        gt_array=ht.entries.map(lambda e: e.GT),
+    )
+
+    def _agg_by_group(
+        ht: hl.Table, agg_func: Callable, agg_expr: Callable, *args
+    ) -> hl.expr.ArrayExpression:
+        """
+        Run a passed function on entries and aggregate the results by group.
```
jkgoodrich, line 1575:
```suggestion
        Aggregate `agg_expr` by group using the `agg_func` function.
```

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1401,220 @@ def annotate_downsamplings(
...
+
+        :param ht: Input Hail Table.
+        :param agg_func: Function to apply to the entries.
```
jkgoodrich, line 1578:
```suggestion
        :param agg_func: Aggregation function to apply to `agg_expr`.
```

gnomad/utils/annotations.py
```diff
@@ -1446,3 +1401,220 @@ def annotate_downsamplings(
...
+        :param agg_expr: Aggregation to run on the result of `agg_func`.
```
jkgoodrich, line 1579:
```suggestion
        :param agg_expr: Expression to aggregate by group.
```

jkgoodrich (changes requested):
Small doc string changes

jkgoodrich (approved):
LGTM!

PR #576, opened 2023-08-13
Avoid DeprecationWarnings from superseded hail function and import [minor] 

jmarshall:
Rewrite remaining invocations of `hl.cond()` to `hl.if_else()`. There are already numerous instances of `hl.if_else()` in gnomad_methods, so this has no compatibility considerations.

Also (in _gnomad/utils/file_utils.py_) import `GoogleStorageAsyncFS` from its modern location. This has been `hailtop.aiocloud.aiogoogle` since hail 2.0.78 (see hail-is/hail#10898 and hail-is/hail#10974). If gnomad_methods needs to be compatible with versions of hail prior to 2.0.78, this part of the change should be omitted.

Very minor, but a few of these appear in our test logs which makes for noise when tracking down other problems via the logs.

### DIFF
--- a/gnomad/sample_qc/relatedness.py
+++ b/gnomad/sample_qc/relatedness.py
@@ -1097,7 +1097,7 @@ def generate_trio_stats_expr(
                 locus.in_autosome(),
                 proband_gt.is_het() & father_gt.is_hom_ref() & mother_gt.is_hom_ref(),
             )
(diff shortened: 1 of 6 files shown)


KoalaQin (approved):
Are we changing this in `gnomad_qc` too? Or keep hl.cond in older releases?

PR #579, opened 2023-08-17
Run count merge through dict of lists rather than single list

mike-w-wilson:
This is a breaking change but this is so new that I doubt it's being used outside of our group. This just allows more than a single count array annotation to be merged at the same time. I'll use it to merge the freq meta sample counts. We could add the option to just pass a single list, so union on count arrays and convert it to a dict if you think that would be better.


jkgoodrich (approved):
LGTM

PR #580, opened 2023-08-18
function to update sample annotations

KoalaQin:
Update highly structured annotations such as gnomAD sample meta.


KoalaQin:
@jkgoodrich I get this error: 
ImportError: cannot import name 'add_filters_expr' from partially initialized module 'gnomad.utils.filtering' (most likely due to a circular import)
I searched around, it seems that I can't import functions from filtering.py because filtering.py is importing something from annotations.py
Where should I put the update_sample_annotations?

KoalaQin:
I put this function in release.py, passed checks, is it Okay?

gnomad/utils/release.py
```diff
@@ -151,3 +152,31 @@ def make_freq_index_dict_from_meta(
             ] = i
 
     return index_dict
+
+
+def update_sample_annotations(expr: hl.expr, sample_annotations: Dict[str, hl.expr]):
```
jkgoodrich, line 157: I think this should go in annotations, it makes a lot more sense there. You can handle the circular import using `import as` for example `import gnomad.utils.filtering as filter_utils` 

Also, I modified it a little:

```
def update_structured_annotations(
    ht: hl.Table,
    annotation_update_exprs: Dict[str, hl.Expression],
    annotation_update_label: Optional[str] = None,
) -> hl.Table:
    """
    Update highly structured annotations on a Table.

    This function recursively updates annotations defined by `annotation_update_exprs`
    and if `annotation_update_label` is supplied, it checks if the sample annotations
    are different from the input and adds a flag to the Table, indicating which
    annotations have been updated for each sample.

    :param ht: Input Table with structured annotations to update.
    :param annotation_update_exprs: Dictionary of annotations to update, structured as
        they are structured on the input `ht`.
    :param annotation_update_label: Optional string of the label to use for an
        annotation indicating which annotations have been updated. Default is None, so
        no annotation is added.
    :return: Table with updated annotations and optionally a flag indicating which
        annotations were changed.
    """
    def _update_struct(
        struct_expr: hl.expr.StructExpression,
        update_exprs: Union[Dict[str, hl.Expression], hl.expr.Expression]
    ) -> Tuple[Dict[str, hl.expr.BooleanExpression], hl.expr.StructExpression]:
        """
        Update a StructExpression.

        :param struct_expr: StructExpression to update.
        :param update_exprs: Dictionary of annotations to update.
        :return: Tuple of the updated annotations and the updated flag.
        """
        if isinstance(update_exprs, dict):
            updated_struct_expr = {}
            updated_flag_expr = {}
            for ann, expr in update_exprs.items():
                updated_flag, updated_ann = _update_struct(struct_expr[ann], expr)
                updated_flag_expr.update(
                    {ann + ("." + k if k else ""): v for k, v in updated_flag.items()}
                )
                updated_struct_expr[ann] = updated_ann
            return updated_flag_expr, struct_expr.annotate(**updated_struct_expr)
        else:
            return {"": update_exprs != struct_expr}, update_exprs

    annotation_update_flag, updated_rows = _update_struct(
        ht.row_value, annotation_update_exprs
    )
    if annotation_update_label is not None:
        updated_rows = updated_rows.annotate(
            **{
                annotation_update_label: filter_utils.add_filters_expr(
                    filters=annotation_update_flag
                )
            }
        )

    return ht.annotate(**updated_rows)
```

jkgoodrich (changes requested):

gnomad/utils/release.py
KoalaQin: Okay, I put it back to annotations.py, and modified the return type in _updata_struct().

gnomad/utils/filtering.py
```diff
@@ -217,7 +217,7 @@ def filter_to_autosomes(
 def add_filters_expr(
     filters: Dict[str, hl.expr.BooleanExpression],
     current_filters: hl.expr.SetExpression = None,
-) -> hl.expr.SetExpression:
+) -> hl.expr:
```
jkgoodrich, line 220: Why is this being changed?

gnomad/utils/annotations.py
```diff
@@ -1795,3 +1796,65 @@ def _agg_by_group(
...
+    :param annotation_update_label: Optional string of the label to use for an
+        annotation indicating which annotations have been updated. Default is None, so
+        no annotation is added.
+    :return: Table with updated annotations and optionally a flag indicating which
+        annotations were changed.
+    """
+
+    def _update_struct(
+        struct_expr: hl.expr.StructExpression,
+        update_exprs: Union[Dict[str, hl.Expression], hl.expr.Expression],
```
jkgoodrich, line 1826:
```suggestion
        update_exprs: Union[Dict[str, hl.expr.Expression], hl.expr.Expression],
```

gnomad/utils/annotations.py
```diff
@@ -1795,3 +1796,65 @@ def _agg_by_group(
...
+    ) -> Tuple[Dict[str, Any], hl.expr]:
```
jkgoodrich, line 1827: is the first one not always `Dict[str, hl.expr.BooleanExpression]`? and why `hl.expr`?

jkgoodrich (commented):
a couple questions about types

gnomad/utils/filtering.py
KoalaQin: My PyCharm had warnings about types.

gnomad/utils/annotations.py
KoalaQin: I had this warning for the return in else: 
```
Expected type 'tuple[dict[str, BooleanExpression], StructExpression]', got 'tuple[dict[str, Any], Expression]' instead: 1846. 
```

gnomad/utils/filtering.py
KoalaQin: Expected type 'SetExpression', got 'Expression' instead: 236. Do you have this too? Or we can just ignore?

gnomad/utils/annotations.py
jkgoodrich: Try this `Tuple[Dict[str, hl.expr.BooleanExpression], Any]` I think the first should be a Boolean, the suppose the second can be multiple different things

gnomad/utils/filtering.py
jkgoodrich: I get that too, but it's just because the return type of hl.fold is the more general `hl.expr.Expression` since it can also be used on other expression types. However, `current_filters` is `hl.expr.SetExpression` and the output will only ever be `hl.expr.SetExpression`. For example, it is the same as below, which won't have that warning, but they return the same thing:

```
    if current_filters is None:
        current_filters = hl.empty_set(hl.tstr)

    for filter_name, filter_condition in filters.items():
        current_filters = current_filters.union(
            hl.if_else(filter_condition, hl.set([filter_name]), hl.empty_set(hl.tstr))
        )

    return current_filters
```

so I think it's fine as is

jkgoodrich (approved):
LGTM!

PR #583, opened 2023-08-25
Split vds change from list to dict

mike-w-wilson:
This changes the split_vds_by_strata's return to a dictionary instead of a list of VDSs. This gives more utility in seeing which VDS belongs to which strata. This also fixes a bug in the annotate_freq function.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -1339,7 +1339,7 @@ def annotate_freq(
         ds_ht = annotate_downsamplings(mt, downsamplings, pop_expr=pop_expr).cols()
         downsamplings = hl.eval(ds_ht.downsamplings)
         ds_pop_counts = hl.eval(ds_ht.ds_pop_counts)
-        downsampling_expr = ds_ht[mt.col_key]
+        downsampling_expr = ds_ht[mt.col_key].downsampling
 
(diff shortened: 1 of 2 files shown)


gnomad/utils/annotations.py
```diff
@@ -1339,7 +1339,7 @@ def annotate_freq(
         ds_ht = annotate_downsamplings(mt, downsamplings, pop_expr=pop_expr).cols()
         downsamplings = hl.eval(ds_ht.downsamplings)
         ds_pop_counts = hl.eval(ds_ht.ds_pop_counts)
-        downsampling_expr = ds_ht[mt.col_key]
+        downsampling_expr = ds_ht[mt.col_key].downsampling
```
jkgoodrich, line 1342: Thank you for catching!

gnomad/utils/filtering.py
```diff
@@ -515,17 +515,19 @@ def filter_for_mu(
 
 def split_vds_by_strata(
     vds: hl.vds.VariantDataset, strata_expr: hl.expr.Expression
-) -> List[hl.vds.VariantDataset]:
+) -> Dict[str, hl.vds.VariantDataset]:
```
jkgoodrich, line 518: Nice, that makes a lot of sense

jkgoodrich (approved):
LGTM!

PR #585, opened 2023-08-29
Add optional filter to freq meta sample counts

mike-w-wilson:
This adds in an optional filtering of another array that is based on freq_meta (filtering freq_meta_sample_counts is how I will use it)  but I think we could generalize this more since its just adding an optional step with the same exact filtering that happens on freq. Maybe make the function `filter_array_by_meta` or something and pass a dictionary of strings and array expressions that all get filtered the same.....all that said I didnt want to change too much without your inputs since this just went in.

### DIFF
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -600,12 +600,15 @@ def filter_freq_by_meta(
         ),
     )
     freq_expr = freq_meta_expr.map(lambda x: freq_expr[x[0]])
-    freq_meta_expr = freq_meta_expr.map(lambda x: x[1])
 
     if freq_meta_sample_count_expr is not None:
         freq_meta_sample_count_expr = freq_meta_expr.map(
             lambda x: freq_meta_sample_count_expr[x[0]]
         )
+
+    freq_meta_expr = freq_meta_expr.map(lambda x: x[1])
+
+    if freq_meta_sample_count_expr is not None:
         return freq_expr, freq_meta_expr, freq_meta_sample_count_expr
     else:
         return freq_expr, freq_meta_expr
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -539,6 +539,7 @@ def filter_freq_by_meta(
     items_to_filter: Union[Dict[str, List[str]], List[str]],
     keep: bool = True,
     combine_operator: str = "and",
+    freq_meta_sample_count_expr: Optional[hl.ArrayExpression] = None,
 ) -> Tuple[hl.expr.ArrayExpression, hl.expr.ArrayExpression]:
     """
     Filter frequency and frequency meta expressions specified by `items_to_filter`.
@@ -564,6 +565,7 @@ def filter_freq_by_meta(
     :param keep: Whether to keep or remove the items specified by `items_to_filter`.
     :param combine_operator: Whether to use "and" or "or" to combine the items
         specified by `items_to_filter`.
+    :param freq_meta_sample_count: Optional frequency meta sample count expression to be filtered.
     :return: Tuple of the filtered frequency and frequency meta expressions.
     """
     freq_meta_expr = freq_meta_expr.collect(_localize=False)[0]
@@ -600,4 +602,10 @@ def filter_freq_by_meta(
     freq_expr = freq_meta_expr.map(lambda x: freq_expr[x[0]])
     freq_meta_expr = freq_meta_expr.map(lambda x: x[1])
 
-    return freq_expr, freq_meta_expr
+    if freq_meta_sample_count_expr is not None:
+        freq_meta_sample_count_expr = freq_meta_expr.map(
+            lambda x: freq_meta_sample_count_expr[x[0]]
+        )
+        return freq_expr, freq_meta_expr, freq_meta_sample_count_expr
+    else:
+        return freq_expr, freq_meta_expr
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -552,7 +552,7 @@ def filter_arrays_by_meta(
 
     The items can be kept or removed from `meta_indexed_expr` and `meta_expr` based on
     the value of `keep`. For example if `meta_indexed_exprs` is {'freq': ht.freq,
-    'freq_meta_sample_count`: ht.index_globals().freq_meta_sample_count} and `meta_expr`
+'freq_meta_sample_count': ht.index_globals().freq_meta_sample_count} and `meta_expr`
     is ht.freq_meta then if `keep` is True, the items specified by `items_to_filter`
     such as  'pop' = 'han' will be kept and all other items will be removed from the
     ht.freq, ht.freq_meta_sample_count, and ht.freq_meta.
@@ -570,7 +570,6 @@ def filter_arrays_by_meta(
     :param keep: Whether to keep or remove the items specified by `items_to_filter`.
     :param combine_operator: Whether to use "and" or "or" to combine the items
         specified by `items_to_filter`.
-    :param meta_based_array_expr: Optional array based on freq meta expression to be filtered.
     :return: Tuple of the filtered metadata expression and a dictionary of metadata indexed expressions.
     """
     meta_expr = meta_expr.collect(_localize=False)[0]
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -541,7 +541,7 @@ def filter_arrays_by_meta(
     items_to_filter: Union[Dict[str, List[str]], List[str]],
     keep: bool = True,
     combine_operator: str = "and",
-) -> Tuple[hl.expr.ArrayExpression, Dict[str, hl.expr.ArrayExpression]]:
+) -> Tuple[hl.expr.ArrayExpression, Union[Dict[str, hl.expr.ArrayExpression], hl.expr.ArrayExpression]]:
     """
     Filter both metadata array expression and meta data indexed expression by `items_to_filter`.
 
--- a/gnomad/utils/filtering.py
+++ b/gnomad/utils/filtering.py
@@ -605,9 +605,9 @@ def filter_arrays_by_meta(
         ),
     )
 
-    meta_indexed_exprs = meta_indexed_exprs.map_values(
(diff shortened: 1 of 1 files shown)


gnomad/utils/filtering.py
```diff
@@ -564,6 +565,7 @@ def filter_freq_by_meta(
     :param keep: Whether to keep or remove the items specified by `items_to_filter`.
     :param combine_operator: Whether to use "and" or "or" to combine the items
         specified by `items_to_filter`.
+    :param freq_meta_based_array_expr: Optional array based on freq meta expression to be filtered.
     :return: Tuple of the filtered frequency and frequency meta expressions.
```
KoalaQin, line 569: You might need to change return, I got warnings from my Pycharm.

KoalaQin (changes requested):

gnomad/utils/filtering.py
```diff
@@ -539,6 +539,7 @@ def filter_freq_by_meta(
     items_to_filter: Union[Dict[str, List[str]], List[str]],
     keep: bool = True,
     combine_operator: str = "and",
+    freq_meta_based_array_expr: Optional[hl.ArrayExpression] = None,
```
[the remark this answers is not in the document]
mike-w-wilson: It's common practice to leave a trailing comma to reduce git diffs should we add more agrs.

gnomad/utils/filtering.py
mike-w-wilson: Good catch!

mike-w-wilson:
I decided to generalize this a bit more. No worries if you dont like it but this cleans up the args a bit from my previous commit and now you just pass a single dict with arrays that are indexed by the metadata expression, e.g. freq_meta. This is needs for freq, freq_meta_sample_count, and the high_ab_hets_by_group annotations in v4.

gnomad/utils/filtering.py
```diff
@@ -533,40 +533,47 @@ def split_vds_by_strata(
...
     - By a list of keys, e.g. ["sex", "downsampling"].
     - By specific key: value pairs, e.g. to filter where 'pop' is 'han' or 'papuan'
     {"pop": ["han", "papuan"]}, or where 'pop' is 'afr' and/or 'sex' is 'XX'
     {"pop": ["afr"], "sex": ["XX"]}.
 
-    The items can be kept or removed from `freq_expr` and `freq_meta_expr` based on the
-    value of `keep`.
+    The items can be kept or removed from `meta_indexed_expr` and `meta_expr` based on
+    the value of `keep`. For example if `meta_indexed_exprs` is {'freq': ht.freq,
+    'freq_meta_sample_count`: ht.index_globals().freq_meta_sample_count} and `meta_expr`
```
KoalaQin, line 555:
```suggestion
'freq_meta_sample_count': ht.index_globals().freq_meta_sample_count} and `meta_expr`
```

gnomad/utils/filtering.py
```diff
@@ -533,40 +533,47 @@ def split_vds_by_strata(
...
+    is ht.freq_meta then if `keep` is True, the items specified by `items_to_filter`
+    such as  'pop' = 'han' will be kept and all other items will be removed from the
+    ht.freq, ht.freq_meta_sample_count, and ht.freq_meta.
```
KoalaQin, line 559: I was testing your code, I have two questions about sample_count:
1. If it's indexed by freq_meta, should it have the same length as freq_meta? I found different length before filtering.  
2. was it called `freq_sample_count` in v3, now we change it to `freq_meta_sample_count`?

KoalaQin (changes requested):

gnomad/utils/filtering.py
mike-w-wilson: 1. It should have the same length, what HT are you testing on?
2. Yes, its more clear in my opinion.

KoalaQin: I tested 2 HT: 
v3 release HT and an new HT I generated for the added HGDP/TGP: 
```commandline
from gnomad_qc.v3.resources.release import release_ht_path
ht = hl.read_table(release_ht_path(release_version="3.1.2"))
```
and 
```
ht = hl.read_table(
    'gs://gnomad-tmp/gnomad.exomes.v4.0.qc_data/gnomad.genomes.v4.0.hgdp_1kg_subset_updated_callstats_added.ht')
```
they both didn't have equal length before filtering.

gnomad/utils/filtering.py
```diff
@@ -533,40 +533,46 @@ def split_vds_by_strata(
...
     - By a list of keys, e.g. ["sex", "downsampling"].
     - By specific key: value pairs, e.g. to filter where 'pop' is 'han' or 'papuan'
     {"pop": ["han", "papuan"]}, or where 'pop' is 'afr' and/or 'sex' is 'XX'
     {"pop": ["afr"], "sex": ["XX"]}.
 
-    The items can be kept or removed from `freq_expr` and `freq_meta_expr` based on the
-    value of `keep`.
+    The items can be kept or removed from `meta_indexed_expr` and `meta_expr` based on
+    the value of `keep`. For example if `meta_indexed_exprs` is {'freq': ht.freq,
+'freq_meta_sample_count': ht.index_globals().freq_meta_sample_count} and `meta_expr`
```
KoalaQin, line 555: sorry, you have to add the indentation. I forgot to copy that, like the , from last time.

gnomad/utils/filtering.py
mike-w-wilson: I think you found a bug in v3.1.2. I'm fairly certain we'd want freq_sample_count to match freq_meta unless there is some sort of implied understanding of sample counts in downsamplings and those arent included? maybe @jkgoodrich knows why it doesnt? If 3.1.2 is incorrect and you used in for your subset, then it will also be incorrect right?

gnomad/utils/filtering.py
mike-w-wilson, line 555:
```suggestion
    'freq_meta_sample_count': ht.index_globals().freq_meta_sample_count} and `meta_expr`
```

gnomad/utils/filtering.py
mike-w-wilson: My guess is that the freq_sample_counts do not include the hgdp tgp freq groups

KoalaQin: it doesn't add up though, 598 - hgdp (145)  - tgp (82) = 371 (!= 201)

mike-w-wilson: Maybe also - subsets will get there

KoalaQin:
> I think you found a bug in v3.1.2. I'm fairly certain we'd want freq_sample_count to match freq_meta unless there is some sort of implied understanding of sample counts in downsamplings and those arent included? maybe @jkgoodrich knows why it doesnt? If 3.1.2 is incorrect and you used in for your subset, then it will also be incorrect right? 

For the second HT, I started from MT to calculate callstats by providing a sample list, why could that be different again?

gnomad/utils/filtering.py
jkgoodrich: Just to follow this up with code showing we didn't concatenate the sample count global for the subsets, only the freq_meta
 https://github.com/broadinstitute/gnomad_qc/blob/0c52cf47e48fa5b503d874e96482ea4286474c71/gnomad_qc/v3/create_release/create_release_sites_ht.py#L293

jkgoodrich: Qin, can you add a ticket to update this for the v4 genomes?

KoalaQin: here: https://app.zenhub.com/workspaces/gnomad-5f4d127ea61afc001d6be50b/issues/gh/broadinstitute/gnomad_production/1117

KoalaQin (approved):
LGTM!

gnomad/utils/filtering.py
```diff
@@ -533,40 +533,46 @@ def split_vds_by_strata(
     }
 
 
-def filter_freq_by_meta(
-    freq_expr: hl.expr.ArrayExpression,
-    freq_meta_expr: hl.expr.ArrayExpression,
+def filter_arrays_by_meta(
+    meta_expr: hl.expr.ArrayExpression,
+    meta_indexed_exprs: Dict[str, hl.expr.ArrayExpression],
```
jkgoodrich, line 538: I have one small suggestion. Just to make other code that uses this a little cleaner in the case that only one array needs to be changed, we could have this be:

```suggestion
    meta_indexed_exprs: Union[Dict[str, hl.expr.ArrayExpression], hl.expr.ArrayExpression],
```

Then also update the output type, and add in 

```
if isinstance(meta_indexed_exprs, hl.expr.ArrayExpression):
    meta_indexed_exprs = {"_tmp": meta_indexed_exprs}
```

And then right before the return add:
```
if "_tmp" in meta_indexed_exprs:
    meta_indexed_exprs = meta_indexed_exprs["_tmp"]
```

jkgoodrich (commented):
This looks good to me, just one small suggestion, but if you disagree I will approve

gnomad/utils/filtering.py
mike-w-wilson: Good call, added it!

gnomad/utils/filtering.py
```diff
@@ -533,40 +533,57 @@ def split_vds_by_strata(
...
+def filter_arrays_by_meta(
+    meta_expr: hl.expr.ArrayExpression,
+    meta_indexed_exprs: Union[
+        Dict[str, hl.expr.ArrayExpression], hl.expr.ArrayExpression
+    ],
     items_to_filter: Union[Dict[str, List[str]], List[str]],
     keep: bool = True,
     combine_operator: str = "and",
-) -> Tuple[hl.expr.ArrayExpression, hl.expr.ArrayExpression]:
+) -> Tuple[hl.expr.ArrayExpression, Dict[str, hl.expr.ArrayExpression]]:
```
jkgoodrich, line 544:
```suggestion
) -> Tuple[hl.expr.ArrayExpression, Union[Dict[str, hl.expr.ArrayExpression], hl.expr.ArrayExpression]]:
```

jkgoodrich (approved):
One small change, then LGTM!

PR #589, opened 2023-09-05
Fix `annotate_downsamplings` when used with a Table instead of a MatrixTable

jkgoodrich:
Fix for error `Error summary: AssertionError: assertion failed: Key type struct{r: float64} is not a subset of struct{s: str, global_idx: int64, pop_idx: int64}`


mike-w-wilson (approved):
LGTM

PR #590, opened 2023-09-06
fix negative values issue with 'diff'

KoalaQin:
I think it misses one case: when set_negatives_to_zero=False & the ann or count >=0, we can just use the original value. Can you check if it makes sense?


jkgoodrich (approved):
Great catch! Thank you for fixing!

PR #591, opened 2023-09-06
fix ValueError for count_arrays in merge_freq_arrays function

KoalaQin:
The length check should be for each list of values in the dict, instead of the dict itself.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -1037,8 +1037,9 @@ def merge_freq_arrays(
     if operation not in ["sum", "diff"]:
         raise ValueError("Operation must be either 'sum' or 'diff'!")
     if count_arrays is not None:
-        if len(count_arrays) != len(fmeta):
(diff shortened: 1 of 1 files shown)


gnomad/utils/annotations.py
```diff
@@ -1037,8 +1037,9 @@ def merge_freq_arrays(
     if operation not in ["sum", "diff"]:
         raise ValueError("Operation must be either 'sum' or 'diff'!")
     if count_arrays is not None:
-        if len(count_arrays) != len(fmeta):
-            raise ValueError("Length of count_arrays and fmeta must be equal!")
+        for k, count_array in count_arrays.items():
+            if len(count_array) != len(fmeta):
+                raise ValueError("Length of each count_array and fmeta must be equal!")
```
mike-w-wilson, line 1042:
```suggestion
                raise ValueError(f"Length of  count_array '{k}' and fmeta must be equal!")
```
Sorry I approved but forgot to add this. This makes the error more informative so the user can see what they need to fix.

PR #592, opened 2023-09-07
Modify `apply_rf_model` to use `vector_to_array` from `pyspark.ml.functions` instead of `udf`

matren395:
Update file's apply_rf_model() function to work on GnomAD v4 - still not clear why the prior formulation wasn't working? Will this be a permanent fix , or is it intended to be ? 

JG edit: Current use of `udf` to convert a vector to an array is causing a memory error. `vector_to_array` didn't exist when the original function was written, so this is intended to be a permanent fix that uses a newer function.

### DIFF
--- a/gnomad/variant_qc/random_forest.py
+++ b/gnomad/variant_qc/random_forest.py
@@ -325,6 +325,9 @@ def test_model(
     return test_results
 
 
+from pyspark.ml.functions import vector_to_array
+
+
 def apply_rf_model(
     ht: hl.Table,
     rf_model: pyspark.ml.PipelineModel,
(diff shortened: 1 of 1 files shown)


gnomad/variant_qc/random_forest.py
```diff
@@ -325,6 +325,9 @@ def test_model(
     return test_results
 
 
+from pyspark.ml.functions import vector_to_array
```
jkgoodrich, line 328: This should be added to the top of the file with the other imports. Also, remove any imports that are no longer used after making the change below. 

I think this:
```
from pyspark.sql import SparkSession
from pyspark.sql.functions import col, udf  # pylint: disable=no-name-in-module
from pyspark.sql.types import ArrayType, DoubleType
```

Can be changed to this:
```
from pyspark.ml.functions import vector_to_array
from pyspark.sql.functions import col
```

jkgoodrich (changes requested):
Just one comment about imports

jkgoodrich (approved):
LGTM!

PR #593, opened 2023-09-07
When merging freq arrays, AF will be missing when AN is 0

mike-w-wilson:
Previously the merge function set AF to 0 when AN was 0. This was misleading as it implied coverage/information at these sites.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -1100,7 +1100,7 @@ def merge_freq_arrays(
     new_freq = freq_meta_idx.map(
         lambda x: hl.bind(
             lambda y: y.annotate(
-                AF=hl.if_else(y.AN > 0, y.AC / y.AN, hl.missing(hl.tfloat64))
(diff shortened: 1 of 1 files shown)


gnomad/utils/annotations.py
```diff
@@ -1099,9 +1099,9 @@ def _sum_or_diff_fields(
     callstat_ann_af = ["AC", "AF", "AN", "homozygote_count"]
     new_freq = freq_meta_idx.map(
         lambda x: hl.bind(
-            lambda y: y.annotate(AF=hl.if_else(y.AN > 0, y.AC / y.AN, 0)).select(
-                *callstat_ann_af
-            ),
+            lambda y: y.annotate(
+                AF=hl.if_else(y.AN > 0, y.AC / y.AN, hl.missing(hl.tfloat64))
```
jkgoodrich, line 1103:
```suggestion
                AF=hl.or_missing(y.AN > 0, y.AC / y.AN)
```

jkgoodrich (changes requested):

jkgoodrich (approved):

PR #594, opened 2023-09-08
Modify `split_info_annotation` to allow for splitting an info expression that doesn't include `AS_SB_TABLE`

jkgoodrich:


klaricch (approved):
lgtm

PR #599, opened 2023-09-13
Add back write to parquet file in `apply_rf_model` and make `label` parameter optional

jkgoodrich:
Fixes an error in `apply_rf_model` that resulted in some missing prediction values by adding back in the write to a temp parquet file.


matren395 (approved):
I don't see anything out of the ordinary, lgtm

PR #596, opened 2023-09-11
Copy of: Add gnomad_gks() and get_gks() for extracting gks information for a specified variant

matren395:
Copy of Pull Request "Add gnomad_gks() and get_gks() for extracting gks information for a specified variant" with only relevant changes included. The Git History got extremely confused when consulting with Steve Jahl, and his suggestion in the face of the problem was to copy the relevant changes to a new PR/branch, given how complicated to resolve the GitHistory would have been

PR #601, opened 2023-09-19
Optimize a couple collect calls and add option to disable checkpointing

theferrit32:
- VA freq code needs values from global `ht.freq_index_dict`, but since it's a global and same for every row, it only needs to be read once
- The coverage table only has one row per variant, so it can be joined with the gnomad variant table using a `.join` which is indexed and faster than `.filter`, and makes it so the coverage stats for all the variants in the locus interval can be retrieved in one `.collect` call
- For small-ish tables of tens of thousands of variants or fewer (opposed to millions or more), checkpointing is a significant time overhead to add to what is otherwise relatively fast. So leaving it enabled but having an option to disable it is helpful

### DIFF
--- a/gnomad/resources/grch38/gnomad.py
+++ b/gnomad/resources/grch38/gnomad.py
@@ -497,6 +497,7 @@ def gnomad_gks(
             coverage_ht = hl.read_table(
                 coverage("genomes").versions[coverage_version].path
             )
+        ht = ht.annotate(mean_cov=coverage_ht[ht.locus].mean)
 
     # Retrieve ancestry groups from the imported POPS dictionary.
     pops_list = list(POPS[high_level_version]) if by_ancestry_group else None
@@ -517,8 +518,11 @@ def gnomad_gks(
 
     # Select relevant fields, checkpoint, and filter to interval before adding
     # annotations
-    ht = ht.select(ht.freq, ht.info.vrs, ht.popmax)
-    ht = ht.checkpoint(hl.utils.new_temp_file("vrs_checkpoint", extension="ht"))
+    if not skip_coverage:
+        ht = ht.select(ht.freq, ht.info.vrs, ht.popmax, ht.mean_cov)
+    else:
+        ht = ht.select(ht.freq, ht.info.vrs, ht.popmax)
+    # ht = ht.checkpoint(hl.utils.new_temp_file("vrs_checkpoint", extension="ht"))
     ht = hl.filter_intervals(ht, [locus_interval])
 
     # Collect all variants as structs, so all dictionary construction can be
@@ -547,7 +551,6 @@ def gnomad_gks(
                 input_struct=variant,
                 label_name="gnomAD",
                 label_version=version,
-                coverage_ht=coverage_ht,
                 ancestry_groups=pops_list,
                 ancestry_groups_dict=POP_NAMES,
                 by_sex=by_sex,
--- a/gnomad/resources/grch38/gnomad.py
+++ b/gnomad/resources/grch38/gnomad.py
@@ -524,6 +524,7 @@ def gnomad_gks(
     # Collect all variants as structs, so all dictionary construction can be
     # done in native Python
     variant_list = ht.collect()
+    ht_freq_index_dict = ht.freq_index_dict.collect()[0]
 
     # Assemble output dicts with VRS and optionally frequency, append to list,
     # then return list
@@ -550,7 +551,7 @@ def gnomad_gks(
                 ancestry_groups=pops_list,
                 ancestry_groups_dict=POP_NAMES,
                 by_sex=by_sex,
-                freq_index_dict=ht.freq_index_dict.collect()[0],
+                freq_index_dict=ht_freq_index_dict,
             )
 
             # Assign existing VRS information to "focusAllele" key
--- a/gnomad/resources/grch38/gnomad.py
+++ b/gnomad/resources/grch38/gnomad.py
@@ -452,6 +452,7 @@ def gnomad_gks(
     by_sex: bool = False,
     vrs_only: bool = False,
     custom_ht: hl.Table = None,
+    skip_checkpoint: bool = False,
     skip_coverage: bool = False,
     custom_coverage_ht: hl.Table = None,
 ) -> list:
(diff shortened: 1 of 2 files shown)


theferrit32:
Using this table (@matren395 gave this downsampled to me for use in local testing and I put it in a public bucket temporarily):
`gs://clingen-public/gnomad-gks-downsampled-100k.ht`

And this interval:
```py
larger_interval = hl.locus_interval(
  contig="chr1",
  start=1,
  end=int(1e7),
  reference_genome="GRCh38")
```
This locus filter applied to the above table gives 303 variants.

I got some timing measurements for `gnomad_gks` with these commits applied and checkpointing disabled for all of them.
```
# 1 globals collect, no cov
# Annotated 303 records in 4.556254863739014 seconds (0.02 sec/rec)

# 1 globals collect, with cov, coverage collect per variant (with coverage filter per-variant) 
# Annotated 303 records in 510.0283250808716 seconds (1.68 sec/rec)

# 1 globals collect, with cov, mean cov annotation and 1 cov collect (with coverage join)
# Annotated 303 records in 53.95080280303955 seconds (0.18 sec/rec)
```

theferrit32:
This builds on https://github.com/broadinstitute/gnomad_methods/pull/596

matren395:
this is good work Kyle, thanks! I'll review this now

gnomad/resources/grch38/gnomad.py
```diff
@@ -497,6 +500,7 @@ def gnomad_gks(
             coverage_ht = hl.read_table(
                 coverage("genomes").versions[coverage_version].path
             )
+        ht = ht.annotate(mean_cov=coverage_ht[ht.locus].mean)
```
matren395, line 503:
```suggestion
        ht = ht.annotate(meanDepth=coverage_ht[ht.locus].mean)
```
only runs if skip-coverage is not set, great implementation. If we are calling it meanDepth later in the code though , would be a good idea to keep it consistent

gnomad/resources/grch38/gnomad.py
```diff
@@ -517,13 +521,20 @@ def gnomad_gks(
 
     # Select relevant fields, checkpoint, and filter to interval before adding
     # annotations
-    ht = ht.select(ht.freq, ht.info.vrs, ht.popmax)
-    ht = ht.checkpoint(hl.utils.new_temp_file("vrs_checkpoint", extension="ht"))
+    if not skip_coverage:
+        ht = ht.select(ht.freq, ht.info.vrs, ht.popmax, ht.mean_cov)
```
matren395, line 525:
```suggestion
        ht = ht.select(ht.freq, ht.info.vrs, ht.popmax, ht.meanDepth)
```

gnomad/resources/grch38/gnomad.py
```diff
@@ -517,13 +521,20 @@ def gnomad_gks(
...
+
+    # Checkpoint narrower set of columns if not skipped.
+    if not skip_checkpoint:
+        ht = ht.checkpoint(hl.utils.new_temp_file("vrs_checkpoint", extension="ht"))
     ht = hl.filter_intervals(ht, [locus_interval])
 
     # Collect all variants as structs, so all dictionary construction can be
     # done in native Python
     variant_list = ht.collect()
+    ht_freq_index_dict = ht.freq_index_dict.collect()[0]
```
matren395, line 537: remind me, what was the data you had said for how much this speeds things up? I'm not clear how long this collect statement was taking for each variant

gnomad/utils/annotations.py
```diff
@@ -2056,14 +2055,13 @@ def add_gks_va(
 
     Populate the dictionary with frequency information conforming to the GKS VA frequency schema.
     If ancestry_groups or by_sex is provided, also include subcohort schemas for each cohort.
+    If input_struct has mean_cov, it is added to ancillaryResults.
```
matren395, line 2058:
```suggestion
    If input_struct has meanDepth, it is added to ancillaryResults.
```

gnomad/utils/annotations.py
```diff
@@ -2199,11 +2197,9 @@ def _create_group_dicts(
             "popFreqId": f"{gnomad_id}.{input_struct.popmax.pop.upper()}",
         }
 
-    # Add mean coverage statistics if a coverage Table is provided
-    if coverage_ht is not None:
-        ancillaryResults["meanDepth"] = coverage_ht.filter(
-            coverage_ht.locus == input_struct.locus
-        ).mean.collect()[0]
+    # Add mean coverage statistics if the input was annotated with coverage information
+    if "mean_cov" in input_struct:
```
matren395, line 2201:
```suggestion
    if "meanDepth" in input_struct:
```

gnomad/utils/annotations.py
```diff
@@ -2199,11 +2197,9 @@ def _create_group_dicts(
...
+        ancillaryResults["meanDepth"] = input_struct.mean_cov
```
matren395, line 2202:
```suggestion
        ancillaryResults["meanDepth"] = input_struct.meanDepth
```

matren395 (commented):
good work , just questions about one variable name really

gnomad/resources/grch38/gnomad.py
matren395: nevermind I see your comments above

gnomad/resources/grch38/gnomad.py
```diff
@@ -467,6 +468,8 @@ def gnomad_gks(
     :param vrs_only: Boolean to pass for only VRS info to be returned
         (will not include allele frequency information).
     :param custom_ht: Table to use instead of what public_release() method would return for the version.
+    :param skip_checkpoint: Bool to pass to skip checkpointing selected columns
+        (checkpointing may be desirable for large datasets by reducing data copies across the cluster).
```
klaricch, line 472:
```suggestion
    :param skip_checkpoint: Bool to pass to skip checkpointing selected fields
        (checkpointing may be desirable for large datasets by reducing data copies across the cluster).
```

gnomad/resources/grch38/gnomad.py
klaricch: i like the suggestion to keep "cov" and "depth" consistent, but we should keep variables as snake case rather than camel case, how about "mean_cov" -> "mean_depth"?

klaricch (changes requested):

gnomad/resources/grch38/gnomad.py
```diff
@@ -521,10 +521,11 @@ def gnomad_gks(
 
     # Select relevant fields, checkpoint, and filter to interval before adding
     # annotations
+    keep_fields = [ht.freq, ht.info.vrs, ht.popmax]
     if not skip_coverage:
-        ht = ht.select(ht.freq, ht.info.vrs, ht.popmax, ht.mean_cov)
-    else:
-        ht = ht.select(ht.freq, ht.info.vrs, ht.popmax)
+        keep_fields.append(ht.mean_cov)
```
klaricch, line 526:
```suggestion
        keep_fields.append(ht.mean_depth)
```

klaricch (changes requested):

PR #606, opened 2023-09-25
Drop alleles from ht returned by compute_coverage_stats

klaricch:


gnomad/utils/sparse_mt.py
```diff
@@ -1050,20 +1050,28 @@ def join_with_ref(mt: hl.MatrixTable) -> hl.MatrixTable:
...
+                for i, x in zip(
+                    range(
+                        len(coverage_over_x_bins) - 1, -1, -1
+                    ),  # Reverse the bin index as count_array_expr has the reverse order
+                    coverage_over_x_bins,
+                )
+            },
+        )
+        .rows()
+        .key_by("locus")
```
jkgoodrich, line 1069: as is, I think this is done 2 times

gnomad/utils/sparse_mt.py
```diff
@@ -1050,20 +1050,28 @@ def join_with_ref(mt: hl.MatrixTable) -> hl.MatrixTable:
...
+        .select_globals()
+        .drop("alleles")
+    )
+
```
jkgoodrich, line 1073: I think it should be keyed by `*row_key_fields` and not drop alleles within this function, the user should do that if wanted

jkgoodrich (changes requested):

gnomad/utils/sparse_mt.py
klaricch: oops, meant to delete that first part

gnomad/utils/sparse_mt.py
klaricch: i edited this to rekey by row_key_fields and drop any previous keys, not sure if that's what's wanted, can also drop alleles after this line in gnomad_qc (https://github.com/broadinstitute/gnomad_qc/blob/dbd66b79bb02a4a3bf73c4541de14a0eccf8e98c/gnomad_qc/v4/annotations/compute_coverage.py#L122C14-L122C14)

jkgoodrich: Ah, yes, I think this is good. I was thinking to drop in gnomad_qc, but I think this makes more sense

jkgoodrich (approved):
LGTM

PR #607, opened 2023-09-26
Changes made to get preliminary v4 vqsr results

jkgoodrich:
Hardcoded a few things for testing that will need to be changed

### DIFF
--- a/gnomad/variant_qc/vqsr.py
+++ b/gnomad/variant_qc/vqsr.py
@@ -227,10 +227,7 @@ def snps_variant_recalibrator(
     j.cpu(4)
     j.storage("20G")
 
-    j.declare_resource_group(
-        recalibration={"index": "{root}.idx", "base": "{root}"},
-        tranches={"base": "{root}"},
-    )
+    j.declare_resource_group(recalibration={"index": "{root}.idx", "base": "{root}"})
 
     tranche_cmdl = " ".join(
         [f"-tranche {v}" for v in utils["SNP_RECALIBRATION_TRANCHE_VALUES"]]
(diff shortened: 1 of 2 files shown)


gnomad/variant_qc/vqsr.py
```diff
@@ -659,7 +662,7 @@ def apply_recalibration(
         j.command(f"""
                 interval=$(cat {interval} | tail -n1 | awk '{{print $1":"$2"-"$3}}')
                 bcftools view -t $interval {j.output_vcf['vcf.gz']} --output-file {j.output_vcf['vcf.gz']} --output-type z
-                tabix {j.output_vcf['vcf.gz']}
+                tabix -f {j.output_vcf['vcf.gz']}
```
matren395, line 665: what's the cause for this? is this causing the empty outputs ?

gnomad/variant_qc/vqsr.py
```diff
@@ -1006,10 +1010,14 @@ def vqsr_workflow(
...
     )
 
+    is_small_callset = False
+    is_large_callset = False
     if run_mode == "small":
         is_small_callset = True
     elif run_mode == "large":
         is_large_callset = True
+    else:
+        raise ValueError("Please specify a run mode: small or large.")
```
matren395, line 1020: shouldn't be necessary , lindo's code includes a default size of a cluster that is then modified if it is set as small or large. can probably be more clear about this in arguments though. should I include a 'standard' option in the list of options that is the default ?

gnomad/variant_qc/vqsr.py
```diff
@@ -552,6 +552,7 @@ def gather_tranches(
     j.command(f"""set -euo pipefail
         gatk --java-options "-Xms6g" \\
           GatherTranches \\
+          --gcs-project-for-requester-pays broad-mpg-gnomad \\
```
matren395, line 555: there's an arg for a user's gcp-billing-project, I'll set these as that later

matren395 (commented):
TODO: clarify run size args and set gcp-billing-project args

matren395 (approved):
good to be merged into my branch 👍

PR #614, opened 2023-10-03
Fix merge_histograms for NAs

mike-w-wilson:
This fixes the histogram logic for NAs. The hl.zip().map() was making every histogram an NA whenever there was a missing value for a histogram, i.e. the value was not found in the subset. This is because value + missing = missing. We also did not have fill_missing=True so we did value + NA a lot. The hl.or_else() statement also requires the `hl.literal([hl.missing(hl.tint)]))` not just a `hl.missing(hl.tarray...` because then we are back to adding value + missing. The way implemented here creates an empty list, not a missing empty list.  The `bin_edges` also needed fixing in the case a variant wasnt present in a subset. Sorry my testing didnt catch this.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -1244,10 +1244,14 @@ def merge_histograms(hists: List[hl.expr.StructExpression]) -> hl.expr.Expressio
     return hl.fold(
         lambda i, j: hl.struct(
             **{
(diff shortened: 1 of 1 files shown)


jkgoodrich (approved):
Wow, those missing values really complicated this. Thank you for finding and addressing the issue!

PR #617, opened 2023-10-05
Minor GKS formatting changes and addition of gnomAD flags to annotation

theferrit32:
Modifications:
- change `popFreqID` to `popFreqId`
- change subcohort ids to use a `.` delimiter instead of `,` to match the popmax syntax

Additions:
- add qcFilters
- add lowComplexityRegion (lcr)
- add allele balance flagged count (>0.9 in the table) as `heterozygousSkewedAlleleCount`

One point is that the allele balance code will fail if the number of entries or the bins in the array `ht.qual_hists.ab_hist_alt.bin_freq` changes. That field is directly used to drive the the UI.

EDIT:
Additional changes:
- add hemizygote counts to variant and subcohorts
- if no popmax faf exists, sets the popmaxFAF95 object to null instead of setting the freq to 0 and having null or other string in other fields (@matren395)
- rename `popmaxFAF95` to `grpMaxFAF95`, `popFreqId` to `grpFreqId` (@matren395)


matren395:
Hey Kyle, could we roll the change to "popFreqId" not returning "None" but instead a string here as well? I had created my own PR for that and the "popFreqID" fix, but it looks like this already contains the latter and would be easy enough to host the former as well

gnomad/utils/annotations.py
```diff
@@ -2294,21 +2294,44 @@ def _create_group_dicts(
...
     }
 
+    # Check allele balance for heterozygotes values.
+    # Flagged values are those in bins >0.9. Each is 0.05, so last 2.
+    if len(input_struct.ab_hist_alt.bin_freq) != 20:
+        raise ValueError(
+            f"{gnomad_id} ab_hist_alt.bin_freq had "
+            f"{len(input_struct.ab_hist_alt.bin_freq)} items, expected 20"
+        )
+    # The bin_freq should be in order but we can verify the order from bin_edges
```
klaricch, line 2304:
```suggestion
    # The bin_freq should be in order but we can verify the order from bin_edges.
```

gnomad/utils/annotations.py
```diff
@@ -2294,21 +2294,44 @@ def _create_group_dicts(
...
+                zip(
+                    input_struct.ab_hist_alt.bin_edges,
+                    input_struct.ab_hist_alt.bin_freq,
+                ),
+                key=lambda tpl: tpl[0],
+            ),
+        )
+    )
+
     # Create ancillaryResults for additional frequency and popMaxFAF95 information
```
klaricch, line 2318:
```suggestion
    # Create ancillaryResults for additional frequency and popMaxFAF95 information.
```

gnomad/utils/annotations.py
klaricch, line 2304:
```suggestion
    # The bin_freq should be in order but we can verify the order from bin_edges.
```

gnomad/utils/annotations.py
```diff
@@ -2294,21 +2294,44 @@ def _create_group_dicts(
...
     ancillaryResults = {
         "homozygotes": overall_freq["homozygote_count"],
         "popMaxFAF95": {
             "frequency": input_struct.faf95.popmax,
             "confidenceInterval": 0.95,
         },
+        "qcFilters": list(input_struct.filters),
+        "lowComplexityRegion": input_struct.region_flag.lcr,
+        "heterozygousAlleleBalanceFlagged": sum(ab_bin_freq[-2:]),
```
klaricch, line 2327: so this is just the count of how many genotypes were flagged?

gnomad/utils/annotations.py
```diff
@@ -2294,21 +2294,44 @@ def _create_group_dicts(
...
+    # Flagged values are those in bins >0.9. Each is 0.05, so last 2.
+    if len(input_struct.ab_hist_alt.bin_freq) != 20:
+        raise ValueError(
+            f"{gnomad_id} ab_hist_alt.bin_freq had "
+            f"{len(input_struct.ab_hist_alt.bin_freq)} items, expected 20"
+        )
+    # The bin_freq should be in order but we can verify the order from bin_edges
+    ab_bin_freq = list(
+        map(
+            lambda tpl: tpl[1],
```
klaricch, line 2307: what does tpl stand for?

klaricch (changes requested):

gnomad/utils/annotations.py
theferrit32: "tuple"

gnomad/utils/annotations.py
theferrit32: Yes the count of those over the 0.9 threshold. Let me know if this seems correct or if there are any weird cases that might need to be handled differently

gnomad/utils/annotations.py
```diff
@@ -2294,21 +2294,44 @@ def _create_group_dicts(
...
 
     if input_struct.faf95.popmax_population is not None:
-        ancillaryResults["popMaxFAF95"]["popFreqID"] = (
-            f"{gnomad_id}.{input_struct.faf95.popmax_population.upper()}",
-        )
+        popFreqId = f"{gnomad_id}.{input_struct.faf95.popmax_population.upper()}"
+        ancillaryResults["popMaxFAF95"]["popFreqId"] = popFreqId
     else:
-        ancillaryResults["popMaxFAF95"]["popFreqID"] = None
+        ancillaryResults["popMaxFAF95"]["popFreqId"] = None
```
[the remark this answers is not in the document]
theferrit32: I've added it just now based on the discussion, but it can be changed if they decide on something else

theferrit32:
One more pending change is to add `count of hemizygotes`. Which needs to be computed, it is not explicitly in a field on the table.

@larrybabb @ahwagner  Was there an idea to move some of these fields out of `ancillaryResults`?

see: https://github.com/ga4gh/va-spec/pull/119#issuecomment-1751097147

I think some of that might have been related to my confusion. I'm not sure the allele balance flagged actually are directly related to the popmax FAF 95. Maybe someone on the gnomad team can clarify.

gnomad/utils/annotations.py
```diff
@@ -2312,21 +2312,45 @@ def _create_group_dicts(
...
+    # Each bin is 0.05, so flagged values are in the last 2 bins.
+    if len(input_struct.ab_hist_alt.bin_freq) != 20:
+        raise ValueError(
+            f"{gnomad_id} ab_hist_alt.bin_freq had "
+            f"{len(input_struct.ab_hist_alt.bin_freq)} items, expected 20"
+        )
+    # The bin_freq should be in order but we can verify the order from bin_edges.
+    ab_bin_freq = list(
+        map(
+            lambda tpl: tpl[1],
```
klaricch, line 2326:
```suggestion
            lambda x: x[1],
```

gnomad/utils/annotations.py
```diff
@@ -2312,21 +2312,45 @@ def _create_group_dicts(
...
+            sorted(
+                zip(
+                    input_struct.ab_hist_alt.bin_edges,
+                    input_struct.ab_hist_alt.bin_freq,
+                ),
+                key=lambda tpl: tpl[0],
```
klaricch, line 2332:
```suggestion
                key=lambda x: x[0],
```

gnomad/utils/annotations.py
```diff
@@ -2312,21 +2312,45 @@ def _create_group_dicts(
...
     }
 
     if input_struct.faf95.popmax_population is not None:
-        ancillaryResults["popMaxFAF95"][
-            "popFreqID"
-        ] = f"{gnomad_id}.{input_struct.faf95.popmax_population.upper()}"
+        popFreqId = f"{gnomad_id}.{input_struct.faf95.popmax_population.upper()}"
     else:
-        ancillaryResults["popMaxFAF95"]["popFreqID"] = None
+        popFreqId = f"{gnomad_id}.NO_COHORTS"
```
klaricch, line 2352: looks like we're no longer going with the NO_COHORTS option so all of popMaxFAF95 should be null in these cases

gnomad/utils/annotations.py
```diff
@@ -2312,21 +2312,45 @@ def _create_group_dicts(
...
+    # Create ancillaryResults for additional frequency and popMaxFAF95 information.
     ancillaryResults = {
         "homozygotes": overall_freq["homozygote_count"],
         "popMaxFAF95": {
             "frequency": input_struct.faf95.popmax,
             "confidenceInterval": 0.95,
         },
+        "qcFilters": list(input_struct.filters),
+        "lowComplexityRegion": input_struct.region_flag.lcr,
+        "heterozygousAlleleBalanceFlagged": sum(ab_bin_freq[-2:]),
```
klaricch, line 2346: regarding your comment, i'm not aware of any way this is related to faf

klaricch (changes requested):

matren395:
commit to roll in 1) nullable popMaxFAF95 and 2) change all references of 'pop' returned to 'grp' when relevant.

gnomad/utils/annotations.py
```diff
@@ -2312,21 +2325,55 @@ def _create_group_dicts(
...
-        ] = f"{gnomad_id}.{input_struct.faf95.popmax_population.upper()}"
+        grpMaxFAF95 = (
+            {
+                "frequency": input_struct.faf95.popmax,
+                "confidenceInterval": 0.95,
+                "grpFreqId": (
+                    f"{gnomad_id}.{input_struct.faf95.popmax_population.upper()}"
+                ),
+            },
+        )
```
matren395, line 2372: @klaricch this is the added code (in this region) to fix their new schema for accepting NONE types - lmk if this looks right

theferrit32:
Waiting for final modifications to upstream schema. There is one more discussion tomorrow. The changes will be relatively easy to implement after that.

See https://github.com/ga4gh/va-spec/pull/121

theferrit32:
Adding commits and force pushing a branch rebased from main

gnomad/resources/grch38/gnomad.py
```diff
@@ -430,6 +430,79 @@ def release_vcf_path(data_type: str, version: str, contig: str) -> str:
...
+                    popmax=fafs[0].faf,
+                    popmax_population=hl.if_else(
+                        fafs[0].faf == 0, hl.missing(hl.tstr), fafs[0].population
+                    ),
+                ),
+                hl.struct(
+                    popmax=hl.missing(hl.tfloat), popmax_population=hl.missing(hl.tstr)
+                ),
+            ),
+        ),
```
matren395, line 466: remind me, is the rbind code the same from before?

gnomad/resources/grch38/gnomad.py
```diff
@@ -430,6 +430,79 @@ def release_vcf_path(data_type: str, version: str, contig: str) -> str:
...
+            lambda fafs: hl.if_else(
+                hl.len(fafs) > 0,
+                hl.struct(
+                    popmax=fafs[0].faf,
+                    popmax_population=hl.if_else(
+                        fafs[0].faf == 0, hl.missing(hl.tstr), fafs[0].population
+                    ),
+                ),
+                hl.struct(
+                    popmax=hl.missing(hl.tfloat), popmax_population=hl.missing(hl.tstr)
```
matren395, line 463: Remind me, the I'm having a hard time parsing the Schema. 'GrpMaxFAF95' is required to exist but can just return values of None and None within it?

gnomad/utils/annotations.py
```diff
@@ -2244,8 +2243,16 @@ def _create_group_dicts(
         :return: Dictionary containing variant frequency information,
             - (by genetic ancestry group and sex if desired) for specified variant.
         """
+        if group_sex:
+            cohort_id = f"{group_id.upper()}.{group_sex}"
+            freq_index_key = f"{group_id}_{group_sex}_adj"
+        else:
+            cohort_id = f"{group_id.upper()}"
+            freq_index_key = f"{group_id}_adj"
+        record_id = f"{gnomad_id}.{cohort_id}"
```
matren395, line 2252: looks like a bit of code - does this speed up performance?

gnomad/utils/annotations.py
```diff
@@ -2326,28 +2339,78 @@ def _create_group_dicts(
...
-    if input_struct.faf95.popmax_population is not None:
-        ancillaryResults["popMaxFAF95"][
-            "popFreqID"
-        ] = f"{gnomad_id}.{input_struct.faf95.popmax_population.upper()}"
-    else:
-        ancillaryResults["popMaxFAF95"]["popFreqID"] = None
+    # Add hemizygote count if not autosomal or PAR.
+    if not input_struct.in_autosome_or_par:
+        hemizygote_count = input_struct.freq[freq_index_dict["XY_adj"]].AC
+        ancillaryResults["hemizygotes"] = hemizygote_count
```
matren395, line 2350: is this the same code that starts at line #2279? This looks like the simpler version here, but I wanna make sure that I'm not just turned around

matren395 (commented):
might take another run at the actual code in annotations.py later today

gnomad/resources/grch38/gnomad.py
theferrit32: @matren395 yes this is just a copy-paste of the rbind code you wrote. Just moved it to a function so I could swap between v3 and v4 strategies.

gnomad/resources/grch38/gnomad.py
theferrit32: It's not required to exist anymore, instead of making it required and allowing None, they just made it optional.

https://github.com/ga4gh/va-spec/blob/1.0-alpha/schema/cohortAlleleFreq.json#L341-L359

gnomad/utils/annotations.py
theferrit32: I think the reason I did this is so all the `freq` keys are created in this inner function instead of where the function is called. They are all based on the group_id without sex information. Previously it was sending both a group_id including sex, and the sex label itself. And now that we are also pulling out the `hemizogytes`, we need the group_id without sex in order to make the key with XY.

gnomad/utils/annotations.py
theferrit32: Essentially yes, the reason this is simpler is because it's the top level group so the key is always just `XY_adj`. In the inner function it needs to check whether the current group is sexed or not, and if not, query the `freq` for the sexed sub-group.

gnomad/resources/grch38/gnomad.py
```diff
@@ -472,9 +545,11 @@ def gnomad_gks(
     # Read coverage statistics if requested
     if high_level_version == "v3":
         coverage_version = "3.0.1"
+    elif high_level_version == "v4":
+        coverage_version = "3.0.1"
```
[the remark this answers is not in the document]
theferrit32: @klaricch is there just exomes coverage available in v4? That's all I see in the gnomad downloads page and in the `coverage` function. If coverage info is requested for a v4 genomes table, what should the behavior be?

theferrit32:
Latest changeset: https://github.com/theferrit32/gnomad_methods/compare/374112031d6b1178e40cfc23161a19b82b9d5693..74d869e022933612e566755f42d8dd2e0445e114

- fixes fafmax for exomes
- uses v4 exomes coverage (for both v4 genomes and v4 exomes, is this correct? or should genomes tables not be given any coverage info?)
- replaces AF = None (AC/AN = 0/0) with 0. (the jsonschema requires number, can't be null)

gnomad/resources/grch38/gnomad.py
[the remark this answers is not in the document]
matren395: ah! gnomAD v4 genomes are the same as gnomAD v3 genomes , so the coverage ought to be the same for v4 genomes as for v3 genomes.  This might require some extra logic with the user selecting a data type - I forget if that is already requested

gnomad/resources/grch38/gnomad.py
matren395: 👍

gnomad/resources/grch38/gnomad.py
matren395: The way I'm always reading it, in this code, it looks like GrpMaxFAF95 always exists but everything it's populated with is allowed to be None. 1) Am I reading this implementation right 2) this implementation is allowed, right? Sorry if this is v base level

gnomad/utils/annotations.py
matren395: Can we remove the code at line #2279 then? Again sorry if I'm parsing this wrong

gnomad/utils/annotations.py
matren395: 👍

gnomad/utils/annotations.py
matren395: adding a new review, pinging this again if we can remove the code at line #2279-2290 given this code

gnomad/resources/grch38/gnomad.py
```diff
@@ -483,9 +549,16 @@ def gnomad_gks(
...
-            )
+            # in v4, only exomes have coverage
+            # in v3, only genomes have coverage (only genomes exist)
+            if high_level_version == "v3":
+                coverage_ht = hl.read_table(
+                    coverage("genomes").versions[coverage_version].path
+                )
+            else:
+                coverage_ht = hl.read_table(
+                    coverage("exomes").versions[coverage_version].path
```
matren395, line 560: this is tricky, since v4 has genomes (from v3 but still) and exomes as well. could include the Data_Type parameter and say: 

if data_type==exomes: 
     if high_level_version == "v3": ERROR
     elif high_level_version == "v4": ....coverage("exomes")...
elif data_type==genomes:
     ...coverage("genomes")

but I'm not sure if the information in v4 genomes is different than v3 genomes ? I know it's the same samples but idk if it's identical

matren395 (commented):
notes! lmk if you wanna chat abt the PR next week

gnomad/resources/grch38/gnomad.py
klaricch: just summing up coverage comments here:
I put in a fix to the coverage function while working on constraint so that it no longer tries to pull coverage versions that don't exist ( so you no longer have to assign "coverage_version" to avoid errors). v4 genomes is v3 genomes, so there are no new genome coverage files for v4. You can just pull call the coverage function and specify data_type and don't need to specify version. So for v4, using a datatype of either exomes or genomes will be fine since genomes will just pull the v3 coverage data. For v3, using a datatype of genomes is fine, but should error out if a user tries to pull exomes for v3.

"over_20" should also be added to the coverage annotation to be consistent with what's shown on the browser

gnomad/utils/annotations.py
theferrit32: @matren395 I don't think the lines 2279-2289 can be removed or combined here, it adds logic to look up the specific population group. It's possible both 2348-2350 and 2279-2289 could be pulled out into a function but it would probably be more lines and more complicated

gnomad/resources/grch38/gnomad.py
matren395: ah okay, makes sense ! thanks for the answer

gnomad/utils/annotations.py
matren395: noted! thanks for the explanation

gnomad/utils/annotations.py
```diff
@@ -2322,32 +2337,84 @@ def _create_group_dicts(
...
+
+    # Add group max FAF if it exists
+    if input_struct.grpMaxFAF95.popmax_population is not None:
+        ancillaryResults["grpMaxFAF95"] = {
```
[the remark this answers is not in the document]
matren395: makes sense, looks good!

gnomad/resources/grch38/gnomad.py
```diff
@@ -430,6 +430,79 @@ def release_vcf_path(data_type: str, version: str, contig: str) -> str:
...
+    Accepts tables with popmax FAF fields in v3 and v4 formats.
+    If the table includes the v4 style joint_fafmax, also adds jointGrpMaxFAF95.
+
+    :param ht: Input hail table. Can be v3 or v4.
+    :return: Annotated hail table.
+    """
+    if "fafmax" in ht.row and "faf95_max" in ht.fafmax:
+        return add_grpMaxFAF95_v4(ht)
+    else:
+        return add_grpMaxFAF95_v3(ht)
```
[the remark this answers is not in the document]
matren395: thinks this was resolved someplace else, function structuring here makes sense

gnomad/resources/grch38/gnomad.py
klaricch: @theferrit32 it looks like coverage is the last item that still needs changes, can you edit the functions to address the above comment regarding coverage?

theferrit32: @klaricch I'm about to push a change to handle that based on your comment above.

For the "over_20" comment, are you suggesting to add the `.over_20` field from the coverage_ht to the annotation output in addition to the `meanDepth` annotation?

@ahwagner @larrybabb do we want to add a field to include the `Fraction of individuals with >20x coverage` field added to the browser UI?

klaricch: yeah over_20 in additions to meanDepth

larrybabb: @theferrit32 I'm fine with adding the two new fields, but I defer to @ahwagner for the final say. I'll make sure he's seeing these questions.

ahwagner: @theferrit32 to make sure I'm following this proposal, is this an addition of `over_20` to the `caf.qualityMeasures` object? And no other additions? If so, this seems like the appropriate place to add the field and I have no issue with it so long as there is a PR to update the [model](https://github.com/ga4gh/va-spec/blob/1.0-alpha/schema/cohortAlleleFreq.yaml) and [examples](https://github.com/ga4gh/va-spec/blob/1.0-alpha/docs/Modeling/CohortFrequency/simple_result_example.yaml) at VA-Spec.

theferrit32: @ahwagner https://github.com/ga4gh/va-spec/pull/124

gnomad/resources/grch38/gnomad.py
```diff
@@ -549,16 +558,9 @@ def gnomad_gks(
...
-            if high_level_version == "v3":
-                coverage_ht = hl.read_table(
-                    coverage("genomes").versions[coverage_version].path
-                )
-            else:
-                coverage_ht = hl.read_table(
-                    coverage("exomes").versions[coverage_version].path
-                )
+            coverage_ht = hl.read_table(
+                coverage(data_type).versions[coverage_version].path
```
theferrit32, line 562: @klaricch this will use `genomes` version `3.0.1` coverage table for v3 and v4 genomes tables, and `exomes` version `4.0` coverage table for v4 exomes tables. If the input is a v3 table with `data_type=exomes`, line 562 will throw an exception `KeyError: '3.0.1'`, which is expected.

gnomad/sample_qc/ancestry.py
```diff
@@ -50,6 +50,7 @@
     "oeu": "Other European",
     "onf": "Other Non-Finnish European",
     "unk": "Unknown",
+    "remaining": "Remaining",
```
theferrit32, line 53: This is resulting in a python error when merging into main, because it was added separately in that branch, on a different line

https://github.com/broadinstitute/gnomad_methods/blob/main/gnomad/sample_qc/ancestry.py#L35

theferrit32:
Rebasing from main to fix https://github.com/broadinstitute/gnomad_methods/pull/617#pullrequestreview-1816172266

matren395 (commented):
Coveraged looks covered!

theferrit32:
I think @klaricch has to submit a final review approval since she had a prior request for changes

gnomad/resources/grch38/gnomad.py
```diff
@@ -470,11 +534,22 @@ def gnomad_gks(
     high_level_version = f"v{version.split('.')[0]}"
 
     # Read coverage statistics if requested
+    coverage_version_3_0_1 = "3.0.1"  # v3/v4 genomes coverage
```
matren395, line 537:
```suggestion
    coverage_version_3_0_1 = "3.0.1"  # v4 genomes coverage
```

gnomad/resources/grch38/gnomad.py
```diff
@@ -470,11 +534,22 @@ def gnomad_gks(
...
+    coverage_version_4_0 = "4.0"  # v4 exomes coverage
+
+    # In v4, exomes have coverage in v4 coverage table,
+    #   genomes have coverage in v3 coverage table.
+    # In v3, only genomes have coverage (only genomes exist).
     if high_level_version == "v3":
```
matren395, line 543:
```suggestion
    if data_type == "genomes":
```

gnomad/resources/grch38/gnomad.py
```diff
@@ -470,11 +534,22 @@ def gnomad_gks(
...
-        coverage_version = "3.0.1"
+        coverage_version = coverage_version_3_0_1
+    elif high_level_version == "v4":
+        if data_type == "exomes":
+            coverage_version = coverage_version_4_0
+        elif data_type == "genomes":
+            coverage_version = coverage_version_3_0_1
```
matren395, line 549:
```suggestion
    else:
       coverage_version = coverage_version_4_0
```

gnomad/resources/grch38/gnomad.py
```diff
@@ -470,11 +534,22 @@ def gnomad_gks(
...
     else:
         raise NotImplementedError(
-            "gnomad_gks() is currently only implemented for gnomAD v3."
+            "gnomad_gks() is currently only implemented for gnomAD v3 and v4."
```
matren395, line 552:
```suggestion
            "gnomad_gks() is currently only implemented for gnomAD v4."
```

gnomad/resources/grch38/gnomad.py
```diff
@@ -470,11 +534,22 @@ def gnomad_gks(
...
+    # In v3, only genomes have coverage (only genomes exist).
     if high_level_version == "v3":
-        coverage_version = "3.0.1"
+        coverage_version = coverage_version_3_0_1
+    elif high_level_version == "v4":
+        if data_type == "exomes":
+            coverage_version = coverage_version_4_0
+        elif data_type == "genomes":
+            coverage_version = coverage_version_3_0_1
     else:
```
matren395, line 550:
```suggestion
    
    if high_level_version != "v4":
```

gnomad/resources/grch38/gnomad.py
```diff
@@ -503,43 +579,45 @@ def gnomad_gks(
...
-            ),
-        )
-    )
 
-    keep_fields = [ht.freq, ht.info.vrs, ht.faf95]
+    # Pull up LCR flag and make referrable in the same field
+    if high_level_version == "v3":  # v3
+        ht = ht.annotate(lcr=ht.region_flag.lcr)
+    else:  # v4
+        ht = ht.annotate(lcr=ht.region_flags.lcr)
```
matren395, line 589:
```suggestion
    ht = ht.annotate(lcr=ht.region_flags.lcr)
```

gnomad/resources/grch38/gnomad.py
```diff
@@ -470,11 +534,22 @@ def gnomad_gks(
     high_level_version = f"v{version.split('.')[0]}"
 
     # Read coverage statistics if requested
+    coverage_version_3_0_1 = "3.0.1"  # v3/v4 genomes coverage
+    coverage_version_4_0 = "4.0"  # v4 exomes coverage
+
+    # In v4, exomes have coverage in v4 coverage table,
+    #   genomes have coverage in v3 coverage table.
+    # In v3, only genomes have coverage (only genomes exist).
```
matren395, line 542:
```suggestion
```

matren395 (commented):
An effort to clean up and show that we only support v4 versions - let me know if I'm missing anything , might take a finer pass at this tomorrow

gnomad/resources/grch38/gnomad.py
```diff
@@ -503,43 +579,45 @@ def gnomad_gks(
...
+    # Pull up v3 / v4 allele balance histogram arrays
+    if high_level_version == "v3":  # v3
+        ht = ht.annotate(ab_hist_alt=ht.qual_hists.ab_hist_alt)
+    else:  # v4
+        ht = ht.annotate(ab_hist_alt=ht.histograms.qual_hists.ab_hist_alt)
+
+    if high_level_version == "v3":
+        ht = add_grpMaxFAF95_v3(ht)
+    else:
+        ht = add_grpMaxFAF95_v4(ht)
```
matren395, line 600:
```suggestion
    ht = ht.annotate(ab_hist_alt=ht.histograms.qual_hists.ab_hist_alt)

    ht = add_grpMaxFAF95_v4(ht)
```

matren395: appropriate change if fully not supporting v3 at all - may need beautified with some more comments or documentation around here, or running black on it again

matren395 (commented):
some more version-relevant changes if we are removing support for v3

gnomad/utils/annotations.py
```diff
@@ -2244,8 +2243,16 @@ def _create_group_dicts(
         :return: Dictionary containing variant frequency information,
             - (by genetic ancestry group and sex if desired) for specified variant.
         """
+        if group_sex:
```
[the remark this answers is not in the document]
theferrit32: Yeah that was used when I was initially implementing by using a VCF export/import strategy. I will delete now.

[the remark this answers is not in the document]
theferrit32: Deleted

gnomad/resources/grch38/gnomad.py
```diff
@@ -470,11 +495,19 @@ def gnomad_gks(
...
+    coverage_version_4_0 = "4.0"  # v4 exomes coverage
+
+    # In v4, exomes have coverage in v4 coverage table,
+    #   genomes have coverage in v3 coverage table.
+    if data_type == "genomes":
+        coverage_version = coverage_version_3_0_1
     else:
+        coverage_version = coverage_version_4_0
+
+    if high_level_version != "v4":
```
matren395, line 508: Just occurred to me - would make more sense to move this to the start of the function (and accompanying high_level_version logic) before reading in any tables. 

@klaricch your other PR has already been merged and we can leave it up to the other functions to fail for bad keys, yes ? We don't want to check here that like 'version' is in the list of valid versions or anything, yeah ?

matren395 (commented):
only one thought to make the logic more logical

klaricch (changes requested):

gnomad/resources/grch38/gnomad.py
theferrit32: I will move the high_level_version stuff up above the first table read

klaricch (approved):

PR #621, opened 2023-10-11
Remove inadvertent tuple in popMaxFAF95 field in add_gks_va function

mattsolo1:
Currently the `popFreqID` field returned by `add_gks_va` is a tuple instead of a string due to an extra comma. This PR removes the comma, ensuring a string is returned.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -2308,9 +2308,9 @@ def add_gks_va(
     }
 
     if input_struct.faf95.popmax_population is not None:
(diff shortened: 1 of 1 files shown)


matren395:
the only change is that one comma , and we're never expecting this field to have more than one value right

mattsolo1:
See: https://codereview.doctor/features/python/best-practice/tuple-accidental-trailing-comma

matren395 (approved):

matren395:
approved! thanks for finding this

PR #622, opened 2023-10-12
Add methylation resource for chrX

klaricch:


klaricch:
I don't have write access, so I can't actually put those files at the specified paths

gnomad/resources/grch38/reference_data.py
```diff
@@ -322,6 +322,15 @@ def _import_ensembl_interval(path) -> hl.Table:
     },
 )
 
+# Methylation scores range from 0-12.
```
jkgoodrich, line 325: Are these from the same `(https://www.biorxiv.org/content/10.1101/2022.03.20.485034v2.full).` or if not is there a comment we can add indicating where they came from?

gnomad/resources/grch38/reference_data.py
```diff
@@ -322,6 +322,15 @@ def _import_ensembl_interval(path) -> hl.Table:
...
+methylation_sites_chrx = GnomadPublicTableResource(
+    path="gs://gcp-public-data--gnomad/resources/grch38/methylation_sites/methylation_chrX.ht",
```
jkgoodrich, line 327:
```suggestion
    path="gs://gnomad-public-requester-pays/resources/grch38/methylation_sites/methylation_chrX.ht",
```

gnomad/resources/grch38/reference_data.py
```diff
@@ -322,6 +322,15 @@ def _import_ensembl_interval(path) -> hl.Table:
...
+    import_func=_import_methylation_sites,
+    import_args={
+        "path": "gs://gcp-public-data--gnomad/resources/grch38/methylation_sites/methylation_chrX.bed",
```
jkgoodrich, line 330:
```suggestion
        "path": "gs://gnomad-public-requester-pays/resources/grch38/methylation_sites/methylation_chrX.bed",
```

jkgoodrich (changes requested):
If using `gs://gnomad-public-requester-pays/` it will work while it's only located in that folder too, and then it will still work as expected when that file is removed and it's only in the `gcp-public-data--gnomad` bucket

jkgoodrich (approved):
lgtm

PR #625, opened 2023-10-15
Update vep struct to csq to allow omitting sift and polyphen

mike-w-wilson:
This allows sift and polyphen to be omitted from the vep csq annotation


jkgoodrich (changes requested):
This looks good, except I think it is accidentally removing the `ensembl_only` parameter that Kristen added and is needed for constraint

mike-w-wilson:
Sorry about that!

jkgoodrich (approved):
lgtm

PR #626, opened 2023-10-16
Move ga4gh imports to their functions

mike-w-wilson:
The VRS module is breaking our loggers. To keep our versions pinned to 0.8.4 and get logging working, we need to move the imports into the functions that use them. This means loggers will only break when using these functions. 

I think what is happening is their init calls logging.BasicConfig without setting a level which will reset the logging level to warning: https://github.com/ga4gh/vrs-python/blob/593508c6e8229336ca1f53a06f69966020cd68f7/src/ga4gh/vrs/__init__.py

I tested this locally and my logging was restored.

It does look like they removed the logging call in later version so another option is to unpin or use another version of vrs in our requirements file.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -7,6 +7,8 @@ import logging
 from timeit import default_timer as timer
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
 
+import ga4gh.core as ga4gh_core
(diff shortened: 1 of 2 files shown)


jkgoodrich (commented):
This works for me, can you just add a comment for why they are not at the top

KoalaQin (approved):

PR #630, opened 2023-10-18
Add pprint globals and a global/row length comparison, updates monoallelic expr in validity checks

mike-w-wilson:
This adds pprint of globals and a comparison of associated row and global annotation's lengths. It also opens up monoallelic_expr to accept a dictionary so we can also pass `only_het`. The comp naming is horrible, suggestions welcome. 

I added the new checks to the validate_release_t as well though we wont use it as currently implemented for v4 because we don't want to write out the  annotations required for the global and row annotation comp.

### DIFF
--- a/gnomad/assessment/validity_checks.py
+++ b/gnomad/assessment/validity_checks.py
@@ -871,7 +871,7 @@ def compare_related_global_and_row_lengths(
     Compare the lengths of global and row annotations.
 
     :param t: Input MatrixTable or Table.
-    :param len_comp_globals_rows: Dictionary with global annotation (key) and list of row annotations (value) to compare.
+    :param len_comp_globals_rows: Dictionary with row annotation (key) and list of associated global annotations (value) to compare.
     :return: None
     """
     t = t.rows() if isinstance(t, hl.MatrixTable) else t
--- a/gnomad/assessment/validity_checks.py
+++ b/gnomad/assessment/validity_checks.py
@@ -882,11 +882,11 @@ def compare_related_global_and_row_lengths(
             outcome = "Failed" if global_len != row_len else "Passed"
             logger.info(
                 f"%s global and row lengths comparison: Length of %s in"
-                f" globals (%d) does %s match length of %s in rows (%d)",
+                f" globals (%d) does %smatch length of %s in rows (%d)",
                 outcome,
                 global_field,
                 global_len,
-                "NOT" if outcome == "Failed" else "",
+                "NOT " if outcome == "Failed" else "",
                 row_field,
                 row_len,
             )
--- a/gnomad/assessment/validity_checks.py
+++ b/gnomad/assessment/validity_checks.py
@@ -229,7 +229,7 @@ def summarize_variant_filters(
     variant_filter_field: str = "RF",
     problematic_regions: List[str] = ["lcr", "segdup", "nonpar"],
     single_filter_count: bool = False,
-    monoallelic_expr: Optional[
+    site_gt_check_expr: Optional[
         Union[hl.expr.BooleanExpression, Dict[str, hl.expr.BooleanExpression]]
     ] = None,
     extra_filter_checks: Optional[Dict[str, hl.expr.Expression]] = None,
@@ -254,7 +254,7 @@ def summarize_variant_filters(
     :param variant_filter_field: String of variant filtration used in the filters annotation on `ht` (e.g. RF, VQSR, AS_VQSR). Default is "RF".
     :param problematic_regions: List of regions considered problematic to run filter check in. Default is ["lcr", "segdup", "nonpar"].
     :param single_filter_count: If True, explode the Table's filter column and give a supplement total count of each filter. Default is False.
-    :param monoallelic_expr: Optional boolean expression or dictionary of strings and boolean expressions used to log how many monoallelic sites are in the Table.
+    :param site_gt_check_expr: Optional boolean expression or dictionary of strings and boolean expressions typically used to log how many monoallelic or 100% heterozygous sites are in the Table.
     :param extra_filter_checks: Optional dictionary containing filter condition name (key) and extra filter expressions (value) to be examined.
     :param n_rows: Number of rows to display only when showing percentages of filtered variants grouped by multiple conditions. Default is 50.
     :param n_cols: Number of columns to display only when showing percentages of filtered variants grouped by multiple conditions. Default is 140.
@@ -270,15 +270,15 @@ def summarize_variant_filters(
         filters = exp_t.aggregate(hl.agg.counter(exp_t.filters))
         logger.info("Exploded variant filter counts: %s", filters)
 
-    if monoallelic_expr is not None:
-        if isinstance(monoallelic_expr, hl.expr.BooleanExpression):
-            monoallelic_expr = {"monoallellic": monoallelic_expr}
-        for k, m_expr in monoallelic_expr.items():
+    if site_gt_check_expr is not None:
+        if isinstance(site_gt_check_expr, hl.expr.BooleanExpression):
+            site_gt_check_expr = {"site GT check": site_gt_check_expr}
+        for k, m_expr in site_gt_check_expr.items():
             if isinstance(t, hl.MatrixTable):
-                mono_sites = t.filter_rows(m_expr).count_rows()
+                gt_check_sites = t.filter_rows(m_expr).count_rows()
             else:
-                mono_sites = t.filter(m_expr).count()
-            logger.info("There are %d %s sites in the dataset.", mono_sites, k)
+                gt_check_sites = t.filter(m_expr).count()
+            logger.info("There are %d %s sites in the dataset.", gt_check_sites, k)
 
     filtered_expr = hl.len(t.filters) > 0
     problematic_region_expr = hl.any(
@@ -755,10 +755,8 @@ def compute_missingness(
     t = t.rows() if isinstance(t, hl.MatrixTable) else t
 
     logger.info(
-        (
-            "Missingness threshold (upper cutoff for what is allowed for missingness"
-            " checks): %.2f"
-        ),
+        "Missingness threshold (upper cutoff for what is allowed for missingness"
+        " checks): %.2f",
         missingness_threshold,
     )
     metrics_missing = {}
@@ -884,10 +882,8 @@ def compare_global_and_row_annot_lengths(
             global_len = hl.eval(hl.len(t[global_field]))
             outcome = "Failed" if global_len != row_len else "Passed"
             logger.info(
-                (
-                    "%s global and row lengths comparison: Length of %s in"
-                    " globals (%d) does %smatch length of %s in rows (%d)"
-                ),
+                "%s global and row lengths comparison: Length of %s in"
+                " globals (%d) does %smatch length of %s in rows (%d)",
                 outcome,
                 global_field,
                 global_len,
@@ -912,7 +908,7 @@ def validate_release_t(
     subsets: List[str] = [""],
     pops: List[str] = POPS[CURRENT_MAJOR_RELEASE],
     missingness_threshold: float = 0.5,
-    monoallelic_expr: Optional[
+    site_gt_check_expr: Optional[
         Union[hl.expr.BooleanExpression, Dict[str, hl.expr.BooleanExpression]]
     ] = None,
     verbose: bool = False,
@@ -953,7 +949,7 @@ def validate_release_t(
     :param subsets: List of subsets to be checked.
     :param pops: List of pops within main callset.
     :param missingness_threshold: Upper cutoff for allowed amount of missingness. Default is 0.5.
-    :param monoallelic_expr: Optional boolean expression or dictionary of strings and boolean expressions used to log how many monoallelic sites are in `t`.
+    :param site_gt_check_expr: Optional boolean expression or dictionary of strings and boolean expressions typically used to log how many monoallelic or 100% heterozygous sites are in the Table.
     :param verbose: If True, display top values of relevant annotations being checked, regardless of whether check conditions are violated; if False, display only top values of relevant annotations if check conditions are violated.
     :param show_percent_sites: Show percentage of sites that fail checks. Default is False.
     :param delimiter: String to use as delimiter when making group label combinations. Default is "-".
@@ -996,7 +992,7 @@ def validate_release_t(
             variant_filter_field,
             problematic_regions,
             single_filter_count,
-            monoallelic_expr,
+            site_gt_check_expr,
         )
 
     if raw_adj_check:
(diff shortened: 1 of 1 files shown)


gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +863,53 @@ def vcf_field_check(
...
+    """
+    Compare the lengths of global and row annotations.
+
+    :param t: Input MatrixTable or Table.
+    :param len_comp_globals_rows: Dictionary with row annotation (key) and list of associated global annotations (value) to compare.
+    :return: None
+    """
+    t = t.rows() if isinstance(t, hl.MatrixTable) else t
+    for row_field, global_fields in len_comp_globals_rows.items():
+        row_len = len(t.head(1)[row_field].collect()[0])
```
ch-kr, line 879: I don't think this will make a difference in performance, but you could do this above
```
t = t.rows() if isinstance(t, hl.MatrixTable) else t
t.head(1)
```
since you only ever need to access the first row in the for loop below

gnomad/assessment/validity_checks.py
```diff
@@ -251,7 +254,7 @@ def summarize_variant_filters(
     :param variant_filter_field: String of variant filtration used in the filters annotation on `ht` (e.g. RF, VQSR, AS_VQSR). Default is "RF".
     :param problematic_regions: List of regions considered problematic to run filter check in. Default is ["lcr", "segdup", "nonpar"].
     :param single_filter_count: If True, explode the Table's filter column and give a supplement total count of each filter. Default is False.
-    :param monoallelic_expr: Optional boolean expression of monoallelic status that logs how many monoallelic sites are in the Table.
+    :param monoallelic_expr: When passed, either a single boolean expression or dictionary of strings and boolean expressions to log how many monoallelic sites are in the Table. Default is None.
```
ch-kr, line 257:
```suggestion
    :param monoallelic_expr: Optional boolean expression or dictionary of strings and boolean expressions used to log how many monoallelic sites are in `t`.
```
minor rewording

gnomad/assessment/validity_checks.py
```diff
@@ -899,7 +948,7 @@ def validate_release_t(
     :param subsets: List of subsets to be checked.
     :param pops: List of pops within main callset.
     :param missingness_threshold: Upper cutoff for allowed amount of missingness. Default is 0.5.
-    :param monoallelic_expr: When passed, log how many monoallelic sites are in the Table.
+    :param monoallelic_expr: When passed, either a single boolean expression or dictionary of strings and boolean expressions to log how many monoallelic sites are in the Table. Default is None.
```
ch-kr, line 951:
```suggestion
    :param monoallelic_expr: Optional boolean expression or dictionary of strings and boolean expressions used to log how many monoallelic sites are in `t`.
```

gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +863,53 @@ def vcf_field_check(
     return True
 
 
+def compare_related_global_and_row_lengths(
```
ch-kr, line 866: this minor, but I read this as applying only to the relatedness fields initially. what about "compare_relevant_global_and_row_lengths"? or "compare_global_and_row_annot_lengths"?

gnomad/assessment/validity_checks.py
```diff
@@ -919,8 +968,18 @@ def validate_release_t(
     :param samples_sum_check: When True, runs the sum_group_callstats method. Default is True.
     :param sex_chr_check: When True, runs the check_sex_chr_metricss method. Default is True.
     :param missingness_check: When True, runs the compute_missingness method. Default is True.
+    :param pprint_globals: When True, Pretty Print the globals of the input Table. Default is True.
+    :param len_comp_globals_rows: When passed, a dictionary of globals (keys) and rows (values) to check that the length of the global key is equal to the length of the row value. Default is None.
```
ch-kr, line 972:
```suggestion
    :param len_comp_globals_rows: Optional dictionary of globals (keys) and rows (values) to be checked. When passed, function checks that the lengths of the global and row annotations are equal.
```

gnomad/assessment/validity_checks.py
```diff
@@ -919,8 +968,18 @@ def validate_release_t(
...
     :return: None (stdout display of results from the battery of validity checks).
     """
+    if pprint_globals:
+        logger.info("GLOBALS OF INPUT TABLE:")
+        pprint_global_anns(t)
+
+    if len_comp_globals_rows is not None:
+        logger.info("COMPARE GLOBAL ANNOTATIONS' LENGHTS TO ROW ANOTATIONS:")
```
ch-kr, line 980:
```suggestion
        logger.info("COMPARE GLOBAL ANNOTATIONS' LENGTHS TO ROW ANNOTATIONS:")
```

gnomad/assessment/validity_checks.py
ch-kr: or I guess maybe `t` should be "Table" for consistency

ch-kr: what about renaming monoallelic_expr if it's going to be used for monoallelic and only_het? I don't have a good suggestion though, the best I could think of is "unexpected_gt_expr"

ch-kr (approved):
minor suggestions but logic LGTM!

I also noticed this in the checks
> gnomad_qc/v4/create_release/create_combined_faf_release_ht.py:27:0: E0401: Unable to import 'statsmodels.stats.contingency_tables' (import-error)

not sure if that's an issue or is already taken care of elsewhere

gnomad/assessment/validity_checks.py
mike-w-wilson: Yeah -- I was going to do something along those lines but figured thats a breaking change and choose slightly confusing over breaking change but there will probably be many breaking changes in the next release and this is pretty minor so `unexpected_gt_expr` works for me.

gnomad/assessment/validity_checks.py
mike-w-wilson: Good point I think I favor "compare_global_and_row_annot_lengths"

gnomad/assessment/validity_checks.py
mike-w-wilson: Yeah I like that better

gnomad/assessment/validity_checks.py
mike-w-wilson: `single_call_site_expr`?

mike-w-wilson: Though I guess that would limit the utility to these two cases

ch-kr: good point about the breaking change, but I guess we have a few of those and will need to make a pypi release in the near future

what about `mono_call_site_expr`? is that too close to monoallelic

gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +863,53 @@ def vcf_field_check(
...
+    pprint(global_pprint, sort_dicts=False)
+
+
 def validate_release_t(
     t: Union[hl.MatrixTable, hl.Table],
     subsets: List[str] = [""],
     pops: List[str] = POPS[CURRENT_MAJOR_RELEASE],
     missingness_threshold: float = 0.5,
-    monoallelic_expr: Optional[hl.expr.BooleanExpression] = None,
+    monoallelic_expr: Optional[
```
mike-w-wilson, line 910: Sorry @ch-kr , I committed suggestions but will update this and L257 above to whatever we decide. You suggested `unexpected_gt_expr` and I put forward `same_call_site_expr` but mine limits it to these two uses, mono and only het.

mike-w-wilson: Also my reasoning for not renaming originally was that it will be a breaking change but seems like a pretty minor one.

ch-kr: I'm bad at naming things 😆 what about `gt_check_expr`? is that too general? otherwise `same_call_site_expr` works for me since we don't have another use for this field other than monoallelic or only het yet (though good point that renaming again in the future would be another breaking change)

gnomad/assessment/validity_checks.py
```diff
@@ -267,12 +270,15 @@ def summarize_variant_filters(
...
-        logger.info("There are %d monoallelic sites in the dataset.", mono_sites)
+    if site_gt_check_expr is not None:
+        if isinstance(site_gt_check_expr, hl.expr.BooleanExpression):
+            site_gt_check_expr = {"site GT check": site_gt_check_expr}
+        for k, m_expr in site_gt_check_expr.items():
+            if isinstance(t, hl.MatrixTable):
+                gt_check_sites = t.filter_rows(m_expr).count_rows()
+            else:
+                gt_check_sites = t.filter(m_expr).count()
+            logger.info("There are %d %s sites in the dataset.", gt_check_sites, k)
```
mike-w-wilson, line 281: This is weird now but I didnt know what to rename it "There are %d site GT check sites in the dataset." I almost feel we should drop support for the single expression since this is a breaking change anyways? Require the dict[str, bool expr]?

ch-kr: I like that idea (requiring the dict)

jkgoodrich: I doubt many others outside our group are using the validity checks and requiring a dict makes a lot more sense

gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +863,54 @@ def vcf_field_check(
     return True
 
 
+def compare_global_and_row_annot_lengths(
```
jkgoodrich, line 866: Maybe add an option to compare all rows with a default of False. Just to make it clear it's only comparing the first row to the globals and with support to compare all if wanted.

gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +863,54 @@ def vcf_field_check(
...
+    t: Union[hl.MatrixTable, hl.Table],
+    len_comp_globals_rows: Dict[str, List[str]] = None,
```
jkgoodrich, line 868: maybe rename to something like row_to_globals_check

gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +863,54 @@ def vcf_field_check(
...
+) -> None:
+    """
+    Compare the lengths of global and row annotations.
```
jkgoodrich, line 871:
```suggestion
    Check that the lengths of row annotations match the lengths of associated global annotations.
```

gnomad/assessment/validity_checks.py
jkgoodrich: Maybe `check_global_and_row_annot_lengths`? or `confirm_global_and_row_annot_lengths`

gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +863,54 @@ def vcf_field_check(
...
+
+    :param t: Input MatrixTable or Table.
+    :param len_comp_globals_rows: Dictionary with row annotation (key) and list of associated global annotations (value) to compare.
```
jkgoodrich, line 874:
```suggestion
    :param len_comp_globals_rows: Dictionary with row annotation (key) and list of associated global annotations (value) to check for consistent lengths.
```

gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +863,54 @@ def vcf_field_check(
...
+    pprint(global_pprint, sort_dicts=False)
+
+
 def validate_release_t(
     t: Union[hl.MatrixTable, hl.Table],
     subsets: List[str] = [""],
     pops: List[str] = POPS[CURRENT_MAJOR_RELEASE],
     missingness_threshold: float = 0.5,
-    monoallelic_expr: Optional[hl.expr.BooleanExpression] = None,
+    site_gt_check_expr: Optional[
```
jkgoodrich, line 911: Same thought here to just force a dict

gnomad/assessment/validity_checks.py
```diff
@@ -919,8 +969,18 @@ def validate_release_t(
...
+    :param pprint_globals: When True, Pretty Print the globals of the input Table. Default is True.
+    :param len_comp_globals_rows: Optional dictionary of globals (keys) and rows (values) to be checked. When passed, function checks that the lengths of the global and row annotations are equal.
     :return: None (stdout display of results from the battery of validity checks).
     """
+    if pprint_globals:
+        logger.info("GLOBALS OF INPUT TABLE:")
+        pprint_global_anns(t)
+
+    if len_comp_globals_rows is not None:
+        logger.info("COMPARE GLOBAL ANNOTATIONS' LENGHTS TO ROW ANOTATIONS:")
```
jkgoodrich, line 981:
```suggestion
        logger.info("COMPARE GLOBAL ANNOTATIONS' LENGTHS TO ROW ANNOTATIONS:")
```

jkgoodrich (commented):
Just a few more small thoughts

mike-w-wilson:
Backto you @jkgoodrich !

gnomad/assessment/validity_checks.py
```diff
@@ -857,12 +859,68 @@ def vcf_field_check(
...
+                " globals (%d) does %smatch length of %s in %d out of %d rows (%d)",
+                outcome,
+                global_field,
+                global_len,
+                "NOT " if outcome == "Failed" else "",
+                row_field,
+                failed_rows if outcome == "Failed" else len(rows),
+                len(rows),
+                row_len,
+            )
```
jkgoodrich, line 905:
```suggestion
    for row_field, global_fields in row_to_globals_check.items():
        if not check_all_rows:
            logger.info(
                "Checking length of %s in first row against length of globals: %s",
                row_field,
                global_fields,
            )
        for global_field in global_fields:
            global_len = hl.eval(hl.len(t[global_field]))
            row_len_expr = hl.len(t[row_field])
            failed_rows = t.aggregate(
                hl.struct(
                    n_fail=hl.agg.count_where(row_len_expr != global_len),
                    row_len=hl.agg.counter(row_len_expr),
                )
            )
            outcome = "Failed" if failed_rows["n_fail"] > 0 else "Passed"
            n_rows = t.count()
            logger.info(
                "%s global and row lengths comparison: Length of %s in"
                " globals (%d) does %smatch length of %s in %d out of %d rows (%s)",
                outcome,
                global_field,
                global_len,
                "NOT " if outcome == "Failed" else "",
                row_field,
                failed_rows["n_fail"] if outcome == "Failed" else n_rows,
                n_rows,
                failed_rows["row_len"],
            )
```

jkgoodrich (changes requested):
Just a little change in the handling of all rows

gnomad/assessment/validity_checks.py
mike-w-wilson: Yes, this is way better

jkgoodrich (approved):
LGTM!

PR #633, opened 2023-10-20
Change `get_downsampling_freq_indices` and `downsampling_counts_expr` to support both 'pop' and 'gen_anc' keys in metadata

jkgoodrich:
Needed for some plots I am creating that use these two functions on the v4 data

### DIFF
--- a/gnomad/utils/constraint.py
+++ b/gnomad/utils/constraint.py
@@ -229,7 +229,7 @@ def count_variants_by_group(
 
 def get_downsampling_freq_indices(
     freq_meta_expr: hl.expr.ArrayExpression,
-    gen_anc: str = "global",
+    pop: str = "global",
     variant_quality: str = "adj",
     genetic_ancestry_label: Optional[str] = None,
     subset: Optional[str] = "None",
@@ -241,7 +241,7 @@ def get_downsampling_freq_indices(
     :param freq_meta_expr: ArrayExpression containing the set of groupings for each
         element of the `freq_expr` array (e.g., [{'group': 'adj'}, {'group': 'adj',
         'pop': 'nfe'}, {'downsampling': '5000', 'group': 'adj', 'pop': 'global'}]).
-    :param gen_anc: Genetic ancestry group to use for filtering by the `genetic_ancestry_label` key in
+    :param pop: Population to use for filtering by the `genetic_ancestry_label` key in
         `freq_meta_expr`. Default is 'global'.
     :param variant_quality: Variant quality to use for filtering by the 'group' key in
         `freq_meta_expr`. Default is 'adj'.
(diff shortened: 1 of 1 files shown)


gnomad/utils/constraint.py
```diff
@@ -230,21 +230,32 @@ def get_downsampling_freq_indices(
     freq_meta_expr: hl.expr.ArrayExpression,
     pop: str = "global",
     variant_quality: str = "adj",
+    genetic_ancestry_label: Optional[str] = None,
 ) -> hl.expr.ArrayExpression:
```
klaricch, line 234: should there also be an option to not pull the non_ukb subset?

jkgoodrich:
This is back to me, but is lower in priority than others.

gnomad/utils/constraint.py
```diff
@@ -257,6 +263,8 @@ def get_downsampling_freq_indices(
         lambda f: (f[1].get("group") == variant_quality)
         & (hl.any([f[1].get(l, "") == pop for l in gen_anc]))
         & f[1].contains("downsampling")
+        & hl.literal(downsamplings).contains(hl.int(f[1].get("downsampling", "0")))
+        & (f[1].get("subset", "None") == subset)
     )
```
jkgoodrich, line 268: I can't remove the one line, so that will also need to be removed for this suggestion to work


```suggestion
    def _get_filter_expr(m: hl.expr.StructExpression) -> hl.expr.BooleanExpression:
        filter_expr = (
            (m.get("group") == variant_quality)
            & (hl.any([m.get(l, "") == pop for l in gen_anc]))
            & m.contains("downsampling")
        )

        if downsamplings is not None:
            filter_expr &= hl.literal(downsamplings).contains(
                hl.int(m.get("downsampling", "0"))
            )

        if subset is None:
            filter_expr &= ~m.contains("subset")
        else:
            filter_expr &= m.get("subset", "") == subset

        return filter_expr

    indices = hl.enumerate(freq_meta_expr).filter(lambda f: _get_filter_expr(f[1]))
```

gnomad/utils/constraint.py
```diff
@@ -270,6 +278,8 @@ def downsampling_counts_expr(
     singleton: bool = False,
     max_af: Optional[float] = None,
     genetic_ancestry_label: Optional[str] = None,
+    subset: Optional[str] = "None",
+    downsamplings: list = DOWNSAMPLINGS["v4"],
```
jkgoodrich, line 282:
```suggestion
    subset: Optional[str] = None,
    downsamplings: Optional[List[int]] = None,
```

gnomad/utils/constraint.py
```diff
@@ -294,12 +304,20 @@ def downsampling_counts_expr(
     :param genetic_ancestry_label: Label defining the genetic ancestry groups. If None,
         "gen_anc" or "pop" is used (in that order of preference) if present. Default is
         None.
+    :param subset: Subset to use for filtering by the 'subset' key in
+        `freq_meta_expr`. Default is "None".
+    :param downsamplings: List of strings specifying what levels of downsamplings to obtain.
```
jkgoodrich, line 309:
```suggestion
    :param subset: Subset to use for filtering by the 'subset' key in `freq_meta_expr`.
        Default is None, which will return all downsampling counts without a 'subset'
        key in `freq_meta_expr`. If specified, only downsamplings with the specified
        subset will be included.
    :param downsamplings: Optional List of integers specifying what downsampling
        indices to obtain. Default is None, which will return all downsampling counts.
```

jkgoodrich (commented):
LGTM

klaricch (approved):

PR #635, opened 2023-10-23
Add optional parameters to `set_female_y_metrics_to_na_expr` to use other frequency fields

jkgoodrich:


gnomad/utils/annotations.py
```diff
@@ -1046,31 +1046,45 @@ def missing_callstats_expr() -> hl.expr.StructExpression:
...
+        freq_expr = t[freq_expr]
+    if isinstance(freq_meta_expr, str):
+        freq_meta_expr = t[freq_meta_expr]
+    if isinstance(freq_index_dict_expr, str):
+        freq_index_dict_expr = t[freq_index_dict_expr]
 
-    :param t: Table or MatrixTable for which to adjust female metrics
-    :return: Hail array expression to set female Y-variant metrics to missing values
-    """
     female_idx = hl.map(
```
ch-kr, line 1073: maybe we should make this 
```
    xx_idx = hl.map(
```
since we've been moving away from using gender terms for our sex assignments (though I guess maybe we should rename the function also, which is also a breaking change)

ch-kr (approved):
a couple comments that aren't urgent; changes LGTM

gnomad/utils/annotations.py
```diff
@@ -1046,31 +1046,45 @@ def missing_callstats_expr() -> hl.expr.StructExpression:
 
 
 def set_female_y_metrics_to_na_expr(
-    t: Union[hl.Table, hl.MatrixTable]
+    t: Union[hl.Table, hl.MatrixTable],
```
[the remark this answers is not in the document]
jkgoodrich: yeah, but I don't feel like going through an changing the things that we already ran this release that use that, I think it's cleaner to just leave as is for the next pypi release

gnomad/utils/annotations.py
jkgoodrich: I think this is a post v4 change, we can do an overhaul for that and population

PR #639, opened 2023-10-26
Check for skip_coverage_model is False in build_models

klaricch:


gnomad/utils/constraint.py
```diff
@@ -597,7 +597,7 @@ def build_models(
             hl.struct(**plateau_models_agg_expr)
         )
 
-    if skip_coverage_model:
+    if skip_coverage_model is False:
```
jkgoodrich, line 600:
```suggestion
    if not skip_coverage_model:
```

jkgoodrich (changes requested):

jkgoodrich (approved):

PR #640, opened 2023-10-27
v4 VCF export updates

mike-w-wilson:
This adds v4 export functionality through a number of fields like `inbreeding_coeff`, new annotations like `grpmax` and `fafmax`, and a `suffix` arg to the info dictionary construction. I tried my best to avoid any breaking change here but the histogram access is one. This is a minor function IMO so I think it is alright and  have assign eda change log ignore label.

### DIFF
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -1329,7 +1329,7 @@ def build_vcf_export_reference(
         ],
     }
     if keep_chrM:
-        ref_args["contigs"].append("chrM")
+        ref_args["contigs"].append("M")
         ref_args.update({"mt_contigs": ref.mt_contigs})
 
     export_reference = hl.ReferenceGenome(**ref_args)
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -1197,6 +1197,7 @@ def make_hist_dict(
     adj: bool,
     hist_metric_list: List[str] = HISTS,
     label_delimiter: str = "_",
+    drop_n_smaller_larger: bool = False,
 ) -> Dict[str, str]:
     """
     Generate dictionary of Number and Description attributes to be used in the VCF header, specifically for histogram annotations.
@@ -1205,6 +1206,7 @@ def make_hist_dict(
     :param adj: Whether to create a header dict for raw or adj quality histograms.
     :param hist_metric_list: List of hists for which to build hist info dict
     :param label_delimiter: String used as delimiter in values stored in hist_metric_list.
+    :param drop_n_smaller_larger: Whether to drop n_smaller and n_larger annotations from header dict. Default is False.
     :return: Dictionary keyed by VCF INFO annotations, where values are Dictionaries of Number and Description attributes.
     """
     header_hist_dict = {}
@@ -1228,21 +1230,41 @@ def make_hist_dict(
                 "Number": "A",
                 "Description": f"Histogram for {hist_text}; bin edges are: {edges}",
             },
-            f"{hist}_n_smaller": {
-                "Number": "A",
-                "Description": (
-                    f"Count of {hist_fields[0].upper()} values falling below lowest"
-                    f" histogram bin edge {hist_text}"
-                ),
-            },
-            f"{hist}_n_larger": {
-                "Number": "A",
-                "Description": (
-                    f"Count of {hist_fields[0].upper()} values falling above highest"
-                    f" histogram bin edge {hist_text}"
-                ),
-            },
         }
+        # These annotations are frequently zero and are dropped from gnomad
+        # releases for most histograms
+        if not drop_n_smaller_larger:
+            hist_dict.update(
+                {
+                    f"{hist}_n_smaller": {
+                        "Number": "A",
+                        "Description": (
+                            f"Count of {hist_fields[0].upper()} values falling below"
+                            f" lowest histogram bin edge {hist_text}"
+                        ),
+                    },
+                    f"{hist}_n_larger": {
+                        "Number": "A",
+                        "Description": (
+                            f"Count of {hist_fields[0].upper()} values falling above"
+                            f" highest histogram bin edge {hist_text}"
+                        ),
+                    },
+                }
+            )
+        # Only add n_larger for dp qual histograms
+        if "dp" in hist:
+            hist_dict.update(
+                {
+                    f"{hist}_n_larger": {
+                        "Number": "A",
+                        "Description": (
+                            f"Count of {hist_fields[0].upper()} values falling above"
+                            f" highest histogram bin edge {hist_text}"
+                        ),
+                    },
+                }
+            )
 
         header_hist_dict.update(hist_dict)
 
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -274,6 +274,12 @@ INFO_DICT = {
             " cohort)"
         )
     },
+    "sibling_singleton": {
+        "Description": (
+            "Variant was a callset-wide doubleton that was present only in two siblings"
+            " (i.e., a singleton amongst unrelated samples in cohort)."
+        )
+    },
     "original_alleles": {"Description": "Alleles before splitting multiallelics"},
     "variant_type": {
         "Description": "Variant type (snv, indel, multi-snv, multi-indel, or mixed)"
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -1299,8 +1299,8 @@ def set_female_y_metrics_to_na(
 def build_vcf_export_reference(
     name: str,
     build: str = "GRCh38",
-    keep_contigs: List[str] = [f"chr{i}" for i in range(1, 23)]
-    + ["chrX", "chrY", "chrM"],
+    keep_contigs: List[str] = [f"chr{i}" for i in range(1, 23)] + ["chrX", "chrY"],
+    keep_chrM: bool = True,
 ) -> hl.ReferenceGenome:
     """
     Create export reference based on reference genome defined by `build`.
@@ -1311,23 +1311,28 @@ def build_vcf_export_reference(
 
     :param name: Name to use for new reference.
     :param build: Reference genome build to use as starting reference genome.
-    :param keep_contigs: Contigs to keep from reference genome defined by `build`. Default is autosomes, sex chromosomes, and chrM.
+    :param keep_contigs: Contigs to keep from reference genome defined by `build`. Default is autosomes and sex chromosomes.
+    :param keep_chrM: Whether to keep chrM. Default is True.
     :return: Reference genome for VCF export containing only contigs in `keep_contigs`.
     """
     ref = hl.get_reference(build)
 
-    export_reference = hl.ReferenceGenome(
-        name=name,
-        contigs=keep_contigs,
-        lengths={contig: ref.lengths[contig] for contig in keep_contigs},
-        x_contigs=ref.x_contigs,
-        y_contigs=ref.y_contigs,
-        par=[
+    ref_args = {
+        "name": name,
+        "contigs": keep_contigs,
+        "lengths": {contig: ref.lengths[contig] for contig in keep_contigs},
+        "x_contigs": ref.x_contigs,
+        "y_contigs": ref.y_contigs,
+        "par": [
             (interval.start.contig, interval.start.position, interval.end.position)
             for interval in ref.par
         ],
-        mt_contigs=ref.mt_contigs,
-    )
+    }
+    if keep_chrM:
+        ref_args["keep_contigs"].append("chrM")
+        ref_args.update({"mt_contigs": ref.mt_contigs})
+
+    export_reference = hl.ReferenceGenome(**ref_args)
 
     return export_reference
 
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -878,7 +878,7 @@ def make_info_dict(
             },
         }
         info_dict.update(popmax_dict)
-    elif grpmax:
+    if grpmax:
         grpmax_dict = {
             f"{prefix}grpmax{suffix}": {
                 "Number": "A",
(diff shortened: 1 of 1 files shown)


gnomad/utils/vcf.py
```diff
@@ -180,6 +189,18 @@
             " variants for VQSR"
         )
     },
+    "positive_train_site": {
```
klaricch, line 192: do POSITIVE_TRAIN_SITE and NEGATIVE_TRAIN_SITE above need to be deleted?

gnomad/utils/vcf.py
```diff
@@ -207,7 +228,25 @@
...
+            "Less than 85 percent of samples meet 20X coverage if variant is in"
+            " autosomal or PAR region or 10X coverage for non-PAR regions of"
+            " chromosomes X and Y."
+        )
+    },
+    "outside_ukb_capture_region": {
+        "Description": "Variant falls outside of UK Biobank exome capture regions."
+    },
+    "outside_broad_capture_region": {
+        "Variant falls outside of Broad exome capture regions."
```
klaricch, line 248: add  "Description":

gnomad/utils/vcf.py
```diff
@@ -297,31 +343,56 @@
...
+            " assigned negative scores) as well as conservation (slower than expected"
+            " evolution, assigned positive scores)."
+        ),
+    },
+    "sift_max": {
+        "Number": "1",
+        "Description": (
+            "Score reflecting the scaled probability of the amino acid substitution"
+            " being tolerated, ranging from 0 to 1. Scores below 0.05 are predicted to"
+            " impact protein function. We prioritize max scores for MANE Select"
```
klaricch, line 383: are we saying "MANE Select" or "MANE SELECT" (like in the description for revel_max?)

gnomad/utils/vcf.py
```diff
@@ -692,41 +767,46 @@ def make_info_dict(
     :param bin_edges: Dictionary keyed by annotation type, with values that reflect the bin edges corresponding to the annotation.
     :param faf: If True, use alternate logic to auto-populate dictionary values associated with filter allele frequency annotations.
     :param popmax: If True, use alternate logic to auto-populate dictionary values associated with popmax annotations.
+    :param grpmax: If True, use alternate logic to auto-populate dictionary values associated with grpmax annotations.
+    :param fafmax: If True, use alternate logic to auto-populate dictionary values associated with fafmax annotations.
+    :param combined_fields: If True, use alternate logic to auto-populate dictionary values associated with combined annotations.
```
klaricch, line 772: combined_fields isn't listed in the def() part

gnomad/utils/vcf.py
```diff
@@ -796,6 +876,85 @@ def make_info_dict(
...
+                ),
+            },
+            f"{prefix}nhomalt{label_delimiter}grpmax{suffix}": {
+                "Number": "A",
+                "Description": (
+                    "Count of homozygous individuals in the genetic ancestry group"
+                    f" with the maximum allele frequency{description_text}"
+                ),
+            },
+            f"{prefix}faf95{label_delimiter}grpmax{suffix}": {
```
klaricch, line 916: i might be misremembering but i thought we dropped this?

gnomad/utils/vcf.py
```diff
@@ -796,6 +876,85 @@ def make_info_dict(
...
+        grpmax_dict = {
+            f"{prefix}grpmax{suffix}": {
+                "Number": "A",
+                "Description": (
+                    "Genetic ancestry group with the maximum allele"
+                    f" frequency{description_text}"
+                ),
+            },
+            f"{prefix}AC{label_delimiter}grpmax{suffix}": {
+                "Number": "A",
```
klaricch, line 889: do we just use Number A for everything even if only one value is expected?

klaricch (changes requested):

gnomad/utils/vcf.py
mike-w-wilson: I left it as to not break any pipeline that looks in the dictionary for it, like v3

gnomad/utils/vcf.py
mike-w-wilson: Oops should be MANE Select I think

gnomad/utils/vcf.py
mike-w-wilson: We did but v3 has it so I'm leaving it in so not to break that pipeline. I see this as more of a collection to pick from and modify rather than what exists in the current version

gnomad/utils/vcf.py
mike-w-wilson: Honestly I just kept what was already here for popmax.

gnomad/utils/vcf.py
```diff
@@ -297,31 +343,56 @@
...
         ),
     },
-    "revel_score": {
+    "revel_max": {
         "Number": "1",
         "Description": (
-            "dbNSFP's Revel score from 0 to 1. Variants with higher scores are"
+            "The maximum dbNSFP's Revel score at a site's MANE Select or canonical"
+            " transcript. Scores ranges from 0 to 1. Variants with higher scores are"
             " predicted to be more likely to be deleterious."
```
KoalaQin, line 351: Based on my understanding, REVEL is independent of dbNSFP. Also I downloaded directly from REVEL's website. 
dbNSFP compiles all kinds of prediction scores. 
```suggestion
            "The maximum REVEL score at a site's MANE Select or canonical transcript."
             "It's an ensemble score for predicting the pathogenicity of missense "
             "variants (based on 13 other variant predictors). Scores ranges from 0 to 1. "
             "Variants with higher scores are predicted to be more likely to be "
             "deleterious."
```

gnomad/utils/vcf.py
```diff
@@ -684,6 +758,7 @@ def make_info_dict(
         - INFO fields for filtering allele frequency (faf) annotations
 
     :param prefix: Prefix string for data, e.g. "gnomAD". Default is empty string.
+    :param suffix: Suffix string for data, e.g. "gnomAD". Default is empty string.
```
KoalaQin, line 761: suffix is usually gnomAD too? Or do you have another example?

gnomad/utils/vcf.py
```diff
@@ -796,6 +875,85 @@ def make_info_dict(
...
+                "Number": "A",
+                "Description": (
+                    "Filtering allele frequency (using Poisson 95% CI) for the genetic"
+                    " ancestry group with the maximum allele"
+                    f" frequency{description_text}"
+                ),
+            },
+        }
+        info_dict.update(grpmax_dict)
+    elif fafmax:
```
KoalaQin, line 925: Should it be `if fafmax:`? I thought `popmax` is degraded, that's why we're copying what it has to `grpmax`, but `grpmax` and `fafmax` coexist, no?

gnomad/utils/vcf.py
mike-w-wilson: Though from the docs, I think this makes sense 
```
• A: The field has one value per alternate allele. The values must be in the same order as listed in the ALT
column (described in section 1.6).
• R: The field has one value for each possible allele, including the reference. The order of the values must be the
reference allele first, then the alternate alleles as listed in the ALT column.
```
Since grpmax is the highest alt allele frequency

gnomad/utils/vcf.py
```diff
@@ -960,7 +1118,9 @@ def make_vcf_filter_dict(
                 " < 20; DP < 10; and AB < 0.2 for het calls)"
             )
         },
-        "InbreedingCoeff": {"Description": f"InbreedingCoeff < {inbreeding_cutoff}"},
+        "InbreedingCoeff": {
+            "Description": f"Inbreeding coefficient < {inbreeding_cutoff}"
+        },
```
KoalaQin, line 1123: Do we need to add `as_vqsr' too? I just saw them in the validity check.

KoalaQin (changes requested):

gnomad/utils/vcf.py
mike-w-wilson: It can be "gnomad" or "non_ukb", would you prefer non_ukb here?

gnomad/utils/vcf.py
mike-w-wilson: So this is a weird conditional structure I tried to update for v4 but I'm going to revisit this. It is built this way because we call each metric one by one but that is kinda weird. I'll update it again

gnomad/utils/vcf.py
KoalaQin: ah, you mean we can put "gnomad" and "non_ukb" either as a prefix or a suffix, if that's the case, it's fine.

gnomad/utils/vcf.py
mike-w-wilson: AS_VQSR is on L1100

gnomad/utils/vcf.py
```diff
@@ -235,6 +274,12 @@
             " cohort)"
         )
     },
+    "sibling_singleton": {
+        "Description": (
+            "Variant was a callset-wide doubleton that was present only in two siblings"
+            " (i.e., a singleton amongst unrelated samples in cohort)."
```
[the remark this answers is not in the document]
mike-w-wilson: I took this from this document so the field would match https://docs.google.com/document/d/1c4hmk7KS2LhwhjPXQ0sae7yloLQQ8H2J2J84o1HVZt4/edit

gnomad/utils/vcf.py
```diff
@@ -796,8 +878,93 @@ def make_info_dict(
...
+    if fafmax:
+        fafmax_dict = {
+            f"{prefix}fafmax{label_delimiter}faf95{label_delimiter}max{suffix}": {
+                "Number": "A",
+                "Description": (
+                    "Maximum filtering allele frequency (using Poisson 95% CI)"
+                    f" across genetic_ancestry groups{description_text}"
+                ),
+            },
+            f"{prefix}fafmax{label_delimiter}faf95{label_delimiter}max{label_delimiter}gen{label_delimiter}anc{suffix}": {
```
KoalaQin, line 930: Just for consistency, I saw below, you used `data_type`, why this `gen_anc` has to used  `{label_delimiter}`?

gnomad/utils/vcf.py
```diff
@@ -796,8 +878,93 @@ def make_info_dict(
...
+                ),
+            },
+            f"{prefix}fafmax{label_delimiter}faf99{label_delimiter}max{suffix}": {
+                "Number": "A",
+                "Description": (
+                    "Maximum filtering allele frequency (using Poisson 99% CI)"
+                    f" across genetic_ancestry groups{description_text}"
+                ),
+            },
+            f"{prefix}fafmax{label_delimiter}faf99{label_delimiter}max{label_delimiter}gen{label_delimiter}anc{suffix}": {
```
KoalaQin, line 944: Same comment about `gen{label_delimiter}anc`.

gnomad/utils/vcf.py
mike-w-wilson: Yeah I think data type should use the delimiter too -- ill fix it

gnomad/utils/vcf.py
```diff
@@ -144,6 +145,14 @@
...
         ),
     },
+    "inbreeding_coeff": {
+        "Number": "A",
+        "Description": (
+            "Inbreeding coefficient, the excess heterozygosity at a variant site,"
+            " computed as 1 - (the number of heterozygous genotypes)/(the number of"
+            " heterozygous genotypes expected under Hardy-Weinberg equilibrium)"
+        ),
+    },
```
KoalaQin, line 155: after L166, you should add "SB", because it's in SITE_FIELDS as in L79. Something like this? 
```
"SB": {
        "Description": (
            "Per-sample component statistics which comprise the Fisher's exact test to detect strand bias. Values are: depth of reference allele on forward strand, depth of reference allele on reverse strand, depth of alternate allele on forward strand, depth of alternate allele on reverse strand."
        )
    },
```

KoalaQin: I'm not sure what number=?, but the type is not Integar for sure, has to be float, decided directly by hail export, right?

mike-w-wilson: Yes -- I noted this in the notebook and slack. I'm trying to find the correct definition, it is not per-sample.

KoalaQin: Sorry, I misread the numbers yesterday, they are integers, and 4 values.
I found this GATK definition might helps: 
https://gatk.broadinstitute.org/hc/en-us/articles/4409917393307-StrandBiasBySample

KoalaQin: Ah, Okay, I found it's defined in FORMAT_dict, L516-L524, but we don't have per-sample entries, I guess it's a sum of REF/ALT read counts across all the samples to make this SB in the INFO_dict (SITE_FIELDS)?

klaricch (changes requested):

gnomad/utils/vcf.py
mike-w-wilson: We're dropping SB from the VCF so not going to add the definition here.

gnomad/utils/vcf.py
```diff
@@ -1321,23 +1311,28 @@ def build_vcf_export_reference(
...
+        "y_contigs": ref.y_contigs,
+        "par": [
             (interval.start.contig, interval.start.position, interval.end.position)
             for interval in ref.par
         ],
-        mt_contigs=ref.mt_contigs,
-    )
+    }
+    if keep_chrM:
+        ref_args["contigs"].append("M")
```
klaricch, line 1332: why M here? should this be ref_args["contigs"].extend(ref.mt_contigs) to be compatible with multiple builds?

mike-w-wilson: Yes that is better

klaricch (approved):
lgtm!

mike-w-wilson:
@KoalaQin , is there anything remaining youd like to see addressed? Well be dropping SB from the VCF so that is no longer needed.

KoalaQin (approved):
I looked over your conversation and the code changes about "chrM", thanks for making it work. LGTMT!

PR #642, opened 2023-10-31
Update resource paths

klaricch:


gnomad/resources/grch38/gnomad.py
```diff
@@ -395,7 +395,7 @@ def coverage_tsv_path(data_type: str, version: Optional[str] = None) -> str:
     if data_type == "exomes":
         if version is None:
             version = CURRENT_EXOME_RELEASE
```
mike-w-wilson, line 397: Should this be updated to `CURRENT_EXOME_COVERAGE_RELEASE` to match genomes?

mike-w-wilson (changes requested):
Just one thing!

gnomad/resources/grch38/gnomad.py
klaricch: yes, thank you!

mike-w-wilson (approved):
:shipit:

mike-w-wilson:
I'm not sure if its helpful, but to get around the error youre seeing in the check, I went back to python 3.8.6 and hail 02.1.20. I'm not entirely sure which fixed it but I think its python and hail is just importing from it?

jkgoodrich:
I think [this](https://github.com/broadinstitute/gnomad_methods/pull/642/commits/4c2cf620b6924d1e2ec093f880406362f946292a) commit will fix that error

PR #643, opened 2023-10-31
Drop async file exists function

mike-w-wilson:
This is function is unused within the group and is breaking imports after hail 0.2.126.


mike-w-wilson:
I wasn't sure if we consider this a bug fix or just ignore it?

jkgoodrich (approved):
LGTM!

PR #645, opened 2023-11-08
add ensg to groupings

klaricch:


ch-kr (approved):
one quick question: does this grouping need to be added here, or can this be added just to the final step to create the release HT? but change LGTM

klaricch:
> one quick question: does this grouping need to be added here, or can this be added just to the final step to create the release HT? but change LGTM

It will be dropped from the input that goes into the last step of the constraint pipeline to create the release ht unless it's added here

PR #647, opened 2023-11-13
Add RSEM summary function 

jkgoodrich:
Starting to move over generic functions from https://github.com/macarthur-lab/tx_annotation/tree/bea56ddf991607f7479358c0c80588f1a2ec9bc1#applying-transcript-expression-aware-annotation-to-your-own-dataset

This function can take in an expression MT and summarize by transcript and tissue, similar to [this](https://github.com/macarthur-lab/tx_annotation/blob/bea56ddf991607f7479358c0c80588f1a2ec9bc1/tx_annotation_resources.py#L125) function after the loading of the resource.

### DIFF
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -18,10 +18,8 @@ def summarize_rsem_mt(
     function to use to summarize the expression by tissue. By default, the median is
     used.
 
-    .. note::
-
-        The output can be returned in one of the following formats (both keyed by
-        "transcript_id" and "gene_id"):
+    The output can be returned in one of the following formats (both keyed by
+    "transcript_id" and "gene_id"):
 
         - A Table with an 'rsem' field containing an array of summarized expression
           values by tissue, where the order of tissues in the array is indicated by
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -14,7 +14,9 @@ logger.setLevel(logging.INFO)
 
 def summarize_transcript_expression(
     mt: hl.MatrixTable,
-    transcript_expression_expr: Union[hl.expr.NumericExpression, str] = "x",
+    transcript_expression_expr: Union[
+        hl.expr.NumericExpression, str
+    ] = "transcript_tpm",
     tissue_expr: Union[hl.expr.StringExpression, str] = "tissue",
     summary_agg_func: Optional[Callable] = None,
 ) -> Tuple[hl.Table, hl.Table]:
@@ -30,9 +32,9 @@ def summarize_transcript_expression(
 
     :param mt: MatrixTable of transcript (rows) expression quantifications (entry) by
         sample (columns).
-    :param tissue_expr: Column expression indicating tissue type. Default is 'tissue'.
     :param transcript_expression_expr: Entry expression indicating transcript expression
-        quantification. Default is 'x'.
+        quantification. Default is 'transcript_tpm'.
+    :param tissue_expr: Column expression indicating tissue type. Default is 'tissue'.
     :param summary_agg_func: Optional aggregation function to use to summarize the
         transcript expression quantification by tissue. Example: `hl.agg.mean`. Default
         is None, which will use a median aggregation.
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -43,7 +43,7 @@ def summarize_rsem_mt(
     )
 
     if tissue_as_row:
-        rsem_ht = rsem_mt.rename("rsem", "").make_table()
+        rsem_ht = rsem_mt.rename({"rsem": ""}).make_table()
     else:
         rsem_ht = rsem_mt.localize_entries(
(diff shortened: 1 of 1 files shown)


gnomad/utils/transcript_annotation.py
```diff
@@ -0,0 +1,59 @@
...
+          values by tissue, where the order of tissues in the array is indicated by
+          the "tissues" global annotation (`tissue_as_row` set to False).
+        - A Table with a row annotation for each tissue containing the summarized
+          tissue expression value (`tissue_as_row` set to True).
+
+    :param rsem_mt: MatrixTable of RSEM quantifications.
+    :param tissue_expr: Column expression indicating tissue type.
+    :param rsem_expr: Entry expression indicating RSEM quantification.
+    :param summary_agg_func: Optional aggregation function to use to summarize the RSEM
+        values by tissue. Default is None, which will use a median aggregation.
```
KoalaQin, line 36:
```suggestion
        values by tissue. Example format is 'lambda x: hl.median(hl.agg.collect(x)'. Default is None, which will use a median aggregation.
```

gnomad/utils/transcript_annotation.py
```diff
@@ -0,0 +1,59 @@
+"""Utils module containing generic functions that are useful for adding transcript expression-aware annotations."""
+from typing import Callable, Optional
+
+import hail as hl
+
+
+def summarize_rsem_mt(
+    rsem_mt: hl.MatrixTable,
+    rsem_expr: hl.expr.NumericExpression,
+    tissue_expr: hl.expr.StringExpression,
```
KoalaQin, line 10: Since rsem_expr will be 'mt.x' and tissue_expr will 'mt.tissue' in the resource mt I'm making, I'm not sure if it's still necessary to define them as arguments.

gnomad/utils/transcript_annotation.py
```diff
@@ -0,0 +1,59 @@
...
+
+        - A Table with an 'rsem' field containing an array of summarized expression
+          values by tissue, where the order of tissues in the array is indicated by
+          the "tissues" global annotation (`tissue_as_row` set to False).
+        - A Table with a row annotation for each tissue containing the summarized
+          tissue expression value (`tissue_as_row` set to True).
+
+    :param rsem_mt: MatrixTable of RSEM quantifications.
+    :param tissue_expr: Column expression indicating tissue type.
+    :param rsem_expr: Entry expression indicating RSEM quantification.
```
KoalaQin, line 34: Docstring here will need to change if we delete the 2 arguments.

KoalaQin (changes requested):
I have some comments about the 2 arguments.

gnomad/utils/transcript_annotation.py
```diff
@@ -0,0 +1,59 @@
...
+
+import hail as hl
+
+
+def summarize_rsem_mt(
+    rsem_mt: hl.MatrixTable,
+    rsem_expr: hl.expr.NumericExpression,
+    tissue_expr: hl.expr.StringExpression,
+    summary_agg_func: Optional[Callable] = None,
+    tissue_as_row: bool = False,
```
KoalaQin, line 12: While I'm writing next steps, I found that we'll have to write code every time for the case 'tissue_as_row'. Is it really useful?

gnomad/utils/transcript_annotation.py
```diff
@@ -0,0 +1,59 @@
+"""Utils module containing generic functions that are useful for adding transcript expression-aware annotations."""
+from typing import Callable, Optional
+
+import hail as hl
+
+
+def summarize_rsem_mt(
```
KoalaQin, line 7: can we also rename this to `summarize_transcript_expression`, then the field 'rsem' to 'transcript_expression', because I will have 'gene_expression' = sum of transcripts expression, then we'll take a proportion.

gnomad/utils/transcript_annotation.py
jkgoodrich: see change. I think we should only store as HT not in array, then use the other function when useful

gnomad/utils/transcript_annotation.py
```diff
@@ -0,0 +1,92 @@
...
+    format="%(asctime)s (%(name)s %(lineno)s): %(message)s",
+    datefmt="%m/%d/%Y %I:%M:%S %p",
+)
+logger = logging.getLogger("transcript_annotation_utils")
+logger.setLevel(logging.INFO)
+
+
+def summarize_transcript_expression(
+    mt: hl.MatrixTable,
+    transcript_expression_expr: Union[hl.expr.NumericExpression, str] = "x",
```
KoalaQin, line 17:
```suggestion
    transcript_expression_expr: hl.expr.NumericExpression
```
`str` won't work.

KoalaQin (changes requested):

gnomad/utils/transcript_annotation.py
jkgoodrich: oh sorry, yes it will, I just forgot to change the code to use it (I meant to...)

KoalaQin (approved):
LGTM!

PR #648, opened 2023-11-17
get gene expression HT

KoalaQin:
I add some lines to get gene expression = sum of transcript expression per tissue in the function.


jkgoodrich (commented):
Sorry, I just added this to the other PR since it was the same function I was working on. I think we can just close this one now

PR #649, opened 2023-11-17
Function to get expression proportion

KoalaQin:
This function is similar to [this](https://github.com/macarthur-lab/tx_annotation/blob/bea56ddf991607f7479358c0c80588f1a2ec9bc1/tx_annotation.py#L211C5-L211C30), by taking the proportions of transcript expression to gene expression per tissue, and a mean proportion. It's extended to use both array-format expression data and tissue-as-row expression data as input.

### DIFF
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -90,3 +90,46 @@ def tissue_expression_ht_to_array(
     ht = ht.select(tissue_expression=[ht[t] for t in tissues])
 
     return ht
+
+
+def get_expression_proportion(
+    transcript_ht: hl.Table,
+    gene_ht: hl.Table,
+    tissues_to_filter: Optional[List[str]] = None,
+) -> hl.Table:
+    """
+    Calculate the proportion of expression of transcript to gene per tissue.
+
+    :param transcript_ht: Table of summarized transcript expression by tissue.
+    :param gene_ht: Table of summarized gene expression by tissue.
+    :param tissues_to_filter: Optional list of tissues to filter out
+    :return: Table with expression proportion of transcript to gene per tissue
+        and mean expression proportion across tissues.
+    """
+    transcript_ht = tissue_expression_ht_to_array(
+        transcript_ht, tissues_to_filter=tissues_to_filter
+    )
+    gene_ht = tissue_expression_ht_to_array(
+        gene_ht, tissues=hl.eval(transcript_ht.tissues)
+    )
+
+    # Join the transcript expression table and gene expression table.
+    transcript_ht = transcript_ht.annotate(
+        gene_expression=gene_ht[transcript_ht.gene_id].tissue_expression
+    )
+
+    # Calculate the proportion of expression of transcript to gene per tissue.
+    transcript_ht = transcript_ht.annotate(
+        exp_prop=hl.or_else(
+            transcript_ht.transcript_expression / transcript_ht.gene_expression,
+            hl.empty_array(hl.tfloat64),
+        ),
+    )
+    # Calculate the mean expression proportion across tissues.
+    transcript_ht = transcript_ht.annotate(
+        exp_prop_mean=hl.mean(
+            hl.filter(lambda e: ~hl.is_nan(e), transcript_ht.exp_prop),
+        )
+    )
+
+    return transcript_ht
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -36,7 +36,7 @@ def summarize_transcript_expression(
         quantification. Default is 'transcript_tpm'.
     :param tissue_expr: Column expression indicating tissue type. Default is 'tissue'.
     :param summary_agg_func: Optional aggregation function to use to summarize the
-        transcript expression quantification by tissue. Example: `hl.agg.mean`. Default
+        transcript expression quantification by tissue. Example: `hl.mean`. Default
         is None, which will use a median aggregation.
     :return: A Table of summarized transcript expression by tissue
     """
@@ -67,6 +67,31 @@ def summarize_transcript_expression(
     )
 
     return ht
+
+
+def get_expression_proportion(ht: hl.Table) -> hl.expr.StructExpression:
+    """
+    Calculate the proportion of expression of transcript to gene per tissue.
+
+    :param ht: Table of summarized transcript expression by tissue.
+    :return: Table with expression proportion of transcript to gene per tissue
+        and mean expression proportion across tissues.
+    """
+    tissues = list(ht.row_value)
+
+    # Calculate the sum of transcript expression by gene per tissue.
+    gene_ht = ht.group_by("gene_id").aggregate(
+        **{tissue: hl.agg.sum(ht[tissue]) for tissue in tissues}
+    )
+
+    # Return the proportion of expression of transcript to gene per tissue.
+    gene = gene_ht[ht.gene_id]
+    return hl.struct(
+        **{
+            tissue: hl.utils.misc.divide_null(ht[tissue], gene[tissue])
+            for tissue in tissues
+        }
+    )
 
 
 def filter_expression_ht_by_tissues(
@@ -113,9 +138,10 @@ def tissue_expression_ht_to_array(
     """
     Convert a Table with a row annotation for each tissue to a Table with tissues as an array.
 
-    The output is a Table with a field 'tissue_expression' containing an array of
-    summarized expression values by tissue, where the order of tissues in the array is
-    indicated by the "tissues" global annotation.
+    The output is a Table with fields in `annotations_to_extract`,
+    each containing an array of summarized expression values or proportion
+    by tissue, where the order of tissues in the array is indicated by
+    the "tissues" global annotation.
 
     :param ht: Table with a row annotation for each tissue.
     :param tissues_to_keep: Optional list of tissues to keep in the 'tissue_expression'
@@ -145,28 +171,3 @@ def tissue_expression_ht_to_array(
         )
 
     return ht
-
-
-def get_expression_proportion(ht: hl.Table) -> hl.expr.StructExpression:
-    """
-    Calculate the proportion of expression of transcript to gene per tissue.
-
-    :param ht: Table of summarized transcript expression by tissue.
-    :return: Table with expression proportion of transcript to gene per tissue
-        and mean expression proportion across tissues.
-    """
-    tissues = list(ht.row_value)
-
-    # Calculate the sum of transcript expression by gene per tissue.
-    gene_ht = ht.group_by("gene_id").aggregate(
-        **{tissue: hl.agg.sum(ht[tissue]) for tissue in tissues}
-    )
-
-    # Return the proportion of expression of transcript to gene per tissue.
-    gene = gene_ht[ht.gene_id]
-    return hl.struct(
-        **{
-            tissue: hl.utils.misc.divide_null(ht[tissue], gene[tissue])
-            for tissue in tissues
-        }
-    )
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -1,6 +1,6 @@
 """Utils module containing generic functions that are useful for adding transcript expression-aware annotations."""
 import logging
-from typing import Callable, List, Optional, Tuple, Union
+from typing import Callable, List, Optional, Union
 
 import hail as hl
 
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -14,10 +14,12 @@ logger.setLevel(logging.INFO)
 
 def summarize_transcript_expression(
     mt: hl.MatrixTable,
-    transcript_expression_expr: Union[hl.expr.NumericExpression, str] = "x",
+    transcript_expression_expr: Union[
+        hl.expr.NumericExpression, str
+    ] = "transcript_tpm",
     tissue_expr: Union[hl.expr.StringExpression, str] = "tissue",
     summary_agg_func: Optional[Callable] = None,
-) -> Tuple[hl.Table, hl.Table]:
+) -> hl.Table:
     """
     Summarize a transcript expression MatrixTable by transcript, gene, and tissue.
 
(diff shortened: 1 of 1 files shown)


gnomad/utils/transcript_annotation.py
```diff
@@ -71,3 +71,77 @@ def summarize_rsem_mt(
...
+        ).annotate_globals(tissues=tissues1)
+
+        tissues2 = list(transcript_ht.row)
+        tissues2.remove("transcript_id")
+        tissues2.remove("gene_id")
+        transcript_ht = transcript_ht.select(
+            transcript_expression=hl.array(
+                [transcript_ht[tissue] for tissue in tissues2]
+            )
+        ).annotate_globals(tissues=tissues2)
```
jkgoodrich, line 107: Now this can just use the function I put into my PR

```suggestion
    transcript_ht = tissue_expression_ht_to_array(
        transcript_ht, tissues_to_filter=tissues_to_filter
    )
    gene_ht = tissue_expression_ht_to_array(
        gene_ht, tissues=hl.eval(transcript_ht.tissues)
    )
```

gnomad/utils/transcript_annotation.py
```diff
@@ -71,3 +71,77 @@ def summarize_rsem_mt(
         )
 
     return transcript_ht.key_by("transcript_id", "gene_id"), gene_ht.key_by("gene_id")
+
+
+def get_expression_proportion(
+    transcript_ht: hl.Table,
+    gene_ht: hl.Table,
+    tissues_to_filter: Optional[List[str]] = None,
+    tissue_as_row: bool = False,
```
jkgoodrich, line 80:
```suggestion
```

gnomad/utils/transcript_annotation.py
```diff
@@ -71,3 +71,77 @@ def summarize_rsem_mt(
...
+) -> hl.Table:
+    """
+    Calculate the proportion of expression of transcript to gene per tissue.
+
+    :param transcript_ht: Table of summarized transcript expression by tissue
+    :param gene_ht: Table of summarized gene expression by tissue
+    :param tissues_to_filter: Optional list of tissues to filter out
+    :param tissue_as_row: If True, the input Table is with a row annotation for each
+        tissue instead of an array of values. Default is False.
```
jkgoodrich, line 89:
```suggestion
```

gnomad/utils/transcript_annotation.py
```diff
@@ -71,3 +71,77 @@ def summarize_rsem_mt(
...
+    :return: Table with expression proportion of transcript to gene per tissue
+        and mean expression proportion across tissues
```
jkgoodrich, line 91:
```suggestion
        and mean expression proportion across tissues.
```

gnomad/utils/transcript_annotation.py
```diff
@@ -71,3 +71,77 @@ def summarize_rsem_mt(
...
+        tissues2 = list(transcript_ht.row)
+        tissues2.remove("transcript_id")
+        tissues2.remove("gene_id")
+        transcript_ht = transcript_ht.select(
+            transcript_expression=hl.array(
+                [transcript_ht[tissue] for tissue in tissues2]
+            )
+        ).annotate_globals(tissues=tissues2)
+
+    # Join the transcript expression table and gene expression table
```
jkgoodrich, line 109:
```suggestion
    # Join the transcript expression table and gene expression table.
```

gnomad/utils/transcript_annotation.py
```diff
@@ -71,3 +71,77 @@ def summarize_rsem_mt(
...
+    transcript_ht = transcript_ht.annotate(
+        gene_expression=gene_ht[transcript_ht.gene_id].gene_expression
```
jkgoodrich, line 111: Or whatever we end up calling it in the other function
```suggestion
        gene_expression=gene_ht[transcript_ht.gene_id].tissue_expression
```

gnomad/utils/transcript_annotation.py
```diff
@@ -71,3 +71,77 @@ def summarize_rsem_mt(
...
+                [transcript_ht.transcript_expression[i] for i in tissues_indices]
+            ),
+            gene_expression=hl.array(
+                [transcript_ht.gene_expression[i] for i in tissues_indices]
+            ),
+        )
+
+        transcript_ht = transcript_ht.select_globals(
+            tissues=hl.array([transcript_ht.tissues[i] for i in tissues_indices])
+        )
```
jkgoodrich, line 131: I think this is now all handled in `tissue_expression_ht_to_array`
```suggestion
```

jkgoodrich (changes requested):
Some changes based on my PR

gnomad/utils/transcript_annotation.py
```diff
@@ -71,3 +71,77 @@ def summarize_rsem_mt(
...
+
+    # Calculate the proportion of expression of transcript to gene per tissue
+    transcript_ht = transcript_ht.annotate(
+        exp_prop=hl.or_else(
+            transcript_ht.transcript_expression / transcript_ht.gene_expression,
```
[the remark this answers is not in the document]
jkgoodrich: Ah, I see now, I had in my mind that `summarize_transcript_expression` was running `summary_agg_func` on the grouped `gene_id` rather than `hl.agg.sum`.  Now that I look at this I think we probably want to change the other function because that is confusing. 


There are two ways we can do this. 

The first is to change the other function so it returns a single table with structs for `transcript_expression` and `expression_proportion`. Then modify the other functions accordingly. Something like this:

```
def summarize_transcript_expression(
    mt: hl.MatrixTable,
    transcript_expression_expr: Union[hl.expr.NumericExpression, str] = "transcript_tpm",
    tissue_expr: Union[hl.expr.StringExpression, str] = "tissue",
    summary_agg_func: Optional[Callable] = None,
) -> hl.Table:
    if summary_agg_func is None:
        summary_agg_func = lambda x: hl.median(hl.agg.collect(x))

    if isinstance(transcript_expression_expr, str):
        transcript_expression_expr = mt[transcript_expression_expr]

    if isinstance(tissue_expr, str):
        tissue_expr = mt[tissue_expr]

    mt = mt.group_cols_by(tissue=tissue_expr).aggregate(
        tx=summary_agg_func(transcript_expression_expr)
    )

    transcript_ht = mt.rename({"tx": ""}).make_table()
    transcript_ht = transcript_ht.key_by("transcript_id", "gene_id")

    tissues = list(transcript_ht.row_value)
    
    # Calculate the sum of transcript expression by gene per tissue.
    gene_ht = transcript_ht.group_by("gene_id").aggregate(
        **{
            tissue: hl.agg.sum(transcript_ht[tissue])
            for tissue in tissues
        }
    )

    # Calculate the proportion of expression of transcript to gene per tissue.
    gene_prop = gene_ht[transcript_ht.gene_id]
    transcript_ht = transcript_ht.select(
        transcript_expression=transcript_ht.row_value,
        expression_proportion=hl.struct(
            **{
                tissue: transcript_ht[tissue]/gene_prop[tissue]
                for tissue in tissues
            }
        )
    )
        
    return transcript_ht


def tissue_expression_struct_to_array(
    tissue_struct_expr: hl.expr.StringExpression,
    tissues: Optional[List[str]] = None,
    tissues_to_filter: Optional[List[str]] = None,
) -> Tuple[hl.expr.ArrayExpression, str]:
    if tissues is None:
        tissues = list(tissue_struct_expr.keys())

    if tissues_to_filter is not None:
        logger.info("Filtering tissues: %s", tissues_to_filter)
        tissues = [t for t in tissues if t not in tissues_to_filter]

    tissue_expression = hl.array([tissue_struct_expr[t] for t in tissues])

    return tissue_expression, tissues


def get_expression_proportion(
    tissue_struct_expr: hl.expr.StringExpression,
    tissues_to_filter: Optional[List[str]] = None,
) -> hl.expr.Float64Expression:
    tissue_expr, _ = tissue_expression_struct_to_array(
        tissue_struct_expr=tissue_struct_expr, 
        tissues_to_filter=tissues_to_filter
    )
    
    # Calculate the mean expression proportion across tissues.
    return hl.mean(hl.filter(lambda e: ~hl.is_nan(e), tissue_expr))

```

The other option is to change the summarize function back to only returning the transcript HT and then get the transcript proportion within this function instead of passing in 2 HTs. Something like this:

```
def summarize_transcript_expression(
    mt: hl.MatrixTable,
    transcript_expression_expr: Union[hl.expr.NumericExpression, str] = "transcript_tpm",
    tissue_expr: Union[hl.expr.StringExpression, str] = "tissue",
    summary_agg_func: Optional[Callable] = None,
) -> hl.Table:
    if summary_agg_func is None:
        summary_agg_func = lambda x: hl.median(hl.agg.collect(x))

    if isinstance(transcript_expression_expr, str):
        transcript_expression_expr = mt[transcript_expression_expr]

    if isinstance(tissue_expr, str):
        tissue_expr = mt[tissue_expr]

    mt = mt.group_cols_by(tissue=tissue_expr).aggregate(
        tx=summary_agg_func(transcript_expression_expr)
    )

    transcript_ht = mt.rename({"tx": ""}).make_table()
    transcript_ht = transcript_ht.key_by("transcript_id", "gene_id")
        
    return transcript_ht


def get_expression_proportion(
    transcript_ht: hl.Table,
    tissues_to_filter: Optional[List[str]] = None,
) -> hl.Table:
    transcript_ht = tissue_expression_ht_to_array(
        transcript_ht, tissues_to_filter=tissues_to_filter
    )
    
    tissues = hl.eval(transcript_ht.tissues)
    
    # Calculate the sum of transcript expression by gene per tissue.
    gene_ht = transcript_ht.group_by("gene_id").aggregate(
        expression_sum=hl.agg.array_sum(transcript_ht.tissue_expression)
    )

    # Calculate the mean expression proportion across tissues.
    transcript_ht = transcript_ht.annotate(
        exp_prop_mean=hl.mean(
            hl.filter(
                lambda e: ~hl.is_nan(e), 
                hl.map(
                    lambda x, y: x/y,
                    transcript_ht.tissue_expression,
                    gene_ht[transcript_ht.gene_id].expression_sum
                )
            )
        )
    )
    
    return transcript_ht

```

It sort of depends on whether it's worth storing the proportion as well (or if the HT with that info needs to be released), or if it can just be computed each time it is needed

jkgoodrich (commented):
OK, I finally got around to looking at this again and figured out why I was confused

gnomad/utils/transcript_annotation.py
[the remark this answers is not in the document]
KoalaQin: Im my original code, I put transcript_expression, gene_expression, and expression_proportion in one HT, it was clearer to me that way. And pext scores on v2 exomes are actually showing the expression_proportion, unless we will change that.

gnomad/utils/transcript_annotation.py
```diff
@@ -36,10 +36,9 @@ def summarize_transcript_expression(
         quantification. Default is 'transcript_tpm'.
     :param tissue_expr: Column expression indicating tissue type. Default is 'tissue'.
     :param summary_agg_func: Optional aggregation function to use to summarize the
-        transcript expression quantification by tissue. Example: `hl.agg.mean`. Default
+        transcript expression quantification by tissue. Example: `hl.mean`. Default
```
jkgoodrich, line 39: It needs to use an aggregation function. There is no hl.agg.median, so that is why the `hl.median(hl.agg.collect(x))` is used below. We could also switch to use `hl.agg.approx_median` as the default, but it's a small dataset which is why the collect here wasn't that big of an issue.

```suggestion
        transcript expression quantification by tissue. Example: `hl.agg.mean`. Default
```

gnomad/utils/transcript_annotation.py
```diff
@@ -53,48 +52,122 @@ def summarize_transcript_expression(
     mt = mt.group_cols_by(tissue=tissue_expr).aggregate(
         tx=summary_agg_func(transcript_expression_expr)
     )
+    ht = mt.rename({"tx": ""}).make_table().key_by("transcript_id", "gene_id")
+
+    # Annotate with the proportion of expression of transcript to gene per tissue.
```
jkgoodrich, line 57: Since this now includes the proportion, and the structure of the output table is different, I think this statement in the docstring needs to change, and maybe have an example of the output schema

```
The returned Table has a row annotation for each tissue containing the summarized
tissue expression value.
```

gnomad/utils/transcript_annotation.py
```diff
@@ -53,48 +52,122 @@ def summarize_transcript_expression(
...
+
+def get_expression_proportion(ht: hl.Table) -> hl.expr.StructExpression:
+    """
+    Calculate the proportion of expression of transcript to gene per tissue.
 
-    transcript_ht = mt.rename({"tx": ""}).make_table()
-    transcript_ht = transcript_ht.key_by("transcript_id", "gene_id")
+    :param ht: Table of summarized transcript expression by tissue.
+    :return: Table with expression proportion of transcript to gene per tissue
+        and mean expression proportion across tissues.
```
jkgoodrich, line 78: This needs to change since it no longer returns a table

gnomad/utils/transcript_annotation.py
```diff
@@ -53,48 +52,122 @@ def summarize_transcript_expression(
...
     """
     Convert a Table with a row annotation for each tissue to a Table with tissues as an array.
 
-    The output is a Table with a field 'tissue_expression' containing an array of
-    summarized expression values by tissue, where the order of tissues in the array is
-    indicated by the "tissues" global annotation.
+    The output is a Table with fields in `annotations_to_extract`,
+    each containing an array of summarized expression values or proportion
+    by tissue, where the order of tissues in the array is indicated by
+    the "tissues" global annotation.
```
jkgoodrich, line 144: Not exactly, it either has an annotation of `tissue_expression` that is an array of information per tissue, where each element of the array is the Table's row value for a given tissue. 

Or it can be a Table with one array annotation for each field defined in `annotations_to_extract`, where each array is an array of the given field value per Tissue. 

Maybe an example of the two options would make it clear?

gnomad/utils/transcript_annotation.py
```diff
@@ -53,48 +52,122 @@ def summarize_transcript_expression(
...
 
     :param ht: Table with a row annotation for each tissue.
-    :param tissues: Optional list of tissues to keep in the 'tissue_expression' array.
-        Default is all non-key rows in the Table.
+    :param tissues_to_keep: Optional list of tissues to keep in the 'tissue_expression'
```
jkgoodrich, line 147: Te wording of this should change a little to since it isn't always 'tissue_expression'

jkgoodrich (commented):
Just a few small requests to docstring changes

gnomad/utils/transcript_annotation.py
```diff
@@ -53,48 +52,135 @@ def summarize_transcript_expression(
...
+    # tissue_expression': array<struct {
+    #     transcript_expression: float64,
+    #     expression_proportion: float64
+    # }>
+
+    or a Table with one array annotation for each field defined in the
+    'annotations_to_extract' argument, where each array is an array of the given field
+    values by tissue, like this:
+    # 'transcript_expression': array<float64>
+    # 'expression_proportion': array<float64>
```
jkgoodrich, line 155:
```suggestion
    The output is a Table with one of the two formats:
        - An annotation of 'tissue_expression' containing an array of structs by
          tissue, where each element of the array is the Table's row value for a given
          tissue.

            Example::

                tissue_expression': array<struct {
                    transcript_expression: float64,
                    expression_proportion: float64
                }>

        - A Table with one array annotation for each field defined in the
          'annotations_to_extract' argument, where each array is an array of the given
          field values by tissue.

            Example::

                'transcript_expression': array<float64>
                'expression_proportion': array<float64>
```

jkgoodrich (commented):
Very close, just some small doc changes

jkgoodrich (approved):
LGTM!

PR #652, opened 2023-12-12
make vep_struct_to_csq compatible with vep 101

klaricch:
I ran into errors trying to use vep_struct_to_csq on a file that used v101 for vep annotations because the code references some fields not found in vep101. This PR edits the function to avoid errors when using vep v101.

### DIFF
--- a/gnomad/utils/vep.py
+++ b/gnomad/utils/vep.py
@@ -478,8 +478,7 @@ def vep_struct_to_csq(
 
         # Add exception for transcripts
         if feature_type == "Transcript":
-            fields.update(
-                {
+            transcript_dict = {
                     "canonical": hl.if_else(element.canonical == 1, "YES", ""),
                     "ensp": element.protein_id,
                     "gene": element.gene_id,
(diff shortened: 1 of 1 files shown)


KoalaQin:
I will wait until the checks to pass first.

gnomad/utils/vep.py
```diff
@@ -478,32 +478,43 @@ def get_csq_from_struct(
...
+                ),
+                "mirna": hl.delimit(element.mirna, "&") if "mirna" in element else None,
+                "protein_position": hl.str(element.protein_start) + hl.if_else(
+                    element.protein_start == element.protein_end,
+                    "",
+                    "-" + hl.str(element.protein_end),
+                ),
+                "uniprot_isoform": (
+                    hl.delimit(element.uniprot_isoform, "&")
+                    if "uniprot_isoform" in element
```
KoalaQin, line 505: How about those all-missing fields? Like `swissprot, trembl, uniparc`? And the new fields: `mane_select, mane_plus_clinical, mirna`?

KoalaQin: I know `uniprot_isoform` is new.

klaricch: i don't think the other fields are impacted

KoalaQin (approved):
LGTM! 
I tried this on DRD2 of v4 genomes, the older version of the function didn't run into errors, I guess this is only for a larger dataset. 
It worked fine with your version too.

PR #653, opened 2023-12-19
Suggestions to get_expression_proportion PR

jkgoodrich:
OK, so this is one suggestion to change this based on your comment

> Im my original code, I put transcript_expression, gene_expression, and expression_proportion in one HT, it was clearer to me that way. And pext scores on v2 exomes are actually showing the expression_proportion, unless we will change that.

We could also add gene_expression if you think that would be helpful, and it would have the `expression_proportion` you mention is what is actually shown.

### DIFF
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -1,6 +1,6 @@
 """Utils module containing generic functions that are useful for adding transcript expression-aware annotations."""
 import logging
-from typing import Callable, List, Optional, Union
+from typing import Callable, List, Optional, Tuple, Union
 
 import hail as hl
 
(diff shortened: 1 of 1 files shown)


gnomad/utils/transcript_annotation.py
```diff
@@ -52,17 +52,63 @@ def summarize_transcript_expression(
...
+                expression_proportion=ht.expression_proportion[t],
+            )
+            for t in ht.expression_proportion
+        }
+    )
+
+    return ht
+
+
+def filter_expression_ht_by_tissues(
```
KoalaQin, line 72: Are you putting this function just to filter the tissue-per-column Table?

KoalaQin (changes requested):

PR #657, opened 2024-01-09
amr dict value description fix

mike-w-wilson:
We missed updating "amr" from "Latino" to "Admixed American" in v4.0. This addresses https://github.com/broadinstitute/gnomad_production/issues/1305. 

 This also adds "remaining" to the pop_name dict so we do not have to treat it special like we did in v4.0:https://github.com/broadinstitute/gnomad_qc/blob/b87f62a7fba9473b73bee9cea927bf4559fabfb2/gnomad_qc/v4/create_release/validate_and_export_vcf.py#L135


klaricch (approved):
LGTM!

PR #658, opened 2024-01-09
Add mid to FAF and grpmax calcs

mike-w-wilson:
The genetic ancestry group Middle Eastern should be considered in `grpmax` and `faf` calculations for v4. This was not considered in v3 so we're splitting these globals into version specific dictionaries where the keys are gnomad versions and values are the genetic ancestry groups. This follows the behavior of POPS.

### DIFF
--- a/gnomad/resources/grch38/gnomad.py
+++ b/gnomad/resources/grch38/gnomad.py
@@ -221,7 +221,10 @@ TGP_POP_NAMES = {
 """
 
 POPS_STORED_AS_SUBPOPS = TGP_POPS + HGDP_POPS
-POPS_TO_REMOVE_FOR_POPMAX = {"asj", "fin", "oth", "ami", "remaining"}
+POPS_TO_REMOVE_FOR_POPMAX = {
+    "v3": {"asj", "fin", "mid", "oth", "ami", "remaining"},
+    "v4": {"asj", "fin", "oth", "ami", "remaining"},
+}
 """
 Populations that are removed before popmax calculations.
 """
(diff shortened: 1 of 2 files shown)


klaricch (approved):
LGTM!

gnomad/resources/grch38/gnomad.py
```diff
@@ -221,7 +221,10 @@
 """
 
 POPS_STORED_AS_SUBPOPS = TGP_POPS + HGDP_POPS
-POPS_TO_REMOVE_FOR_POPMAX = {"asj", "fin", "oth", "ami", "mid", "remaining"}
+POPS_TO_REMOVE_FOR_POPMAX = {
+    "v3": {"asj", "fin", "mid", "oth", "ami", "remaining"},
+    "v4": {"asj", "fin", "oth", "ami", "remaining"},
```
klaricch, line 226: either way this is fine, just wondering does v3 have "remaining" and does v4 have "oth"? i don't remember what steps the name changes went in

klaricch (approved):

gnomad/resources/grch38/gnomad.py
mike-w-wilson: Yeah this is a bit annoying, v4 genomes uses remaining but v3 genomes used oth but the code needs both for backwards compatibility since the v4 genomes uses the v3 set since we still have fewer than 1000 mid samples in it....this is why we need a resources overhaul.

mike-w-wilson: I kept oth in v4 as a safety measure here because of the joint_frequency work, thats also why ami is in v4 even though v4 exomes do not have ami samples

PR #660, opened 2024-01-10
Clean up `compute_coverage_stats`, change it to use `agg_by_strata` and have an optional `group_membership_ht` parameter

jkgoodrich:
The following notebook was used for testing: http://localhost:8123/notebooks/gnomad-julia/gnomad_v4/test_v4_coverage_function_cleanup.ipynb

### DIFF
--- a/gnomad/utils/sparse_mt.py
+++ b/gnomad/utils/sparse_mt.py
@@ -960,11 +960,13 @@ def compute_coverage_stats(
         raise ValueError(
             "Only one of 'group_membership_ht' or 'strata_expr' can be specified."
         )
-        
-    # Initialize no_strata and default strata_expr if neither group_membership_ht nor strata_expr is provided
+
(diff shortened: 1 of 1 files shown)


gnomad/utils/sparse_mt.py
```diff
@@ -934,38 +936,66 @@ def compute_coverage_stats(
...
+
+    gt_field = gt_field.pop()
+
+    no_strata = False
+    if group_membership_ht is not None:
+        if strata_expr is not None:
+            raise ValueError(
+                "Only one of 'group_membership_ht' or 'strata_expr' can be specified."
+            )
     else:
```
mike-w-wilson, line 965:
```suggestion
```

mike-w-wilson (approved):
Another reorg at the top similar to the last review but smaller and I could do it in here rather than a PR. Other than that, this looks good considering the follow up PRs.

PR #662, opened 2024-01-11
Add `compute_stats_per_ref_site` to generalize computation of aggregate stats at all sites in a reference Table

jkgoodrich:
Also add a helper function `compute_allele_number_per_ref_site` to use `compute_stats_per_ref_site` to compute AN

### DIFF
--- a/gnomad/utils/sparse_mt.py
+++ b/gnomad/utils/sparse_mt.py
@@ -1051,20 +1051,20 @@ def compute_stats_per_ref_site(
     # If the adj annotation is needed then "adj" must be present on mt, or AD/LAD, DP,
     # and GQ must be present.
     en = set(mt.entry)
-    gt_field = (en & {"GT"} or en & {"LGT"}).pop()
-    ad_field = (en & {"AD"} or en & {"LAD"}).pop()
-    adj_fields = en & {"adj"} or {"DP", "GQ", ad_field} if adj else set([])
+    gt_field = en & {"GT"} or en & {"LGT"}
+    ad_field = en & {"AD"} or en & {"LAD"}
+    adj_fields = en & {"adj"} or ({"DP", "GQ"} | ad_field) if adj else set([])
 
     if not gt_field:
-        logger.info("No genotype field found in entry fields!")
+        raise ValueError("No genotype field found in entry fields!")
 
     if adj and not adj_fields.issubset(en):
-        logger.info(
+        raise ValueError(
             "No 'adj' found in entry fields, and one of AD/LAD, DP, and GQ is missing "
             "so adj can't be computed!"
         )
 
-    entry_keep_fields = set(entry_keep_fields or set([])) | {gt_field} | adj_fields
+    entry_keep_fields = set(entry_keep_fields or set([])) | gt_field | adj_fields
 
     # Initialize no_strata and default strata_expr if neither group_membership_ht nor
     # strata_expr is provided.
@@ -1197,10 +1197,11 @@ def compute_coverage_stats(
         mt = mtds
 
     # Determine the genotype field.
-    en = set(mt.entry)
-    gt_field = (en & {"GT"} or en & {"LGT"}).pop()
+    gt_field = set(mt.entry) & {"GT", "LGT"}
     if not gt_field:
-        logger.info("No genotype field found in entry fields!")
+        raise ValueError("No genotype field found in entry fields!")
+
+    gt_field = gt_field.pop()
 
     # Add function to compute coverage stats.
     cov_bins = sorted(coverage_over_x_bins)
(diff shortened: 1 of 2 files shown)


gnomad/utils/sparse_mt.py
```diff
@@ -1105,7 +1124,84 @@ def compute_coverage_stats(
...
     )
 
-    # Compute coverage stats.
+    # Annotate with adj if needed.
+    if adj and "adj" not in mt.entry:
+        mt = annotate_adj(mt)
+
+    ht = agg_by_strata(mt, entry_agg_funcs, group_membership_ht=group_membership_ht)
+    ht = ht.checkpoint(hl.utils.new_temp_file("agg_stats", "ht"))
+
```
mike-w-wilson, line 1133:
```suggestion

    # Drop no longer needed fields
```

mike-w-wilson (commented):
This looks good. Sorry it took awhile, it's dense. I only have one thing with a value error, everything else is small comments or nitpicks with style.

gnomad/utils/sparse_mt.py
```diff
@@ -991,67 +992,91 @@ def densify_all_reference_sites(
...
+    adj = hl.eval(
+        hl.any(g.get("adj_groups", hl.empty_array("bool")))
+        | hl.any(
+            g.get("freq_meta", hl.empty_array("dict<str, str>")).map(
+                lambda x: x.get("group", "NA") == "adj"
+            )
+        )
+    )
+
+    # Determine the entry fields on mt that should be densified.
```
jkgoodrich, line 1049: OK, I fixed this to be what I actually wanted here. Here is some testing:

```
def test_fields(en, adj, entry_keep_fields):
    print(f"\nTesting: en: {en}, adj: {adj}, entry_keep_fields: {entry_keep_fields}")
    gt_field = (en & {"GT"} or en & {"LGT"})
    ad_field = (en & {"AD"} or en & {"LAD"})
    adj_fields = en & {"adj"} or ({"DP", "GQ"} | ad_field) if adj else set([])

    if not gt_field:
        print("\tNo genotype field found in entry fields!")
        return

    if adj and not adj_fields.issubset(en):
        print(
            "\tNo 'adj' found in entry fields, and one of AD/LAD, DP, and GQ is missing "
            "so adj can't be computed!"
        )
        return

    entry_keep_fields = set(entry_keep_fields or set([])) | gt_field | adj_fields
    
    print("\tgt_field", gt_field)
    print("\tad_field", ad_field)
    print("\tadj_fields", adj_fields)
    print("\tentry_keep_fields", entry_keep_fields)
    

test_fields(en=set([]), adj=False, entry_keep_fields=set([]))
test_fields(en={"GT"}, adj=False, entry_keep_fields=set([]))
test_fields(en={"GT", "AD", "adj"}, adj=False, entry_keep_fields=set([]))

test_fields(en=set([]), adj=True, entry_keep_fields=set([]))
test_fields(en={"GT"}, adj=True, entry_keep_fields=set([]))
test_fields(en={"GT", "adj"}, adj=True, entry_keep_fields=set([]))
test_fields(en={"GT", "AD"}, adj=True, entry_keep_fields=set([]))
test_fields(en={"GT", "adj", "AD"}, adj=True, entry_keep_fields=set([]))
test_fields(en={"GT", "adj", "DP", "GQ"}, adj=True, entry_keep_fields=set([]))
test_fields(en={"GT", "AD", "DP", "GQ"}, adj=True, entry_keep_fields=set([]))
test_fields(en={"GT", "adj", "AD", "DP", "GQ"}, adj=True, entry_keep_fields=set([]))
```

I switched GT to this to be consistent with what I needed for AD. And I needed AD to be like this because I needed it to only choose one field in case (though unlikely) there was both AD and LAD, I only need one for the set below, and this way since I don't need to keep the actual value, I don't need the pop
    gt_field = en & {"GT"} or en & {"LGT"}
    ad_field = en & {"AD"} or en & {"LAD"}

gnomad/utils/sparse_mt.py
```diff
@@ -991,67 +992,91 @@ def densify_all_reference_sites(
...
+        VariantDataset/MatrixTable. The keys of the dict are the names of the
+        annotations and the values are tuples of functions. The first function is used
+        to transform the `mt` entries in some way, and the second function is used to
+        aggregate the output from the first function.
+    :param row_key_fields: Fields to use as row key. Defaults to locus.
+    :param interval_ht: Optional table of intervals to filter to.
+    :param entry_keep_fields: Fields to keep in entries before performing the
+        densification in `densify_all_reference_sites`. Should include any fields
+        needed for the functions in `entry_agg_funcs`. By default, only GT or LGT is
+        kept.
```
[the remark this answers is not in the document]
jkgoodrich: OK, I fixed the code below to be what I actually wanted, and I think this is OK as is now

gnomad/utils/sparse_mt.py
mike-w-wilson: Ah Ok, that makes a lot of sense, thanks for that testing too!

gnomad/utils/sparse_mt.py
```diff
@@ -1305,19 +1307,21 @@ def compute_allele_number_per_ref_site(
...
-    # Determine the genotype field.
     if isinstance(mtds, hl.vds.VariantDataset):
         mt = mtds.variant_data
     else:
         mt = mtds
 
+    # Determine the genotype field.
     en = set(mt.entry)
-    gt_field = (en & {"GT"} or en & {"LGT"}).pop()
+    gt_field = set(mt.entry) & {"GT", "LGT"}
```
mike-w-wilson, line 1317: Following similar logic from above with AD, on the off chance both of these are present, we wouldnt actually know which was used right? .pop() could grab either. I think the old logic makes sense since we would know GT was used if they were both present? I'm not sure it actually matters and its a weird edge case but the code couldnt help you figure out which was used.

mike-w-wilson (approved):
Ok one tiny thing on the GT logic now but other than that this looks good!

PR #663, opened 2024-01-11
Function to get max pext per gene

KoalaQin:
This is an optional function to get maximum pext score per gene, by filtering to only coding variants and aggregating by gene, in case the transcript expression proportion of coding transcripts is smaller than that of non-coding transcripts in a gene.

### DIFF
--- a/gnomad/utils/vep.py
+++ b/gnomad/utils/vep.py
@@ -51,6 +51,10 @@ CSQ_CODING_LOW_IMPACT = [
     "coding_sequence_variant",
 ]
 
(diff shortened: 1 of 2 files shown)


gnomad/utils/vep.py
```diff
@@ -51,6 +51,10 @@
     "coding_sequence_variant",
 ]
 
+CSQ_CODING_ALL_IMPACT = (
```
jkgoodrich, line 54: This is the same as `CSQ_CODING` below

PR #668, opened 2024-01-22
Account for missingness in int64 to int32 VCF type conversion

mike-w-wilson:
The adjust_vcf_incompatible_types function does not account for missingness in the int64 -> int32 conversion. Current behavior will make any missing int64 the max int32 value, `2147483647`. This was seen in the v4 exome chrY VCF where the int64 field was converted to int32s but all missing values (homozygote counts  within the freq array for XX groupings on chrY) where made 2147483647. This fixes the function by only converting defined values and leaving them missing otherwise.


klaricch (approved):
lgtm!

PR #671, opened 2024-01-24
Add v4 to genome release constants

klaricch:


klaricch:
Add "v4.0" to GENOME_RELEASES and set CURRENT_GENOME_RELEASE to "v4.0"

Currently public_release("genomes").versions['4.0'] will fail with a key error. I think the only line this would break is https://github.com/broadinstitute/gnomad_qc/blob/901f0d8629eb69197548c3ae94ccae441bd46a4f/gnomad_qc/v3/annotations/compare_freq.py#L29 but since we are no longer updating v3 we might be able to hard code the version in that script instead

mike-w-wilson (approved):
LGTM!  I agree we should just hardcode "3.1.2" here: https://github.com/broadinstitute/gnomad_qc/blob/901f0d8629eb69197548c3ae94ccae441bd46a4f/gnomad_qc/v3/annotations/compare_freq.py#L67C19-L67C37.

PR #672, opened 2024-01-26
Add `read_args` parameter to the read functions of Resource Classes

jkgoodrich:
It would be useful to have the option of a repartition on read for existing resources. This adds a `read_args` parameter to the read functions of relevant Resource classes so any additional desired parameters can be passed to the hail read functions. To repartition a HT on read it would just be: `my_resource.ht(read_args={"_n_partitions": 50})`

### DIFF
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -350,7 +350,7 @@ class ExpressionResource(BaseResource):
 
         :param force_import: If ``True``, force the import of the resource even if it
             already exists.
-        :param read_args: Any additional arguments to pass to hl.experimental.
+        :param read_args: Any additional arguments to pass to hl.experimental.read_expression.
         :return: Hail Expression resource.
         """
         if self.path is None or force_import:
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -109,16 +109,23 @@ class TableResource(BaseResource):
 
     expected_file_extensions: List[str] = [".ht"]
 
-    def ht(self, force_import: bool = False) -> hl.Table:
+    def ht(
+        self,
+        force_import: bool = False,
+        read_args: Optional[Dict[str, Any]] = None,
+    ) -> hl.Table:
         """
         Read and return the Hail Table resource.
 
+        :param force_import: If ``True``, force the import of the resource even if it
+            already exists.
+        :param read_args: Any additional arguments to pass to hl.read_table.
         :return: Hail Table resource
         """
         if self.path is None or force_import:
             return self.import_func(**self.import_args)
         else:
-            return hl.read_table(self.path)
+            return hl.read_table(self.path, **(read_args or {}))
 
     def import_resource(self, overwrite: bool = True, **kwargs) -> None:
         """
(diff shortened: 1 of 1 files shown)


gnomad/resources/resource_utils.py
```diff
@@ -109,16 +109,23 @@ class TableResource(BaseResource):
...
-    def ht(self, force_import: bool = False) -> hl.Table:
+    def ht(
+        self,
+        force_import: bool = False,
+        read_args: Optional[Dict[str, Any]] = None,
+    ) -> hl.Table:
         """
         Read and return the Hail Table resource.
 
+        :param force_import: If ``True``, force the import of the resource even if it
```
ch-kr, line 120: as a basic question, why is this script the only place in the repo where True is wrapped in two backticks in docstrings? (is this the format we should be using for our other docstrings?)

gnomad/resources/resource_utils.py
```diff
@@ -109,16 +109,23 @@ class TableResource(BaseResource):
...
+            already exists.
+        :param read_args: Any additional arguments to pass to hl.read_table.
         :return: Hail Table resource
         """
         if self.path is None or force_import:
             return self.import_func(**self.import_args)
         else:
-            return hl.read_table(self.path)
+            return hl.read_table(self.path, **(read_args or {}))
```
ch-kr, line 128: another basic question: if `read_args` defaults to `{}` instead of None, then you wouldn't need the `(read_args or {})`. did you set this to None just to keep the argument Optional?

gnomad/resources/resource_utils.py
```diff
@@ -318,16 +340,23 @@ class ExpressionResource(BaseResource):
...
+        self,
+        force_import: bool = False,
+        read_args: Optional[Dict[str, Any]] = None,
+    ) -> hl.expr.Expression:
         """
         Read and return the Hail Expression resource.
 
+        :param force_import: If ``True``, force the import of the resource even if it
+            already exists.
+        :param read_args: Any additional arguments to pass to hl.experimental.
```
ch-kr, line 353:
```suggestion
        :param read_args: Any additional arguments to pass to hl.experimental.read_expression.
```

ch-kr: I like having adding `read_args` for consistency here, but I'm not sure we actually need it for HEs (the only other arg accepted right now is `_assert_type`) -- what do you think?

ch-kr (commented):
YAY thank you so much for adding this! I have a few basic questions and one tiny suggestion

gnomad/resources/resource_utils.py
jkgoodrich: Yeah I'm not completely sure either. It was suggested by co-pilot and was already used in other parts of this file. It looks like maybe we should be using two, but I'm not positive.

https://documatt.com/restructuredtext-reference/element/inline-literal.html

```
Common issues:
Don’t interchange ` and ``
There is a big difference between `code example` and ``code example``! Only the latter with double `` is inline code example! Text wrapped in single ` is so-called default role interpreted text.

This is tricky part of reStructuredText because in Markdown, code examples are in wrapped in single backtick.
```

[Here is what the two backticks looks like](https://broadinstitute.github.io/gnomad_methods/api_reference/resources/resource_utils.html#gnomad.resources.resource_utils.BaseResource)

gnomad/resources/resource_utils.py
jkgoodrich: Yeah, I was really just following what was done above for `import_args: Optional[Dict[str, Any]] = None,`

I'm fine with changing to default to {} if you prefer

gnomad/resources/resource_utils.py
jkgoodrich: I think it doesn't hurt to leave, then it's there if the HEs evolve to have more

gnomad/resources/resource_utils.py
ch-kr: whoa, TIL. I like how the double backticks looks in our docs, so definitely happy to adopt this practice moving forward

gnomad/resources/resource_utils.py
ch-kr: ah OK, works for me!

ch-kr (approved):
LGTM!

PR #677, opened 2024-02-22
Add `sex_karyotype_field` as an argument to `compute_stats_per_ref_site` to include sex ploidy adjustment after densify

jkgoodrich:


mike-w-wilson (approved):
I just added an inline doc to the HT write because at first pass I didnt understand why we just dont reference the field in the MT. Looks good to me though!

PR #678, opened 2024-02-23
Fix sex ploidy adjustment so XX samples still get set to missing on chrY

jkgoodrich:


gnomad/sample_qc/sex.py
```diff
@@ -42,14 +42,17 @@ def adjusted_sex_ploidy_expr(
 
     return (
         hl.case(missing_false=True)
+        # Added to reduce the checks by entry.
+        .when(row_idx.in_autosome, gt_expr)
+        .when((row_idx.y_par | row_idx.y_nonpar) & col_idx.xx, hl.missing(hl.tcall))
+        # Added to reduce the checks by entry.
```
mike-w-wilson, line 48:
```suggestion
```

mike-w-wilson (approved):
This makes sense

PR #681, opened 2024-03-04
Add function for adding gencode annotation

klaricch:


ch-kr (commented):
one suggestion and a few nitpicks

gnomad/utils/constraint.py
```diff
@@ -1282,3 +1282,58 @@ def calculate_raw_z_score_sd(
...
+        gencode_ht.filter(gencode_ht.feature == "transcript")
+        .select("chromosome", "transcript_id", "level", "transcript_type")
+        .key_by("transcript_id")
+        .drop("interval")
+    )
+
+    # Obtain CDS annotations from GENCODE file and calculate CDS length and
+    # number of exons.
+    gencode_cds = gencode_ht.filter(gencode_ht.feature == "CDS").select(
+        "transcript_id", "length"
```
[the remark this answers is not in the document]
klaricch: had split into two because still wanted "level" and "transcript_type" for non-CDS features

gnomad/utils/constraint.py
```diff
@@ -1310,30 +1314,32 @@ def add_gencode_annotations(
         chromosome=gencode_ht.interval.start.contig,
     )
 
-    # Obtain transcript annotations from GENCODE file.
-    gencode_transcript = (
-        gencode_ht.filter(gencode_ht.feature == "transcript")
-        .select("chromosome", "transcript_id", "level", "transcript_type")
+    # Obtain CDS annotations from GENCODE file and calculate CDS length and
+    # number of exons.
+    annotations_to_add = annotations + ("chromosome", "transcript_id", "length")
```
ch-kr, line 1319:
```suggestion
    annotations_to_add = set(annotations + ["chromosome", "transcript_id", "length"])
```
minor update to make the concatenation work + drop potentially duplicated annotations

gnomad/utils/constraint.py
```diff
@@ -1310,30 +1314,32 @@ def add_gencode_annotations(
...
         .key_by("transcript_id")
         .drop("interval")
     )
 
-    # Obtain CDS annotations from GENCODE file and calculate CDS length and
-    # number of exons.
-    gencode_cds = gencode_ht.filter(gencode_ht.feature == "CDS").select(
-        "transcript_id", "length"
-    )
+    annotations_to_add.remove("length")
```
ch-kr, line 1328: doesn't this mean that length gets dropped?

gnomad/utils/constraint.py
```diff
@@ -1284,24 +1284,28 @@ def calculate_raw_z_score_sd(
...
     gencode_ht: hl.Table,
+    annotations: List[str] = ["level", "transcript_type"],
 ) -> hl.Table:
     """
     Add GENCODE annotations to Table based on transcript id.
 
     .. note::
-        Added annotations are:
-        - chromosome
+        Added annotations by default are:
```
ch-kr, line 1296: maybe should add a note here that the added annotations become keys for the group by (so `cds_length` will get calculated over the combination of all added annotations, for example)

ch-kr (commented):
minor suggestion + question about `length`

gnomad/utils/constraint.py
klaricch: we only need it to calculate the total cds length but don't need it for grouping

gnomad/utils/constraint.py
```diff
@@ -1282,3 +1282,65 @@ def calculate_raw_z_score_sd(
...
+    gencode_cds = (
+        gencode_cds.group_by(*annotations_to_add)
+        .aggregate(
+            cds_length=hl.agg.sum(gencode_cds.length), num_coding_exons=hl.agg.count()
+        )
+        .key_by("transcript_id")
+    )
+
+    gencode_cds = gencode_cds.checkpoint(
+        new_temp_file(prefix="gencode_cds", extension="ht")
```
ch-kr, line 1340: one minor comment that I missed before -- should you specify overwrite=True here? (or add an overwrite argument to the function to control _read_if_exists vs overwrite)

ch-kr (approved):
one minor note about `overwrite` but LGTM!

gnomad/utils/constraint.py
klaricch: it doesn't matter what overwrite is set to if new_temp_file is used

ch-kr: ooh, really? that's really cool! but then I guess you lose the functionality of reading from existing files for slower pipelines

PR #685, opened 2024-03-15
Add option to exclude polyphen from process consequences

KoalaQin:
This change makes an option to not use polyphen_prediction in csq_order and its prioritization score. Test on vep_context 105.

### DIFF
--- a/gnomad/utils/vep.py
+++ b/gnomad/utils/vep.py
@@ -304,10 +304,15 @@ def process_consequences(
     :param csq_order: Optional list indicating the order of VEP consequences, sorted
         from high to low impact. Default is None, which uses the value of the
         `CSQ_ORDER` global.
-    :param has_polyphen: Whether the input VEP Struct has PolyPhen and SIFT.
-        Default is True. Note: VEP105-annotated context HT still has PolyPhen inside
-        vep struct but removed from the VEP105-annotated release HTs, so this has to be
-        set to False if using gnomAD v4.0 and later release HTs.
+:param has_polyphen: Whether the input VEP Struct has a PolyPhen annotation which
+        will be used to modify the consequence score. Default is True.
+
+              .. note::
+
+            From gnomAD v4.0 on, the PolyPhen annotation was removed from the VEP
+            Struct in the release HTs. When using this function with gnomAD v4.0 or
+            later, set `has_polyphen` to False.
+
     :return: MT with better formatted consequences.
     """
     if csq_order is None:
--- a/gnomad/utils/vep.py
+++ b/gnomad/utils/vep.py
@@ -290,7 +290,7 @@ def process_consequences(
     vep_root: str = "vep",
     penalize_flags: bool = True,
     csq_order: Optional[List[str]] = None,
-    has_polyphen_sift: bool = False,
+    has_polyphen: bool = True,
 ) -> Union[hl.MatrixTable, hl.Table]:
     """
     Add most_severe_consequence into [vep_root].transcript_consequences, and worst_csq_by_gene, any_lof into [vep_root].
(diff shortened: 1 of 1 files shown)


mike-w-wilson (changes requested):
One suggestion on the default and a question!

gnomad/utils/vep.py
```diff
@@ -290,6 +290,7 @@ def process_consequences(
     vep_root: str = "vep",
     penalize_flags: bool = True,
     csq_order: Optional[List[str]] = None,
+    has_polyphen_sift: bool = False,
```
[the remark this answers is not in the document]
KoalaQin: I will add a note in VEP 105, all the release HT won't have polyphen, we have to set this to `False`.

gnomad/utils/vep.py
```diff
@@ -303,44 +304,42 @@ def process_consequences(
...
+
+    flag_score = 500
+    no_flag_score = flag_score * (1 + penalize_flags)
+
+    def csq_score_modifier(tc: hl.expr.StructExpression) -> float:
+        modifier = csq_score(tc)
+        flag_condition = (tc.lof == "HC") & (tc.lof_flags != "")
+        modifier -= hl.if_else(flag_condition, flag_score, no_flag_score)
+        modifier -= hl.if_else(tc.lof == "OS", 20, 0)
+        modifier -= hl.if_else(tc.lof == "LC", 10, 0)
```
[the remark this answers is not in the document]
KoalaQin: I wanted to keep the previous structure, but if I add if in, it will be a bit redundant, so I worked with chatGPT for a few rounds to simplify it.

gnomad/utils/vep.py
```diff
@@ -303,44 +304,44 @@ def process_consequences(
...
+        vep struct but removed from the VEP105-annotated release HTs, so this has to be
+        set to False if using gnomAD v4.0 and later release HTs.
     :return: MT with better formatted consequences.
     """
     if csq_order is None:
         csq_order = CSQ_ORDER
     csqs = hl.literal(csq_order)
     csq_dict = hl.literal(dict(zip(csq_order, range(len(csq_order)))))
 
+    def csq_score(tc: hl.expr.StructExpression) -> int:
```
mike-w-wilson, line 318: I think we should make these "private" with a _ prefix since they only make sense inside this function. We haven't always done this through the repo, this function included but I think its a helpful thing to add.

gnomad/utils/vep.py
```diff
@@ -303,44 +304,44 @@ def process_consequences(
...
+        return csq_dict[tc.most_severe_consequence]
+
+    flag_score = 500
+    no_flag_score = flag_score * (1 + penalize_flags)
+
+    def csq_score_modifier(tc: hl.expr.StructExpression) -> float:
```
mike-w-wilson, line 324: Same comment as above -> `_csq_score_modifier`

mike-w-wilson (commented):
Two little things

gnomad/utils/vep.py
```diff
@@ -303,44 +304,44 @@ def process_consequences(
     :param csq_order: Optional list indicating the order of VEP consequences, sorted
         from high to low impact. Default is None, which uses the value of the
         `CSQ_ORDER` global.
+    :param has_polyphen: Whether the input VEP Struct has PolyPhen and SIFT.
+        Default is True. Note: VEP105-annotated context HT still has PolyPhen inside
+        vep struct but removed from the VEP105-annotated release HTs, so this has to be
+        set to False if using gnomAD v4.0 and later release HTs.
```
mike-w-wilson, line 310:
```suggestion
:param has_polyphen: Whether the input VEP Struct has a PolyPhen annotation which
        will be used to modify the consequence score. Default is True.

              .. note::

            From gnomAD v4.0 on, the PolyPhen annotation was removed from the VEP
            Struct in the release HTs. When using this function with gnomAD v4.0 or
            later, set `has_polyphen` to False.

```

mike-w-wilson (commented):
I rewrote a comment but then LGTM

mike-w-wilson (approved):

KoalaQin:
I can't pass check for the "note", so I moved them before. Is it Okay?

PR #687, opened 2024-03-27
Update ancestry population names in gks method to always use v4

theferrit32:
Annotating v4 genomes now fails since the v4 genomes table uses ancestry population names from v4. This fixes that.

### DIFF
--- a/gnomad/resources/grch38/gnomad.py
+++ b/gnomad/resources/grch38/gnomad.py
@@ -572,11 +572,7 @@ def gnomad_gks(
         ht = ht.annotate(fraction_cov_over_20=coverage_ht[ht.locus].over_20)
 
     # Retrieve ancestry groups from the imported POPS dictionary.
-    pops_list = (
-        list(POPS["v4" if data_type == "exomes" else "v3"])
-        if by_ancestry_group
-        else None
-    )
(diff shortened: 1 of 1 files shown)


klaricch:
can you clarify the issue? does it involve the change from oth -> remaining?
just switching to always use POPS v4 isn't suitable as currently written because it would drop 'ami' from genomes

klaricch (changes requested):
https://github.com/broadinstitute/gnomad_methods/pull/687#issuecomment-2025483805

theferrit32:
Yes @klaricch this was because some of the v3 genomes population labels are no longer in the v4 genomes so it was getting a KeyError. I guess this change is too imprecise though because it might miss some that are there. Do you have an idea of how to better keep the population labels we reference in sync with what's in the tables?

Should another set of labels be added? If v3 genomes and v4 genomes and v4 exomes don't all have the same ancestry groups then do we need 3 sets of keys instead of just the two v3 and v4 in hardcoded in there now? Or can we infer the groups from whatever's in the table?

klaricch:
I think the issue you were seeing should be resolved with this PR (https://github.com/broadinstitute/gnomad_methods/pull/690), but if not, let me know

theferrit32:
Thank you @klaricch !

PR #688, opened 2024-03-29
Add row_name option in case histograms is nested under a row

KoalaQin:
In exomes or genomes release HT, `histograms` is a row, but in the joint releast HT, it's under the row of `exomes`, `genomes`, or `joint`. This change make it possible to access the histograms in both cases.

### DIFF
--- a/gnomad/utils/vcf.py
+++ b/gnomad/utils/vcf.py
@@ -1158,6 +1158,7 @@ def make_vcf_filter_dict(
 def make_hist_bin_edges_expr(
     ht: hl.Table,
     hists: List[str] = HISTS,
+    row_name: str = "",
     prefix: str = "",
     label_delimiter: str = "_",
     include_age_hists: bool = True,
@@ -1167,24 +1168,32 @@ def make_hist_bin_edges_expr(
 
     :param ht: Table containing histogram variant annotations.
     :param hists: List of variant histogram annotations. Default is HISTS.
+    :param row_name: Name of row annotation containing histogram data. In exomes or
+        genomes release HT, `histograms` is a row, but in the joint releast HT, it's
+        under the row of `exomes`, `genomes`, or `joint`.
     :param prefix: Prefix text for age histogram bin edges.  Default is empty string.
     :param label_delimiter: String used as delimiter between prefix and histogram annotation.
     :param include_age_hists: Include age histogram annotations.
-    :return: Dictionary keyed by histogram annotation name, with corresponding reformatted bin edges for values.
+    :return: Dictionary keyed by histogram annotation name, with corresponding
+        reformatted bin edges for values.
     """
     # Add underscore to prefix if it isn't empty
     if prefix != "":
         prefix += label_delimiter
 
     edges_dict = {}
+    if row_name and hasattr(ht[row_name], "histograms"):
+        ht_row = ht.head(1)[row_name]
+    else:
+        ht_row = ht.head(1)
+
     if include_age_hists:
         edges_dict.update(
             {
                 f"{prefix}{call_type}": "|".join(
                     map(
                         lambda x: f"{x:.1f}",
-                        ht.head(1)
-                        .histograms.age_hists[f"age_hist_{call_type}"]
+                        ht_row.histograms.age_hists[f"age_hist_{call_type}"]
                         .collect()[0]
                         .bin_edges,
                     )
@@ -1203,7 +1212,7 @@ def make_hist_bin_edges_expr(
             edges_dict[hist_name] = "|".join(
                 map(
                     lambda x: f"{x:.2f}" if "ab" in hist else str(int(x)),
-                    ht.head(1).histograms[hist_type][hist].collect()[0].bin_edges,
+                    ht_row.histograms[hist_type][hist].collect()[0].bin_edges,
                 )
             )
 


gnomad/utils/vcf.py
```diff
@@ -1158,6 +1158,7 @@ def make_vcf_filter_dict(
 def make_hist_bin_edges_expr(
     ht: hl.Table,
     hists: List[str] = HISTS,
+    row_name: str = "",
```
mike-w-wilson, line 1161: Can we be more descriptive than`row_name`? Maybe `ann_w_hists` or something? Could you also make this Optional? And  whats the benefit of defaulting to an empty string instead of None?

gnomad/utils/vcf.py
```diff
@@ -1167,24 +1168,32 @@ def make_hist_bin_edges_expr(
...
-    :return: Dictionary keyed by histogram annotation name, with corresponding reformatted bin edges for values.
+    :return: Dictionary keyed by histogram annotation name, with corresponding
+        reformatted bin edges for values.
     """
     # Add underscore to prefix if it isn't empty
     if prefix != "":
         prefix += label_delimiter
 
     edges_dict = {}
+    if row_name and hasattr(ht[row_name], "histograms"):
```
mike-w-wilson, line 1185: Maybe this answers my question above, does ht[None] fail? Though my gut is if None was passed this would short circuit and `hasattr(ht[None],...)` wouldnt be evaluated

mike-w-wilson (changes requested):
Some questions and suggestions

gnomad/utils/vcf.py
KoalaQin: I tried to use ht[None], it won't work. You meant something like this? 
```commandline
    row_name: Optional[str] = None,
...
                        ht.head(1)[row_name].histograms.age_hists[f"age_hist_{call_type}"]

```

mike-w-wilson: Yes, can you explain how it didnt work? I'd expect the `if row_name` to evaluate to False and `hasattr(..` wouldnt even be evaluated so your second line would never be an issue with None. But perhaps hail is getting angry about it when its creating the DAG.

KoalaQin: This error: 
`TypeError: Table.__getitem__: invalid index argument(s)
  Usage 1: field selection: ht['field']
  Usage 2: Left distinct join: ht[ht2.key] or ht[ht2.field1, ht2.field2]`

gnomad/utils/vcf.py
```diff
@@ -1167,24 +1243,32 @@ def make_hist_bin_edges_expr(
...
+
     if include_age_hists:
         edges_dict.update(
             {
                 f"{prefix}{call_type}": "|".join(
                     map(
                         lambda x: f"{x:.1f}",
-                        ht.head(1)
-                        .histograms.age_hists[f"age_hist_{call_type}"]
+                        ht_row.histograms.age_hists[f"age_hist_{call_type}"]
```
mike-w-wilson, line 1271: This will fail if the datatype has missing values in the first row since you cannot index into a NoneType. We need to add logic to handle the missing values.

gnomad/utils/vcf.py
```diff
@@ -1203,7 +1287,7 @@ def make_hist_bin_edges_expr(
             edges_dict[hist_name] = "|".join(
                 map(
                     lambda x: f"{x:.2f}" if "ab" in hist else str(int(x)),
-                    ht.head(1).histograms[hist_type][hist].collect()[0].bin_edges,
+                    ht_row.histograms[hist_type][hist].collect()[0].bin_edges,
```
mike-w-wilson, line 1290: Same comment as above.

gnomad/utils/vcf.py
```diff
@@ -1158,6 +1233,7 @@ def make_vcf_filter_dict(
 def make_hist_bin_edges_expr(
     ht: hl.Table,
     hists: List[str] = HISTS,
+    ann_with_hists: str = "",
```
mike-w-wilson, line 1236: I know we pass other empty strings but I think given how this one works in the conditional, it makes more sense to make it 
```suggestion
    ann_with_hists: Optional[str] = None,
```
This works fine calling the function on its own for me.

mike-w-wilson: I can't suggest it here because you havent changed them at all but the Optional will need to be added to the typing import at the beginning of the module.

gnomad/utils/vcf.py
```diff
@@ -993,7 +1020,16 @@ def make_info_dict(
...
-            metrics = ["AC", "AN", "AF", "nhomalt", "faf95", "faf99"]
+            metrics = [
+                "AC",
+                "AN",
+                "AF",
+                "nhomalt",
+                "faf95",
+                "faf99",
+                "CTT_odds_ratio",
+                "CTT_p_value",
```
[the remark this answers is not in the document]
KoalaQin: Yes, thanks for catching this!

gnomad/utils/vcf.py
KoalaQin: head(1) worked very fast, my current change is very slow to run, do you have a faster suggestion?

KoalaQin: Unless I add more secondary workers...

KoalaQin:
I restructured `make_hist_bin_edges_expr` to implement our thinking, I wanted to simplify it more but it didn't work, let me know if this version can still be improved.

gnomad/utils/vcf.py
```diff
@@ -327,6 +327,40 @@
 Dictionary used during VCF export to export row (variant) annotations.
 """
 
+REGION_FLAGS_JOINT_INFO_DICT = {
```
mike-w-wilson, line 330:
```suggestion
JOINT_REGION_FLAGS_INFO_DICT = {
```

mike-w-wilson (approved):
LGTM

PR #691, opened 2024-04-11
remove hardcoding of v4 in gnomad_gks

klaricch:
Just making this more flexible for the future, there's a warning a few lines above that gnomad_gks() is currently only implemented for gnomAD v4


mike-w-wilson (approved):
LGTM

PR #694, opened 2024-04-17
Add explode_downsamplings function

klaricch:


gnomad/utils/constraint.py
```diff
@@ -372,6 +372,47 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
+                gen_anc=pop,
+                downsampling=downsampling,
+                **{
+                    f"{metric}.exp": ht[metric]["gen_anc_exp"][pop][i]
+                    for metric in metrics
+                },
+                **{
+                    f"{metric}.obs": hl.if_else(
+                        hl.is_missing(ht[metric]["gen_anc_obs"][pop][i])
+                        & hl.is_defined(ht[metric]["gen_anc_obs"][pop][i]),
```
jkgoodrich, line 400: I'm confused, both are checking `ht[metric]["gen_anc_obs"][pop][i]` so this won't ever be True, right?

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,47 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
 
 
+def explode_downsamplings(
+    ht: hl.Table,
+    downsampling_meta: Dict[str, List[str]],
+    metrics: List[str] = ["syn", "lof", "mis"],
+) -> hl.Table:
+    """
+    Explode downsampling counts for each genetic ancestry group and metric.
+
```
jkgoodrich, line 382: It might help to have an example of the input and output, and required fields

jkgoodrich (changes requested):
A couple suggestions, and one question that likely requires a fix

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,47 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
+                        0,
+                        ht[metric]["gen_anc_obs"][pop][i],
+                    )
+                    for metric in metrics
+                },
+            )
+            for pop, downsamplings in downsampling_meta.items()
+            for i, downsampling in enumerate(downsamplings)
+        ]
+    ).select(*["data"])
```
jkgoodrich, line 410: This only need to be this right?

```suggestion
    ).select("data")
```

And actually, if just selecting, that can be done instead of `annotate` above right?

jkgoodrich (changes requested):
One more thing

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,47 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
+                    f"{metric}.obs": hl.if_else(
+                        hl.is_missing(ht[metric]["gen_anc_obs"][pop][i])
+                        & hl.is_defined(ht[metric]["gen_anc_obs"][pop][i]),
+                        0,
+                        ht[metric]["gen_anc_obs"][pop][i],
+                    )
+                    for metric in metrics
+                },
+            )
+            for pop, downsamplings in downsampling_meta.items()
```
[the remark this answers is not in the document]
klaricch: i was considering doing that but we use pop as a name in all the other functions, so was thinking could just change all instances of pop to gen_anc in a future pr

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,42 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
+    """
+    Explode downsampling counts for each genetic ancestry group and metric.
+
+    :param ht: Input Table. Should contain struct of downsampling information for genetic ancestry groups under each metric name. Example: 'lof': struct {gen_anc_exp: struct {global: array<float64>}.
+    :param metrics: List of metrics to explode. Default is '['syn', 'lof', 'mis']'.
+    :param downsampling_meta: Dictionary containing downsampling metadata. Keys are the genetic ancestry group names and values are the list of downsamplings for that genetic ancestry group. Example: {'global': ['5000', '10000'], 'afr': ['5000', '10000']}.
+    :return: Table with downsampling counts exploded so that observed and expected metric counts for each pair of genetic ancestry groups and downsamplings is a separate row.
+    """
+    ht = ht.select(
+        data=[
```
jkgoodrich, line 389:
```suggestion
        _data=[
```

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,42 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
+                    f"{metric}.obs": ht[metric]["gen_anc_obs"][pop][i]
+                    for metric in metrics
+                },
+            )
+            for pop, downsamplings in downsampling_meta.items()
+            for i, downsampling in enumerate(downsamplings)
+        ]
+    )
+    ht = ht.explode("data")
+    ht = ht.transmute(**ht.data)
```
jkgoodrich, line 407:
```suggestion
    ht = ht.explode("_data")
    ht = ht.transmute(**ht._data)
```

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,42 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
+                gen_anc=pop,
+                downsampling=downsampling,
+                **{
+                    f"{metric}.exp": ht[metric]["gen_anc_exp"][pop][i]
+                    for metric in metrics
+                },
+                **{
+                    f"{metric}.obs": ht[metric]["gen_anc_obs"][pop][i]
+                    for metric in metrics
+                },
```
jkgoodrich, line 400: Since these are not different anymore, can simplify with this:

```suggestion
                **{
                    f"{metric}.{oe}": ht[metric][f"gen_anc_{oe}"][pop][i]
                    for metric in metrics
                    for oe in ["exp", "obs"]
                },
```

However, in gnomad_lof it looks like it was:
```
hl.cond(hl.is_missing(ht[f'obs_lof_{pop}'][i]) & hl.is_defined(ht.obs_lof), 0, ht[f'obs_lof_{pop}'][i])
```
Do we need to check that the non pop obs is defined?

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,42 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
+    downsampling_meta: Dict[str, List[str]],
+    metrics: List[str] = ["syn", "lof", "mis"],
+) -> hl.Table:
+    """
+    Explode downsampling counts for each genetic ancestry group and metric.
+
+    :param ht: Input Table. Should contain struct of downsampling information for genetic ancestry groups under each metric name. Example: 'lof': struct {gen_anc_exp: struct {global: array<float64>}.
+    :param metrics: List of metrics to explode. Default is '['syn', 'lof', 'mis']'.
+    :param downsampling_meta: Dictionary containing downsampling metadata. Keys are the genetic ancestry group names and values are the list of downsamplings for that genetic ancestry group. Example: {'global': ['5000', '10000'], 'afr': ['5000', '10000']}.
+    :return: Table with downsampling counts exploded so that observed and expected metric counts for each pair of genetic ancestry groups and downsamplings is a separate row.
```
jkgoodrich, line 386: Just reformatting

```suggestion
    The input `ht` must contain struct of downsampling information for genetic ancestry 
    groups under each metric name. For example: 'lof': struct {gen_anc_exp: struct 
    {global: array<float64>}.

    :param ht: Input Table.
    :param metrics: List of metrics to explode. Default is '['syn', 'lof', 'mis']'.
    :param downsampling_meta: Dictionary containing downsampling metadata. Keys are the 
        genetic ancestry group names and values are the list of downsamplings for that 
        genetic ancestry group. Example: {'global': ['5000', '10000'], 'afr': ['5000', 
        '10000']}.
    :return: Table with downsampling counts exploded so that observed and expected 
        metric counts for each pair of genetic ancestry groups and downsamplings is a 
        separate row.
```

jkgoodrich (changes requested):
A couple more things

gnomad/utils/constraint.py
klaricch: with the way the pipeline is written i don't think we ever get missing values here

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,51 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
...
+                gen_anc=pop,
+                downsampling=downsampling,
+                **{
+                    f"{metric}.exp": ht[metric]["gen_anc_exp"][pop][i]
+                    for metric in metrics
+                },
+                **{
+                    f"{metric}.obs": ht[metric]["gen_anc_obs"][pop][i]
+                    for metric in metrics
+                },
```
jkgoodrich, line 409:
```suggestion
                **{
                    f"{metric}.{oe}": ht[metric][f"gen_anc_{oe}"][pop][i]
                    for oe in ["obs", "exp"]
                    for metric in metrics
                },
```

gnomad/utils/constraint.py
```diff
@@ -372,6 +372,51 @@ def _get_criteria(i: hl.expr.Int32Expression) -> hl.expr.Int32Expression:
     return hl.agg.array_sum(hl.map(_get_criteria, sorted_indices))
 
 
+def explode_downsamplings(
```
jkgoodrich, line 375: Maybe rename to be more specific. Something like `explode_downsamplings_oe`

jkgoodrich (commented):
Just two small things

jkgoodrich (approved):
LGTM!

PR #699, opened 2024-04-30
Change the number of values for stat_union_gen_ancs to unknown

KoalaQin:


ch-kr (approved):
thanks for adding!

also, for future tracking purposes, putting the rationale for this change here:

Qin ran `vcf-validator`, which flagged this error:
```
INFO field at chrY:56855635 .. Could not validate the float [Infinity],INFO tag [stat_union_gen_ancs=amr,mid,nfe,sas,eas,afr] expected different number of values (expected 1, found 6),Could not validate the float [Infinity]
```

Having some infinite values is expected for this stat (see [slack thread](https://atgu.slack.com/archives/CRA2TKTV0/p1713373207289189)), so we've ignored that warning. However, Qin clarified that the number of values per row for this annotation varies from 1-8 (number of gen anc groups in both exomes and genomes). Per the VCF 4.2 spec
image
`.` seems like a more appropriate value here than `A`.

Finally, Qin also clarified that the previous VCF header info dict was written with copilot (so no reason not to change `A` > `.` since `.` makes more sense)

PR #701, opened 2024-05-09
Add `get_summary_stats_variant_filter_expr` and `get_summary_stats_csq_filter_expr` to build filtering expressions for summary stats

jkgoodrich:


gnomad/utils/filtering.py
```diff
@@ -149,28 +149,30 @@ def filter_low_conf_regions(
     :param high_conf_regions: Paths to set of high confidence regions to restrict to (union of regions)
     :return: MatrixTable or Table with low confidence regions removed
```
matren395, line 150:
```suggestion
    :return: Bool expression of whether loci are not low confidence (TRUE) or low confidence (FALSE)
```

gnomad/utils/filtering.py
```diff
@@ -179,19 +181,36 @@ def filter_low_conf_regions(
...
-            mt = mt.filter_rows(filter_criteria)
-        else:
-            mt = mt.filter(filter_criteria)
+        return filter_criteria
+    else:
+        raise ValueError("No low confidence regions requested for filtering.")
+
+
+def filter_low_conf_regions(
+    mt: Union[hl.MatrixTable, hl.Table],
```
matren395, line 199:
```suggestion
    t: Union[hl.MatrixTable, hl.Table],
```

this type gets called 't' elsewhere - having an 'mt' possible actually be an 'ht' is a bit odd!

gnomad/utils/filtering.py
```diff
@@ -179,19 +181,36 @@ def filter_low_conf_regions(
...
+    **kwargs,
+) -> Union[hl.MatrixTable, hl.Table]:
+    """
+    Filter low-confidence regions.
+
+    :param mt: MatrixTable or Table to filter.
```
matren395, line 205:
```suggestion
    :param t: MatrixTable or Table to filter.
```

gnomad/utils/filtering.py
```diff
@@ -179,19 +181,36 @@ def filter_low_conf_regions(
...
+    :param kwargs: Keyword arguments to pass to `low_conf_regions_expr`.
+    :return: MatrixTable or Table with low confidence regions removed.
+    """
+    filter_criteria = low_conf_regions_expr(mt.locus, **kwargs)
```
matren395, line 209:
```suggestion
    filter_criteria = low_conf_regions_expr(t.locus, **kwargs)
```

gnomad/utils/filtering.py
```diff
@@ -179,19 +181,36 @@ def filter_low_conf_regions(
...
+    if isinstance(mt, hl.MatrixTable):
+        mt = mt.filter_rows(filter_criteria)
+    else:
+        mt = mt.filter(filter_criteria)
 
     return mt
```
matren395, line 215:
```suggestion
    if isinstance(t, hl.MatrixTable):
        t = t.filter_rows(filter_criteria)
    else:
        t = t.filter(filter_criteria)

    return t
```

gnomad/utils/annotations.py
```diff
@@ -90,6 +90,43 @@
 }
 
 
+def annotate_with_ht(
+    t: Union[hl.MatrixTable, hl.Table],
+    annotation_ht: hl.Table,
+    fields: Optional[List] = None,
+    annotate_cols: bool = False,
+    filter_missing: bool = False,
+) -> hl.MatrixTable:
```
matren395, line 99:
```suggestion
    t: Union[hl.MatrixTable, hl.Table],
    annotation_ht: hl.Table,
    fields: Optional[List] = None,
    annotate_cols: bool = False,
    filter_missing: bool = False,
) -> Union[hl.MatrixTable, hl.Table]:
```

gnomad/utils/annotations.py
```diff
@@ -90,6 +90,43 @@
...
+    """
+    Annotate a MatrixTable/Table with additional annotations from another Table.
+
+    :param t: MatrixTable/Table to be annotated.
+    :param annotation_ht: Table containing additional annotations to be joined on `t`.
+    :param fields: Optional list of fields to select from `annotation_ht` and add to `t`
+    :param annotate_cols: If True, annotate columns instead of rows. Default is False.
+    :param filter_missing: If True, filter out missing rows/cols in `t` that are not
+        present in `annotation_ht`. Default is False.
+    :return: Annotated MatrixTable.
```
matren395, line 109:
```suggestion
    :return: Annotated MatrixTable/Table.
```

if a table is the input, then the code should just return a table as well, yes?

matren395 (commented):
just pedantry about types and documentation - I'll look in a bit at the filter exprs themselves

gnomad/assessment/summary_stats.py
```diff
@@ -355,6 +356,185 @@ def get_tx_expression_expr(
...
+            max_af = [max_af]
+        for af in max_af:
+            log_list.append(f"variants with (AF < {af:.2e})")
+            ss_filter_expr[f"max_af_{af}"] = freq_expr < af
+    if min_an_proportion is not None:
+        log_list.append(
+            "variants that meet a minimum call rate of %.2f (using AN as a call rate "
+            "proxy)" % min_an_proportion,
+        )
+        ss_filter_expr[f"min_an_{max_af}"] = get_an_criteria(
```
matren395, line 405:
```suggestion
        ss_filter_expr[f"min_an_{min_an_proportion}"] = get_an_criteria(
```

unless I'm unclear about what this is doing here?

gnomad/assessment/summary_stats.py
```diff
@@ -355,6 +356,185 @@ def get_tx_expression_expr(
...
+    :param additional_csq_sets: Dictionary containing additional consequence sets.
+        Default is None.
+    :param additional_csqs: Set of additional consequences to keep. Default is None.
+    :param collapse_filters: Whether to collapse all filters into a single expression.
+        Default is False.
+    :return: BooleanExpression or Dict of BooleanExpressions for filtering consequences.
+    """
+    # Set up filters for specific consequences or sets of consequences.
+    csq_filters = {
+        **({"csq_set_lof": lof_csq_set or {}}),
```
matren395, line 453: defaults to {} if lof_csq_set is None?

matren395 (commented):
I think variant_filter_expr looks good to me, putting a pin in csq_filter_expr for now

gnomad/assessment/summary_stats.py
jkgoodrich: yup

gnomad/assessment/summary_stats.py
```diff
@@ -355,6 +356,185 @@ def get_tx_expression_expr(
...
+    }
+
+    # Add filtering expressions for HC LoF variants with no flags or any flags.
+    lof_hc_flags = {}
+    if lof_no_flags or lof_any_flags:
+        hc_expr = hl.or_else(t.lof == "HC", False)
+        if "no_lof_flags" in t.row:
+            no_lof_flags_expr = t.no_lof_flags
+        elif "lof_flags" in t.row:
+            no_lof_flags_expr = hl.is_missing(t.lof_flags) | (t.lof_flags == "")
```
matren395, line 502: why would there be different behavior in the input 't' here? Just curious

gnomad/assessment/summary_stats.py
```diff
@@ -355,6 +356,185 @@ def get_tx_expression_expr(
...
+        else:
+            raise ValueError(
+                "No LoF flag info found in input Table/MatrixTable/StructExpression."
+            )
+        if lof_no_flags:
+            lof_hc_flags["lof_HC_no_flags"] = hc_expr & no_lof_flags_expr
+        if lof_any_flags:
+            lof_flags["lof_HC_with_flags"] = hc_expr & ~no_lof_flags_expr
```
matren395, line 510: should we add 'only calculated for HC' to docstrings for 'lof_no_flags' and 'lof_any_flags' , or is this always the case?

matren395 (commented):
last few questions - think I've got a good handle on the PR now !

gnomad/assessment/summary_stats.py
```diff
@@ -355,6 +356,185 @@ def get_tx_expression_expr(
...
+    }
+
+    # Add filtering expressions for LoF consequence labels.
+    lof_labels = {
+        f"lof_{lof_label}": hl.or_else(t.lof == lof_label, False)
+        for lof_label in lof_label_set or []
+    }
+
+    # Add filtering expressions for LoF consequence flags.
+    lof_flags = {
```
KoalaQin, line 488: Even thought we're using it this time, do you think it would be nice to have `LOFTEE_FLAGS` in vep.py?

```commandline
LOFTEE_FLAGS = {
    'NAGNAG_SITE',
    'NO_EXON_NUMBER',
    'NON_CAN_SPLICE',
    'PHYLOCSF_UNLIKELY_ORF',
    'PHYLOCSF_WEAK',
    'SINGLE_EXON',
}
"""
Constant that contains LOFTEE flags.
"""
```

KoalaQin (changes requested):
I have two major comments: 
1. where the `lof_flags` is; 
2. if we could simplify the name of the keys, later the row name.

gnomad/assessment/summary_stats.py
```diff
@@ -355,6 +356,185 @@ def get_tx_expression_expr(
...
+
+    if not ss_filter_expr:
+        logger.warning("No filtering applied to consequences for summary stats.")
+        return True if collapse_filters else ss_filter_expr
+
+    # Collapse all filters into a single expression if requested.
+    if collapse_filters:
+        ss_filter_expr = functools.reduce(operator.iand, ss_filter_expr.values())
+
+    return ss_filter_expr
```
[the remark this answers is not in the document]
jkgoodrich: Yeah, I have a restructured version of `default_generate_gene_lof_matrix` that will come in a future PR and uses the collapse. I can remove it for now and add in that PR if that would be better

gnomad/assessment/summary_stats.py
```diff
@@ -355,6 +356,185 @@ def get_tx_expression_expr(
...
+        for lof_label in lof_label_set or []
+    }
+
+    # Add filtering expressions for LoF consequence flags.
+    lof_flags = {
+        f"lof_flag_{lof_flag}": hl.or_else(
+            t.lof_flags.split(",").contains(lof_flag), False
+        )
+        for lof_flag in lof_flag_set or []
+    }
```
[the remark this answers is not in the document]
jkgoodrich: I'm not actually sure why I added in the `lof_flag_set`, I think I read the lof_labels wrong and I had added it, so then I just went with it. I'm going to just remove it for now, and we can add back when we have a use for it

gnomad/assessment/summary_stats.py
jkgoodrich: I could be wrong, but I think this is needed because of a version change. It used to only need to check for "", but now that doesn't work and it needs to check for missingness. This will allow it to work for either version

gnomad/assessment/summary_stats.py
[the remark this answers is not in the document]
jkgoodrich: removed for now, we can add in another PR if/when needed

gnomad/assessment/summary_stats.py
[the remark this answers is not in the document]
jkgoodrich: removed for now, I will add that option in another PR when updating `default_generate_gene_lof_matrix`

gnomad/assessment/summary_stats.py
jkgoodrich: Removed that option for now, so we can just add if/when needed

KoalaQin (changes requested):
Small things, you can move the 2 functions here.

gnomad/assessment/summary_stats.py
```diff
@@ -355,6 +354,181 @@ def get_tx_expression_expr(
     )
 
 
+def get_summary_stats_variant_filter_expr(
+    t: Union[hl.Table, hl.MatrixTable],
+    filter_lcr: bool = False,
+    filter_expr: hl.expr.SetExpression = None,
```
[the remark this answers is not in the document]
jkgoodrich: I don't think that the longer name is needed, we often use `filter_expr` in our code, and in this context I don't see how it would be confused with the function name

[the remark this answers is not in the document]
KoalaQin: Okay, I was looking at gnomad_qc, where you had filter_exprs={...} but you had a note there. So fine with me.

PR #703, opened 2024-05-24
Update VEP csqs in impact categories to match VEP

mike-w-wilson:
This moves `start_lost` from medium to high impact and `splice_region_variant` from medium to low impact. These match the current ensembl table.


ch-kr (approved):
one comment update then LGTM!

PR #711, opened 2024-06-14
Update csq impact lists with v105 splice terms

mike-w-wilson:
Adds: splice_donor_5th_base_variant, splice_donor_region_variant , and splice_polypyrimidine_tract_variant

### DIFF
--- a/gnomad/utils/vep.py
+++ b/gnomad/utils/vep.py
@@ -44,7 +44,10 @@ CSQ_CODING_MEDIUM_IMPACT = [
 ]
 
 CSQ_CODING_LOW_IMPACT = [
+    "splice_donor_5th_base_variant",  # new in v105
     "splice_region_variant",  # Considered low impact in v105, previously medium
+    "splice_donor_region_variant",  # new in v105
+    "splice_polypyrimidine_tract_variant",  # new in v105
     "incomplete_terminal_codon_variant",
     "start_retained_variant",  # new in v92
     "stop_retained_variant",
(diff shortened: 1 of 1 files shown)


ch-kr (approved):
these changes LGTM. I have a few notes below of edits if we want to be fully consistent with https://www.ensembl.org/info/genome/variation/prediction/predicted_data.html?redirect=no:

HIGH:
- transcript_amplification (previously MEDIUM)

These are also now HIGH but I don't love moving these into high impact:
- feature_elongation
- feature_truncation

MODIFIER:
- coding_sequence_variant (previously LOW impact)
- coding_transcript_variant (looks new)
- sequence_variant (looks new)

ch-kr:
yeah, I agree we need to decide whether to imitate VEP or just loosely follow their mapping with some adjustments. I was initially thinking we'd stay consistent with VEP, but I don't think that actually serves us in our potential downstream applications using these groups, so I vote we do the latter (use all of VEP's listed consequence terms and adjust the associated impacts where needed).

maybe we should move `start_lost` and `transcript_amplification` back to medium, keep `feature_elongation`/`feature_truncation`/`coding_sequence_variant` where they are currently (non-coding for the first two and low for the last one), add `coding_transcript_variant` to `CSQ_CODING_LOW_IMPACT` , and add `sequence_variant` to `CSQ_NON_CODING`?

^I know this is more complex than the initial PR, so I'd be happy to merge this PR and start the discussion via slack to finalize which terms should go where

mike-w-wilson:
@ch-kr  That sounds good to me. I've made updates to the PR. Would you still like to start the slack discussion?

ch-kr (approved):

ch-kr:
let's merge -- I have a meeting with Kaitlin next week and will plan to ask her about these then (and move into a larger public channel if needed)

PR #713, opened 2024-06-25
Add `filter_vep_transcript_csqs_expr`, a version of `filter_vep_transcript_csqs` that takes and returns an ArrayExpression

jkgoodrich:


gnomad/utils/vep.py
```diff
@@ -802,9 +802,69 @@ def filter_vep_transcript_csqs(
     vep_root: str = "vep",
     synonymous: bool = True,
     canonical: bool = True,
-    mane_select: bool = False,
-    filter_empty_csq: bool = True,
     ensembl_only: bool = True,
+    filter_empty_csq: bool = True,
```
mike-w-wilson, line 806: Thoughts on updating to `remove_empty_csq` ? 
```suggestion
    remove_empty_csq: bool = True,
```

mike-w-wilson: Looks like we'd only need to update gnomad_qc in one place and then there would be a bulk replace in your other VEP PR. I just think remove is more clear since filter is a bit overloaded and can be ambiguous as to whether you are keeping or removing the filtered material.

gnomad/utils/vep.py
```diff
@@ -802,9 +802,69 @@ def filter_vep_transcript_csqs(
...
+    If `filter_empty_csq` parameter is set to True, the Table/MatrixTable is filtered
+    to variants where 'transcript_consequences' within the VEP annotation is not empty
+    after the specified filtering criteria is applied.
+
+    .. note::
+
+        By default, the Table/MatrixTable is filtered to variants where
+        'transcript_consequences' within the VEP annotation is not empty after filtering
+        to Ensembl canonical transcripts with a most severe consequence of
+        'synonymous_variant'.
```
mike-w-wilson, line 821:
```suggestion
    Filter VEP transcript consequences based on specified criteria, and optionally remove variants where transcript consequences is empty after filtering.

    If `remove_empty_csq` parameter is set to True, the Table/MatrixTable is filtered
    to variants where 'transcript_consequences' within the VEP annotation is not empty
    after the specified filtering criteria is applied.

    .. note::

        By default, the Table/MatrixTable is filtered to variants where
        'transcript_consequences' within the VEP annotation is not empty after filtering
        to Ensembl canonical transcripts with a most severe consequence of
        'synonymous_variant'.
```
If the above suggestion is accepted.

gnomad/utils/vep.py
```diff
@@ -802,9 +802,69 @@ def filter_vep_transcript_csqs(
...
+    :param vep_root: Root for VEP annotation. Default is 'vep'.
+    :param synonymous: Whether to filter to variants where the most severe consequence
+        is 'synonymous_variant'. Default is True.
+    :param canonical: Whether to filter to only canonical transcripts. Default is True.
+    :param ensembl_only: Whether to filter to only Ensembl transcripts. This option is
+        useful for deduplicating transcripts that are the same between RefSeq and
+        Emsembl. Default is True.
+    :param filter_empty_csq: Whether to filter out rows where 'transcript_consequences'
+        is empty, after filtering 'transcript_consequences' to the specified criteria.
+        Default is True.
```
mike-w-wilson, line 833:
```suggestion
    :param remove_empty_csq: Whether to remove rows where 'transcript_consequences'
        is empty, after filtering 'transcript_consequences' to the specified criteria.
        Default is True.
```

gnomad/utils/vep.py
```diff
@@ -802,9 +802,69 @@ def filter_vep_transcript_csqs(
...
+                synonymous=synonymous,
+                canonical=canonical,
+                ensembl_only=ensembl_only,
+                **kwargs,
+            )
+        )
+    }
+    t = t.annotate_rows(**vep_data) if is_mt else t.annotate(**vep_data)
+
+    if filter_empty_csq:
```
mike-w-wilson, line 852:
```suggestion
    if remove_empty_csq:
```

gnomad/utils/vep.py
mike-w-wilson: Dependent on arg rename suggestion

gnomad/utils/vep.py
mike-w-wilson: Same comment as above

mike-w-wilson (commented):
This split makes sense to me here. I have a suggestion on an arg name though I know this PR was a rearrangement but since were updating the module I put it in. I think its more clear and should be an quick update elsewhere.

gnomad/utils/vep.py
jkgoodrich: I think it makes sense to do it in it's own PR where that is all that is done since it's not a name introduced in this PR, and it is technically a breaking change

mike-w-wilson (approved):

PR #714, opened 2024-06-25
Deprecate `get_most_severe_consequence_for_summary` in favor of more flexible `get_most_severe_csq_from_multiple_csq_lists`

jkgoodrich:
Also adds:
- `filter_to_most_severe_consequences`, which is used by `get_most_severe_csq_from_multiple_csq_lists`
- Adds `loftee_labels` and `no_lof_flags` parameters to `filter_vep_transcript_csqs_expr` for filtering by loftee labels and flags.

Depends on #713 

[updates_to_get_most_severe_consequence_for_summary.html.zip](https://github.com/user-attachments/files/15976998/updates_to_get_most_severe_consequence_for_summary.html.zip) testing to make sure the same results are returned for `get_most_severe_consequence_for_summary`


tests/utils/test_vep.py
```diff
@@ -0,0 +1,249 @@
...
+                biotype="intergenic",
+                lof=hl.missing(hl.tstr),
+                lof_flags=hl.missing(hl.tstr),
+                consequence_terms="intergenic_variant",
+            )
+        ],
+    )
+
+    # Test the function
+    result = get_most_severe_csq_from_multiple_csq_lists(vep_expr)
```
klaricch, line 211: also tested with setting prioritize_loftee_no_flags to True or False -- does it make sense that got the same result regardless of how set this param was set?

gnomad/utils/vep.py
```diff
@@ -741,61 +883,147 @@ def get_most_severe_consequence_for_summary(
...
+    add_order_by_csq_list = add_order_by_csq_list or {}
+    loftee_labels = loftee_labels or LOFTEE_LABELS
+
+    # Create a struct with missing values for each VEP consequences list.
+    ms_csq_list_expr = hl.struct(
+        **{c: hl.missing(vep_expr[c].dtype) for c in csq_list_order if c in vep_expr}
     )
 
+    # Build the case expression to determine the most severe consequence.
+    ms_csq_expr = hl.case(missing_false=True)
```
klaricch, line 979: why use case expressions?

gnomad/utils/vep.py
```diff
@@ -720,12 +718,156 @@ def get_csq_from_struct(
...
+
+        # Filter to only consequences with the highest impact csq label, and return
+        # missing if the most severe consequence is missing.
+        return ms_csq, hl.or_missing(
+            (hl.len(expr) > 0) & hl.is_defined(ms_csq),
+            expr.filter(lambda x: filter_func(x[field], ms_csq)),
+        )
+
+    # Initialize the lof struct with missing values.
+    lof_expr = hl.struct(lof=hl.missing(hl.tstr), no_lof_flags=hl.missing(hl.tbool))
```
klaricch, line 825: so lof and no_lof_flags will always be None if prioritize_loftee/prioritize_loftee_no_flags are False, even if lof flags are present?

klaricch (changes requested):

gnomad/utils/vep.py
```diff
@@ -720,12 +718,156 @@ def get_csq_from_struct(
...
+    the order of `loftee_labels`, over those without LOFTEE annotations. If
+    `prioritize_loftee_no_flags` is True, prioritize LOFTEE consequences with no flags
+    over those with flags.
+
+    If `additional_order` is provided, additional ordering is applied to the
+    consequences in the list after any of the above prioritization. An example use of
+    this parameter is to prioritize by PolyPhen predictions.
+
+    :param csq_expr: ArrayExpression of VEP consequences to filter.
+    :param csq_order: List indicating the order of VEP consequences, sorted from high to
```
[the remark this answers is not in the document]
jkgoodrich: To avoid using mutable defaults

gnomad/utils/vep.py
jkgoodrich: Is there a reason not to? If you have a clearer solution, I'm happy to make modifications, just let me know what your thinking

gnomad/utils/vep.py
```diff
@@ -720,12 +718,156 @@ def get_csq_from_struct(
...
+    additional_order: Optional[List[str]] = None,
+) -> hl.StructExpression:
+    """
+    Filter an array of VEP consequences to all entries that have the most severe consequence.
+
+    Returns a struct with the following annotations:
+
+        - most_severe_consequence: Most severe consequence for variant.
+        - lof: Whether the variant is a loss-of-function variant.
+        - no_lof_flags: Whether the variant has any LOFTEE flags (True if no flags).
```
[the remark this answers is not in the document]
jkgoodrich: I made some changes, so I'm not sure what the result was before, but now in my test on 2 partitions of the exomes result HT, I get `{False: 21, True: 580, None: 78785}`. Let me know if you are still seeing no False

gnomad/utils/vep.py
jkgoodrich: I modified this a bit, let me know if the changes make sense

tests/utils/test_vep.py
```diff
@@ -12,238 +15,589 @@
...
+            (True, False, True, None, None, None, None),
+            (True, False, False, *polyphen_params, None, "possibly_damaging"),
+            (False, True, True, None, None, "stop_gained", None),
+            (False, True, False, *polyphen_params, None, "possibly_damaging"),
+            (False, False, True, *polyphen_params, None, None),
+            (True, True, True, None, None, "stop_gained", None),
+            (True, True, False, *polyphen_params, None, "possibly_damaging"),
+            (True, False, True, *polyphen_params, None, "possibly_damaging"),
+            (False, True, True, *polyphen_params, "stop_gained", None),
+            # Need to figure out class too large error
```
klaricch, line 241: i also got this error

klaricch (changes requested):

PR #718, opened 2024-07-25
Add functions `fill_missing_key_combinations` and `missing_struct_expr`

jkgoodrich:
fill_missing_key_combinations: Fill missing key combinations with requested fill values.
missing_struct_expr: Create a struct of missing values for each field and type in the input struct.

Also added test code for both

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -2408,18 +2408,20 @@ def fill_missing_key_combinations(
 
     Example::
 
-    .. code-block:: python
+        .. code-block:: python
 
-        ht = hl.Table.parallelize(
-                [{'key1': 'A', 'key2': 1, 'value': 10},
-                {'key1': 'A', 'key2': 2, 'value': 20},
-                {'key1': 'B', 'key2': 1, 'value': 30}],
+            ht = hl.Table.parallelize(
+                [
+                    {'key1': 'A', 'key2': 1, 'value': 10},
+                    {'key1': 'A', 'key2': 2, 'value': 20},
+                    {'key1': 'B', 'key2': 1, 'value': 30},
+                ],
                 hl.tstruct(key1=hl.tstr, key2=hl.tint32, value=hl.tint32),
-                key=['key1', 'key2']
+                key=['key1', 'key2'],
             )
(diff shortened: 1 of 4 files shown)


gnomad/utils/annotations.py
```diff
@@ -2388,6 +2388,70 @@ def _update_struct(
...
+    This function fills missing key combinations in the input Table with requested fill
+    values. The fill values are specified in the `fill_values` dictionary. The unique
+    values for each key field are collected from the input Table unless specified in
+    the `key_values` dictionary.
+
+    This is useful when you want to ensure that all possible key combinations are
+    present in a Table, even if they are missing from the input Table. This can
+    happen when you are aggregating data and want to ensure that all possible key
+    combinations are present in the output Table, not only the ones that are present.
+
```
mike-w-wilson, line 2408:
```suggestion

    Example:

    .. code-block:: python

        ht = hl.Table.parallelize(
                [{'key1': 'A', 'key2': 1, 'value': 10},
                {'key1': 'A', 'key2': 2, 'value': 20},
                {'key1': 'B', 'key2': 1, 'value': 30}],
                hl.tstruct(key1=hl.tstr, key2=hl.tint32, value=hl.tint32),
                key=['key1', 'key2']
            )
        fill_values = {'value': hl.missing(hl.tint32)}
        filled_ht = fill_missing_key_combinations(ht, fill_values)
        filled_ht.show()

        +------+------+
        | key1 | key2 | value |
        +------+------+
        |  A   |  1   |  10   |
        |  A   |  2   |  20   |
        |  B   |  1   |  30   |
        |  B   |  2   | null  |
        +------+------+

    In this example, the input table is missing the combination (B, 2).
    After applying `fill_missing_key_combinations`, the missing key combination
    (B, 2) is filled with the specified fill value for 'value' (null in this case).

```

mike-w-wilson: I thought an example would be helpful here since its a bit abstract so I asked ChatGPT. I think this is a reasonable example.

mike-w-wilson (approved):
Just to doc suggestions, then this LGTM!

mike-w-wilson:
I am having trouble getting the tests to run with my IDE but I think this is a me problem

PR #721, opened 2024-07-30
pext runtime changes

jkgoodrich:


gnomad/utils/transcript_annotation.py
```diff
@@ -416,10 +489,152 @@ def perform_tx_annotation_pipeline(
...
+    ht = ht.key_by(*[r for r in ht.row_value if r != "pos"]).collect_by_key("pos")
+
+    # Annotate with 'start' and 'stop' positions for regions by merging adjacent
+    # positions.
+    ht = ht.annotate(
+        pos=hl.fold(
+            lambda i, j: hl.if_else(
+                j.pos > (i[-1][1] + 1),
+                i.append((j.pos, j.pos)),
+                i[:-1].append((i[-1][0], j.pos)),
```
ch-kr, line 625: does this line exist just to make sure the if and else clauses both return Tuples? this line doesn't do anything, right?

gnomad/utils/transcript_annotation.py
```diff
@@ -331,7 +398,10 @@ def tx_annotate_variants(
         **ht[vep_annotation],
         **tx_ht[ht[vep_annotation].transcript_id, ht[vep_annotation].gene_id],
     )
-    ht = ht.annotate_globals(tissues=tissues)
+    ht = ht.annotate_globals(
+        tissues=tissues,
+        exp_prop_mean_tissues=exp_prop_mean_tissues,
```
ch-kr, line 403: this will annotate with two of the same list if the user doesn't specify values for `tissues_to_exclude_from_mean`, right? it's a bit extra code, but this could be:
```
if tissues == exp_prop_mean_tissues:
    return ht.annotate_globals(tissues=tissues)
return ht.annotate_globals(
    tissues=tissues,
    exp_prop_mean_tissues=exp_prop_mean_tissues,
)
```

gnomad/utils/transcript_annotation.py
```diff
@@ -401,7 +472,9 @@ def perform_tx_annotation_pipeline(
         `{vep_root}.{vep_annotation}`.
     :param tx_ht: Table of transcript expression information.
     :param tissues_to_filter: Optional list of tissues to exclude from the output.
-    :param vep_root: Name used for root VEP annotation. Default is 'vep'.
+        Default is None.
+    :param tissues_to_exclude_from_mean: Optional list of tissues to exclude when
+        calculating the mean expression proportion across all tissues. Default is None.    :param vep_root: Name used for root VEP annotation. Default is 'vep'.
```
ch-kr, line 477:
```suggestion
        Default is None.
    :param tissues_to_exclude_from_mean: Optional list of tissues to exclude when
        calculating the mean expression proportion across all tissues. Default is None.
    :param vep_root: Name used for root VEP annotation. Default is 'vep'.
```

gnomad/utils/transcript_annotation.py
```diff
@@ -416,10 +489,152 @@ def perform_tx_annotation_pipeline(
         ),
         tx_ht,
         tissues_to_filter=tissues_to_filter,
+        tissues_to_exclude_from_mean=tissues_to_exclude_from_mean,
         vep_root=vep_root,
         vep_annotation=vep_annotation,
     )
 
+    if "alleles" not in additional_group_by:
+        tx_ht = tx_ht.key_by("locus", "transcript_id").distinct()
```
ch-kr, line 498: is each value the same? `distinct` chooses one of the values at random to deduplicate, right?

gnomad/utils/transcript_annotation.py
```diff
@@ -416,10 +489,152 @@ def perform_tx_annotation_pipeline(
...
+
+    for char in tissue_name:
+        if char.isupper():
+            if len(formatted_name) > 0 and formatted_name[-1] != "_":
+                formatted_name += "_"
+            formatted_name += char.lower()
+        else:
+            formatted_name += char
+
+    replacements = {
```
ch-kr, line 529:
```suggestion
    # Dictionary of tissue names that need to be reformatted
    # that will not be formatted correctly with for loop above
    replacements = {
```

gnomad/utils/transcript_annotation.py
```diff
@@ -416,10 +489,152 @@ def perform_tx_annotation_pipeline(
...
+
+
+def create_tx_annotation_by_region(ht: hl.Table) -> hl.Table:
+    """
+    Create transcript annotation by region for loading into the gnomAD browser.
+
+    This function processes a Hail Table to create transcript annotations by region.
+    It calculates the mean expression proportion, handles missing values, and organizes
+    the data by genomic regions. The key fields are 'gene_id', 'exp_prop_mean', and
+    'tissues', ensuring regions split correctly based on changes in these fields.
```
ch-kr, line 557:
```suggestion
    'tissues', ensuring regions split are correctly based on changes in these fields.
```

ch-kr: though this sentence makes it sound like these are the final keys. maybe:
```
Regions are split based on changes in the following fields: `gene_id`, `exp_prop_mean`, and `tissues`.
```

tests/utils/test_transcript_annotation.py
```diff
@@ -0,0 +1,157 @@
...
+            ("SpinalCord", "spinal_cord"),
+            ("CervicalC", "cervical_c"),
+            ("SubstantiaNigra", "substantia_nigra"),
+            ("CulturedFibroblasts", "cultured_fibroblasts"),
+            ("LowerLeg", "lower_leg"),
+            ("TransformedLymphocytes", "transformed_lymphocytes"),
+            ("AnteriorCingulateCortex", "anterior_cingulate_cortex"),
+            ("BA24", "ba24"),
+            ("BA9", "ba9"),
+            ("EBV", "ebv"),
```
ch-kr, line 29: shouldn't one of these tests be checking for an underscore in the name (like `b_a24`)?

tests/utils/test_transcript_annotation.py
```diff
@@ -0,0 +1,157 @@
...
+class TestCleanTissueNameForBrowser:
+    """Tests for the clean_tissue_name_for_browser function."""
+
+    @pytest.mark.parametrize(
+        "input_name,expected_output",
+        [
+            ("BasalGanglia", "basal_ganglia"),
+            ("NucleusAccumbens", "nucleus_accumbens"),
+            ("SpinalCord", "spinal_cord"),
+            ("CervicalC", "cervical_c"),
```
ch-kr, line 21: naive question, shouldn't one of these test the case where the entire word is lower case (like `basalganglia`)?

gnomad/utils/transcript_annotation.py
```diff
@@ -416,10 +489,152 @@ def perform_tx_annotation_pipeline(
...
+        |         | {'chrom': 'chr1', 'start': 6, 'stop': 6, 'mean': 0.8,        |
+        |         | 'tissues': {'tissue1': 0.6, 'tissue2': 0.7}}                 |
+        +---------+--------------------------------------------------------------+
+
+    :param ht: Input Hail Table with transcript expression information.
+    :return: Hail Table with transcript annotations by region.
+    """
+    # Get a list of tissues and drop the 'tissues' field from the Table.
+    tissues = hl.eval(ht.tissues)
+    ht = ht.drop("tissues")
```
ch-kr, line 600: do you need any of the other globals as well? couldn't you also drop `grange`, `vep_help`, `vep_config`?

ch-kr (changes requested):
a few basic questions and minor changes

docs/_static/theme_overrides.css
```diff
@@ -1,3 +1,15 @@
...
 }
+
+/* override table width restrictions */
+.wy-table-responsive table td, .wy-table-responsive table th {
+    white-space: pre-wrap;
+    word-wrap: break-word;
+}
+
+.wy-table-responsive {
+    margin-bottom: 24px;
```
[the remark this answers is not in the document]
jkgoodrich: It was a copy paste and the result looked good

gnomad/utils/transcript_annotation.py
jkgoodrich: yeah, I guess those are really just because they are on the context table, whereas tissues was an expected global on the input ht, so that is all I was thinking about. I will switch to a `.select_globals()`

gnomad/utils/transcript_annotation.py
jkgoodrich: I think I like having them both on there so I can just use `exp_prop_mean_tissues` if I want to in downstream use rather than needing to check for it's existence and then use `tissues` if it doesn't exist

gnomad/utils/transcript_annotation.py
jkgoodrich: yeah, that's a valid question. As is, it's really just to support the creation of the base-level file, and in that case this does work fine, but in other cases where more than only `gene_symbol` is provided, it might not work. I will modify to something I think is correct, let me know what you think.

gnomad/utils/transcript_annotation.py
jkgoodrich: I think it does do something though. It's like this example in regular python 
```
pos = [6, 7, 8, 10, 11, 12]
start = [(pos[0], pos[0])]
print("start:", start)

regions = start[:]
for current_pos in pos[1:]:
    if current_pos > (regions[-1][1] + 1):
        regions.append((current_pos, current_pos))
    else:
        last_region_first_position = regions[-1][0]
        regions = regions[:-1]
        regions.append((last_region_first_position, current_pos))

print("Final regions:", regions)
```

It prints this:
```
start: [(6, 6)]
Final regions: [(6, 8), (10, 12)]
```

gnomad/utils/transcript_annotation.py
```diff
@@ -361,6 +433,18 @@ def tx_aggregate_variants(
...
+    # keyed by locus and alleles so that the correct transcripts are retained when
+    # filtering variants by transcript (with `tx_filter_variants_by_csqs`) and exploding
+    # the VEP annotation (in `tx_annotate_variants`). However, if "alleles" is not
+    # present in the additional_group_by, a transcript can be counted multiple times if
+    # it is associated with multiple alleles since the transcript expression is keyed
+    # by transcript_id and gene_id. To avoid this, we re-key the tx_ht by locus,
+    # gene_id, transcript_id, and any additional_group_by fields, followed by a distinct
+    # operation to ensure that each transcript is only counted once per locus.
+    if "alleles" not in additional_group_by:
+        ht = ht.key_by(*grouping, "transcript_id").distinct()
```
ch-kr, line 446: sorry for the confusion -- I was initially trying to ask whether the values at each loci for each transcript were the same.

I'm a bit confused now; I think the previous method (of keying by just locus and transcript) deduplicates, this new method could count a transcript multiple times per locus. For example, if `additional_group_by = "lof"`, then grouping at this point would be `["locus", "gene_id", "lof"]`. That means HT would be keyed by `locus, gene_id, lof, transcript_id` here, which means a transcript could still get counted multiple times if one variant in the transcript is pLoF and another variant isn't, right? let me know if I've missed something

gnomad/utils/transcript_annotation.py
ch-kr: ohh thank you for this example! I must have done something weird in my testing, since it looked like it wasn't appending anything

ch-kr (commented):
one question about the deduplicating transcript values per loci

gnomad/utils/transcript_annotation.py
jkgoodrich: > sorry for the confusion -- I was initially trying to ask whether the values at each loci for each transcript were the same.

Yeah, I figured that's what you were asking, and the values for the expression proportions will be the same, but other columns can be different, so by doing the distinct I could be removing some of the values that need to be counted. It's not a problem for the main case where we use the function, but if people change up the additional_grouping, it could return the wrong result.

Here is an example of when the previous method would have failed:

`tx_ht`:
Screenshot 2024-08-14 at 9 53 43 AM

`context_ht` (after exploding by transcript):
Screenshot 2024-08-14 at 9 53 47 AM

annotate `context_ht` with `tx_ht`:
Screenshot 2024-08-14 at 9 53 52 AM

Then if we do the distinct it in the way we would for the base level, it's basically this:
```
always_grouping = ["locus", "gene_id"]
additional_grouping = []
grouping = always_grouping + additional_grouping

context_ht = context_ht.key_by("locus", "transcript_id").distinct()
```
Screenshot 2024-08-14 at 9 53 57 AM

And then the aggregation:
```
agg_ht = context_ht.group_by(*grouping).aggregate(
    sum_exp_prop_mean=hl.agg.sum(context_ht.exp_prop_mean)
)
```
Screenshot 2024-08-14 at 9 54 02 AM

That is fine because, it doesn't matter which `lof` we grab.



However, if we change the `additional_grouping` to include "lof":
```
always_grouping = ["locus", "gene_id"]
additional_grouping = ["lof"]
grouping = always_grouping + additional_grouping

context_ht = context_ht.key_by("locus", "transcript_id").distinct()
```
Screenshot 2024-08-14 at 10 00 27 AM

The aggregation gives:
```
agg_ht = context_ht.group_by(*grouping).aggregate(
    sum_exp_prop_mean=hl.agg.sum(context_ht.exp_prop_mean)
)
```
Screenshot 2024-08-14 at 10 00 31 AM

What we want is: 
```
context_ht = context_ht.key_by(*grouping, "transcript_id").distinct()
```
Screenshot 2024-08-14 at 10 02 16 AM

So the aggregation gives: 
Screenshot 2024-08-14 at 10 02 20 AM

ch-kr: thank you for this! it sounds like the concern isn't about accidentally duplicating transcript representation at a locus but more about making sure the values are represented as requested by the user with `additional_groupings` -- maybe this comment:
```
    # Previous steps of the transcript annotation pipeline require that the input ht is
    # keyed by locus and alleles so that the correct transcripts are retained when
    # filtering variants by transcript (with `tx_filter_variants_by_csqs`) and exploding
    # the VEP annotation (in `tx_annotate_variants`). However, if "alleles" is not
    # present in the additional_group_by, a transcript can be counted multiple times if
    # it is associated with multiple alleles since the transcript expression is keyed
    # by transcript_id and gene_id. To avoid this, we re-key the tx_ht by locus,
    # gene_id, transcript_id, and any additional_group_by fields, followed by a distinct
    # operation to ensure that each transcript is only counted once per locus.
```
could be edited to be:
```
    # Previous steps of the transcript annotation pipeline require that the input ht is
    # keyed by locus and alleles so that the correct transcripts are retained when
    # filtering variants by transcript (with `tx_filter_variants_by_csqs`) and exploding
    # the VEP annotation (in `tx_annotate_variants`). However, if "alleles" is not
    # present in the additional_group_by, a transcript that is associated with multiple alleles
    # at a locus may not be represented correctly after deduplicating locus and transcript
    # combinations with a distinct operation, since this operation deduplicates by selecting
    # a random row. To ensure the desired values are selected during `distinct()` filter, 
    # we re-key the tx_ht by locus, gene_id, transcript_id, and any additional_group_by fields
    # prior to running `distinct()'.
```
or another rewording?

ch-kr (approved):
one comment rewording suggestion, otherwise LGTM!

gnomad/utils/transcript_annotation.py
```diff
@@ -283,6 +283,7 @@ def tx_filter_variants_by_csqs(
     ignore_splicing: bool = True,
     filter_to_protein_coding: bool = True,
     vep_root: str = "vep",
+    include_polyphen_prioritization: bool = False,
```
ch-kr, line 286: is this just to avoid the extra stratification from `process_consequences`?

ch-kr (approved):
one question but LGTM

gnomad/utils/transcript_annotation.py
jkgoodrich: it's because v2 has polyphen on it and v4 doesn't, and process consequences defaults to use polyphen so it fails on the v4 sites HTs

PR #726, opened 2024-08-19
Update default vep version for context table resource

ch-kr:
Update default VEP version from 95 to 105, which is the version we used for gnomAD v4/v4.1


klaricch (approved):
lgtm

PR #727, opened 2024-08-23
Fix `tx_filter_variants_by_csqs` to correctly handle the `ignore_splicing` parameter

jkgoodrich:
As is, the `ignore_splicing` parameter will remove transcripts from transcript_consequences where the worst consequence in "consequence_terms" is a splicing consequence. However, we want to keep the transcript if it has another consequence in "consequence_terms" that is not a splicing consequence. This PR updates the function so that splicing consequences are removed from the "consequence_terms" before the most severe consequence is added.

Example:
```
 "transcript_consequences": [
    {
        "consequence_terms": [
            "splice_region_variant",
            "synonymous_variant"
        ]
    }
]
```

Currently this transcript would get removed, but we want it to be kept as:
```
 "transcript_consequences": [
    {
        "consequence_terms": ["synonymous_variant"],
        "most_severe_consequence": "synonymous_variant",
    }
]
```

### DIFF
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -328,13 +328,15 @@ def tx_filter_variants_by_csqs(
         ht = ht.annotate(
             **{
                 vep_root: ht[vep_root].annotate(
-                    transcript_consequences=ht[vep_root].transcript_consequences.map(
+                    transcript_consequences=ht[vep_root]
+                    .transcript_consequences.map(
                         lambda csq: csq.annotate(
                             consequence_terms=csq.consequence_terms.filter(
                                 lambda x: ~hl.literal(CSQ_SPLICE).contains(x)
                             )
                         )
-                     ).filter(lambda csq: hl.len(csq.consequence_terms) > 0)
+                    )
+                    .filter(lambda csq: hl.len(csq.consequence_terms) > 0)
                 )
             }
         )
--- a/gnomad/utils/transcript_annotation.py
+++ b/gnomad/utils/transcript_annotation.py
@@ -9,6 +9,7 @@ from gnomad.utils.filtering import filter_to_gencode_cds
 from gnomad.utils.vep import (
     CSQ_CODING,
     CSQ_SPLICE,
+    CSQ_ORDER,
     explode_by_vep_annotation,
     filter_vep_transcript_csqs,
     process_consequences,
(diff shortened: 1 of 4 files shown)


tests/utils/test_transcript_annotation.py
```diff
@@ -192,3 +193,188 @@ def test_create_tx_annotation_by_region(self, sample_hail_table: hl.Table) -> No
...
+                            "transcript_id": "ENST00000342066",
+                            "consequence_terms": [
+                                "splice_region_variant",
+                                "synonymous_variant",
+                            ],
+                            "amino_acids": "V",
+                            "biotype": "protein_coding",
+                            "lof": None,
+                            "lof_flags": None,
+                            "canonical": 0,
```
ch-kr, line 220: it shouldn't matter for testing, but the browser lists ENST00000342066 as the canonical transcript for this gene

tests/utils/test_transcript_annotation.py
```diff
@@ -192,3 +193,188 @@ def test_create_tx_annotation_by_region(self, sample_hail_table: hl.Table) -> No
...
+        result_ht = tx_filter_variants_by_csqs(
+            mock_vep_annotated_ht,
+            filter_to_cds=False,
+            ignore_splicing=True,
+            filter_to_protein_coding=False,
+        )
+        assert result_ht.count() == 4
+
+    def test_filter_to_protein_coding(self, mock_vep_annotated_ht):
+        """Test filtering to protein coding variants."""
```
ch-kr, line 373:
```suggestion
        """Test filtering to protein coding transcripts."""
```

tests/resources/test_resource_utils.py
```diff
@@ -14,6 +14,13 @@
 )
 
 
+@pytest.fixture(autouse=True)
+def reset_gnomad_public_resource_configuration():
```
ch-kr, line 18: naive question, why do you need this?

ch-kr (commented):
a couple questions

tests/resources/test_resource_utils.py
jkgoodrich: The tests below set `gnomad_public_resource_configuration` to different things for testing, and this sets it back to None so that the other tests can run correctly. I was very confused for awhile when the tests I added kept coming back with an error about the gencode resource not being found in "gs://my-bucket/gnomad-resources"

tests/utils/test_transcript_annotation.py
jkgoodrich: yeah, I just put random things past what was needed for the gencode filter

jkgoodrich: I wouldn't have put real gene and transcript names if I didn't need to test the gencode filtering

gnomad/utils/transcript_annotation.py
```diff
@@ -323,28 +323,38 @@ def tx_filter_variants_by_csqs(
...
+                vep_root: ht[vep_root].annotate(
+                    transcript_consequences=ht[vep_root]
+                    .transcript_consequences.map(
+                        lambda csq: csq.annotate(
+                            consequence_terms=csq.consequence_terms.filter(
+                                lambda x: ~hl.literal(CSQ_SPLICE).contains(x)
+                            )
+                        )
+                    )
+                    .filter(lambda csq: hl.len(csq.consequence_terms) > 0)
```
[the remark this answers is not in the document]
jkgoodrich: I actually started to add it to the vep utils, but then decided against it because that was clearly filtering based on most severe consequence, and I thought it might be confusing to have both.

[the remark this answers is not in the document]
jkgoodrich: This fix is to make it do what was done in the original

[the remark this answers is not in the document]
ch-kr: yeah, I keep going back and forth on whether this (filtering on any csq) is helpful to have, but maybe you're right that it might be too confusing

ch-kr (approved):
LGTM!

PR #734, opened 2024-10-03
Add `get_mu_annotation_expr` function that prevents a shuffle from happening when annotating a HT with mutation rate and use in `annotate_with_mu`

jkgoodrich:


KoalaQin (approved):
LGTM!

PR #740, opened 2024-11-28
Add `get_loftee_end_trunc_filter_expr` and `update_loftee_end_trunc_filter` to patch the LOFTEE END_TRUNC filter

jkgoodrich:
Tested in this notebook
[test_loftee_end_trunc.html.zip](https://github.com/user-attachments/files/17950701/test_loftee_end_trunc.html.zip)


gnomad/utils/vep.py
```diff
@@ -1003,3 +1003,88 @@ def explode_by_vep_annotation(
...
+
+    The end truncation filter is based on the GERP distance cutoff (`gerp_dist_cutoff`)
+    and the 'GERP_DIST' and '50_BP_RULE' annotations in the LOFTEE annotations.
+
+    True is returned if the GERP distance is less than the cutoff and the '50_BP_RULE'
+    annotation is not 'PASS'.
+
+    :param csq_expr: StructExpression containing the LOFTEE annotation 'lof_info', with
+        'GERP_DIST' and '50_BP_RULE' info.
+    :param gerp_dist_cutoff: GERP distance cutoff for end truncation. Default is 0.0.
```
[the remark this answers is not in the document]
KoalaQin: I only checked variants in BRCA2 gene, the last 2 LOF now have "END_TRUNC" filter and became "LC".

KoalaQin (commented):
I have two minor comments.

gnomad/utils/vep.py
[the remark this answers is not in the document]
jkgoodrich: Yup, I'm hoping Julia can help me add to the documentation to justify the cutoff, and to also confirm that the patch is being done correctly.

gnomad/utils/vep.py
```diff
@@ -1003,3 +1003,88 @@ def explode_by_vep_annotation(
...
+
+    def _update_csq_struct(csq_expr: hl.expr.StructExpression):
+        """
+        Update the LOFTEE end truncation filter in the input Struct.
+
+        :param csq_expr: Struct containing the LOFTEE annotations.
+        :return: Consequence Struct with updated LOFTEE annotations.
+        """
+        end_trunc_expr = get_loftee_end_trunc_filter_expr(csq_expr, gerp_dist_cutoff)
+        filter_expr = hl.or_else(
```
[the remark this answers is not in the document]
jkgoodrich: The `lof_filter` is a list of filters represented as a string delimited by `,`, for example: "5UTR_SPLICE,ANC_ALLELE,GC_TO_GT_DONOR"

If `end_trunc_expr` is True, I add `END_TRUNC` to the filters, but it might already be in the filters and I only want it represented once.

gnomad/utils/vep.py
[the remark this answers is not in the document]
jkgoodrich: I added to the documentation, also here is a histogram of the GERP distance


Screenshot 2025-09-24 at 4 09 52 PM

tests/utils/test_vep.py
```diff
@@ -0,0 +1,257 @@
...
+            ),
+        )
+
+        # Apply the function.
+        ht = ht.annotate(updated_csq=update_loftee_end_trunc_filter(ht.csq))
+
+        # Collect results.
+        results = ht.collect()
+
+        assert results[0].updated_csq.lof_filter == "END_TRUNC"
```
klaricch, line 254: should this be ",END_TRUNC"?

tests/utils/test_vep.py
```diff
@@ -0,0 +1,257 @@
...
+
+        # Still LC because filter is not empty.
+        assert results[1].updated_csq.lof_filter == "SINGLE_EXON"
+        assert results[1].updated_csq.lof == "LC"
+
+        # Still LC because filter is not empty.
+        assert results[2].updated_csq.lof_filter == "SINGLE_EXON"
+        assert results[2].updated_csq.lof == "LC"
+
+        assert results[3].updated_csq.lof_filter == "END_TRUNC"
```
klaricch, line 160: should this be ",END_TRUNC"?

tests/utils/test_vep.py
```diff
@@ -0,0 +1,257 @@
...
+
+        # Apply the function.
+        ht = ht.annotate(updated_csq=update_loftee_end_trunc_filter(ht.csq))
+
+        # Collect results.
+        results = ht.collect()
+
+        # Should handle missing lof gracefully.
+        assert results[0].updated_csq.lof_filter == "END_TRUNC,SINGLE_EXON"
+        assert results[0].updated_csq.lof is None
```
klaricch, line 228: would we want this to still be None? or LC?

tests/utils/test_vep.py
jkgoodrich: I thought I made a fix to main function so it doesn't anymore. I think we want it to be just "END_TRUNC"

tests/utils/test_vep.py
jkgoodrich: same comment

tests/utils/test_vep.py
jkgoodrich: this one shouldn't actually happen, there shouldn't be a case where the filter is defined and lof is not. I can remove it, or just switch to all missing and make sure it handles them correctly

klaricch: I don't think it would  happen either, but maybe can just add a comment here saying we don't expect it?

tests/utils/test_vep.py
```diff
@@ -0,0 +1,257 @@
...
+                lof="HC",
+            ),
+            # Case 3: Should remove existing END_TRUNC filter.
+            hl.Struct(
+                lof_info="GERP_DIST:1.0,50_BP_RULE:PASS",
+                lof_filter="SINGLE_EXON,END_TRUNC",
+                lof="LC",
+            ),
+            # Case 4: Should add END_TRUNC, but will show ",END_TRUNC" due to empty
+            # filter handling.
```
klaricch:
```suggestion
            # Case 4: Should add END_TRUNC
```

tests/utils/test_vep.py
jkgoodrich: Added a comment

PR #741, opened 2024-12-06
Add support for filtering Hail Tables to `filter_to_trios`

jkgoodrich:


gnomad/sample_qc/relatedness.py
```diff
@@ -986,31 +986,34 @@ def maximal_independent_set_keep_samples(
...
-           the variants that are not present in any of the trios.
 
-    :param mtds: A Variant Dataset or a Matrix Table to filter to only trios
-    :param fam_ht: A Table of trios to filter to, loaded using `hl.import_fam`
-    :return: A Matrix Table or a Variant Dataset with only the trios in `fam_ht`
+        Using `filter_cols` in MatrixTable will not affect the number of rows
+        (variants), however, using `filter_samples` in VariantDataset will remove the
+        variants that are not present in any of the trios.
+
+    :param mtds: A Variant Dataset or a Matrix Table to filter to only trios.
```
KoalaQin, line 1000:
```suggestion
    :param mtds: A Variant Dataset or a Matrix Table or a Table to filter to only trios.
```

gnomad/sample_qc/relatedness.py
```diff
@@ -986,31 +986,34 @@ def maximal_independent_set_keep_samples(
...
+    :return: A Table, MatrixTable or VariantDataset with only the trios in `fam_ht`.
     """
     # Filter to samples present in any of the trios.
     fam_ht = fam_ht.annotate(fam_members=[fam_ht.id, fam_ht.pat_id, fam_ht.mat_id])
     fam_ht = fam_ht.explode("fam_members", name="s")
     fam_ht = fam_ht.key_by("s").select().distinct()
 
-    if isinstance(mtds, hl.MatrixTable):
+    if isinstance(mtds, hl.Table):
+        mtds = mtds.semi_join(fam_ht)
```
KoalaQin, line 1010: Is this mtds key_by 's', like a meta_ht instead? Maybe put a note?

KoalaQin (changes requested):

gnomad/sample_qc/relatedness.py
jkgoodrich: I don't think the key needs to be 's' for this to work, the key just needs to be the sample names in fam_ht, but that's the same assumption for the matrix table and vds columns

KoalaQin: Okay, got it.

KoalaQin (approved):
LGTM!

PR #742, opened 2024-12-09
Add import code for GTEx v10 RSEM

KoalaQin:
This copies the import function as for GRCh37 v7, but modified a bit according the current path.

### DIFF
--- a/gnomad/resources/grch38/reference_data.py
+++ b/gnomad/resources/grch38/reference_data.py
@@ -5,6 +5,7 @@ from typing import Optional, Union
 import hail as hl
 from hail import Table
 
+from gnomad.resources.grch37.reference_data import _import_gtex_rsem
 from gnomad.resources.resource_utils import (
     DBSNP_B154_CHR_CONTIG_RECODING,
     NO_CHR_TO_CHR_CONTIG_RECODING,
@@ -17,7 +18,6 @@ from gnomad.resources.resource_utils import (
 )
 from gnomad.utils.constraint import transform_methylation_level
 from gnomad.utils.vep import vep_or_lookup_vep
-from gnomad.resources.grch37.reference_data import _import_gtex_rsem
 
 
 def _import_purcell_5k(path) -> hl.Table:
--- a/gnomad/resources/grch38/reference_data.py
+++ b/gnomad/resources/grch38/reference_data.py
@@ -437,6 +437,7 @@ def get_truth_ht() -> Table:
         .persist()
     )
 
+
 gtex_rsem = VersionedMatrixTableResource(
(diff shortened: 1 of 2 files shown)


gnomad/resources/grch38/reference_data.py
```diff
@@ -19,6 +19,55 @@
 from gnomad.utils.vep import vep_or_lookup_vep
 
 
+def _import_gtex_rsem(gtex_path: str, meta_path: str, **kwargs) -> hl.MatrixTable:
```
jkgoodrich, line 22: Is this exactly the same as the one in grch37? If so, we should probably just import it from there

gnomad/utils/filtering.py
```diff
@@ -460,6 +463,17 @@ def filter_to_gencode_cds(
         "This Gencode CDS interval filter does not filter by transcript! Please see the"
         " documentation for more details to confirm it's being used as intended."
     )
+    if padding:
```
jkgoodrich, line 466: Maybe put this in another PR since it's not related to the resource

jkgoodrich (commented):
Just 2 things, but looks good

gnomad/resources/grch38/reference_data.py
KoalaQin: It is, I just wonder if you used this function to get that MT, it seems the same structure as I see in the MT.

KoalaQin: also because it's giving a warning this underscore named function: access to a protected member * of a module

jkgoodrich: Yeah, don't worry about the warning, or remove the `_`. Let me confirm, but I think it's the same function

jkgoodrich: I did this 
```
from gnomad.resources.grch37.reference_data import _import_gtex_rsem

try:
    mt = hl.read_matrix_table("gs://gnomad/resources/grch38/gtex/gtex_rsem_v10.mt")
except:
    mt = _import_gtex_rsem(
        gtex_path="gs://gnomad/resources/gtex/v10/GTEx_Analysis_2022-06-06_v10_RSEMv1.3.3_transcripts_tpm.txt.bgz",
        meta_path="gs://gnomad/resources/gtex/v10/GTEx_Analysis_v10_Open_Access_Reduced_Annotations_SampleAttributesDS.txt.bgz",
        min_partitions=1000,
    
    ).checkpoint("gs://gnomad/resources/grch38/gtex/gtex_rsem_v10.mt", _read_if_exists=True)
```

jkgoodrich (approved):
LGTM!

PR #746, opened 2024-12-19
Add function `parse_variant` to create a Struct with the locus and alleles from a variant string or contig, position, ref, and alt.

jkgoodrich:


tests/utils/test_parse.py
```diff
@@ -0,0 +1,80 @@
...
+
+class TestParseVariant:
+    """Test the parse_variant function."""
+
+    grch37_out = hl.Struct(locus=hl.Locus("1", 1000, "GRCh37"), alleles=["A", "T"])
+    grch38_out = hl.Struct(locus=hl.Locus("chr1", 1000, "GRCh38"), alleles=["A", "T"])
+
+    @pytest.mark.parametrize(
+        "variant_str, contig, position, ref, alt, build, expected",
+        [
```
KoalaQin, line 17: I've never used pytest myself, you only need to put some representative valid and invalid sennarios, right?

KoalaQin (commented):
Looks good, why did you add pytest for this one but not our other code?

tests/utils/test_parse.py
jkgoodrich: Yeah, it's just adding unit tests so that if we change the code and it breaks it, then hopefully the test will fail

jkgoodrich:
I just forgot to add tests to the other PR where I changed a function. We are trying to add tests now whenever we add a new function or change a function

KoalaQin (approved):
LGTM!

PR #747, opened 2024-12-19
Filter to Gencode CDS by genes and by exon paddings

KoalaQin:
This will filter the Gencode CDS further optionally by genes and by adding a padding.


gnomad/utils/filtering.py
```diff
@@ -403,7 +403,10 @@ def filter_to_clinvar_pathogenic(
 
 
 def filter_to_gencode_cds(
-    t: Union[hl.MatrixTable, hl.Table], gencode_ht: Optional[hl.Table] = None
+    t: Union[hl.MatrixTable, hl.Table],
+    gencode_ht: Optional[hl.Table] = None,
+    genes: Optional[Union[str, List[str]]] = None,
+    padding: Optional[int] = 0,
```
jkgoodrich, line 409: Maybe padding_bp

gnomad/utils/filtering.py
```diff
@@ -436,6 +439,8 @@ def filter_to_gencode_cds(
     :param gencode_ht: Gencode Table to use for filtering the input Table/MatrixTable
         to CDS regions. Default is None, which will use the default version of the
         Gencode Table resource.
+    :param genes: Gene(s) to filter to. Default is None, which will filter to all genes.
```
jkgoodrich, line 442:
```suggestion
    :param genes: Optional gene(s) to filter to. Can be a single gene string or list of genes. Default is None.
```

jkgoodrich, line 442: Might want to support either gene symbol or gene id? Either way, specify which one is expected

gnomad/utils/filtering.py
```diff
@@ -460,12 +465,39 @@ def filter_to_gencode_cds(
...
+                gencode_ht.interval.start.position - padding,
+                gencode_ht.interval.end.position + padding,
+                includes_start=gencode_ht.interval.includes_start,
+                includes_end=gencode_ht.interval.includes_end,
+                reference_genome=gencode_ht.interval.start.dtype.reference_genome,
+            )
+        )
+        gencode_ht = gencode_ht.key_by("padded_interval")
+
+    if genes or padding:
```
jkgoodrich, line 489: I'm not sure about this. I think even if padding is true and genes is false, it will have that issue. What if instead we have a parameter that specifies the number of intervals we are willing to collect and set the default to something that filter_intervals would still be the better option

gnomad/utils/filtering.py
```diff
@@ -460,12 +465,39 @@ def filter_to_gencode_cds(
...
+            padded_interval=hl.locus_interval(
+                gencode_ht.interval.start.contig,
+                gencode_ht.interval.start.position - padding,
+                gencode_ht.interval.end.position + padding,
+                includes_start=gencode_ht.interval.includes_start,
+                includes_end=gencode_ht.interval.includes_end,
+                reference_genome=gencode_ht.interval.start.dtype.reference_genome,
+            )
+        )
+        gencode_ht = gencode_ht.key_by("padded_interval")
```
jkgoodrich, line 487: I think this should only be done if needed, so move below into the `else`

jkgoodrich (changes requested):
A few thoughts. Also, can you undo the changes to the init_scripts? That can get fixed in a different PR, but I don't think it really belongs in this one

KoalaQin:
These init_scripts were changed by my pre-commit, fix end of files, it will be done even when I undo it when I commit again. What do I do?

gnomad/utils/filtering.py
KoalaQin: I think this is a good idea, TTN seems to have the most exons, 363, how about 1000 intervals? Would that cause memory issue based on your experience?

jkgoodrich:
You should be able to disable the pre-commit hook. But also, I think it's only supposed to run on the files that have been modified in the current branch

gnomad/utils/filtering.py
jkgoodrich: I'm not sure, I haven't ever really tried

KoalaQin: I set it to 3000, based on the distribution of CDS per gene and the runtime in my notebook: 
Screenshot 2024-12-20 at 2 54 29 PM

Screenshot 2024-12-20 at 2 51 19 PM
lmk what you think.

KoalaQin: 2 minutes is not so bad?

jkgoodrich: works for me

gnomad/utils/filtering.py
```diff
@@ -436,6 +441,10 @@ def filter_to_gencode_cds(
     :param gencode_ht: Gencode Table to use for filtering the input Table/MatrixTable
         to CDS regions. Default is None, which will use the default version of the
         Gencode Table resource.
+    :param genes: Optional gene(s) to filter to. Can be a single gene string or list of genes. Default is None.
+    :param by_gene_symbol: Whether to filter by gene symbol. Default is True. If False, will filter by gene ID.
+    :param padding_bp: Number of bases to pad the CDS intervals by. Default is 0.
+    :param max_intervals: Maximum number of intervals to collect before filtering. Default is 1000.
```
jkgoodrich, line 447:
```suggestion
    :param max_collect_intervals: Maximum number of intervals for the use of `hl.filter_intervals` for filtering. When the number of intervals to filter is greater than this number, `filter`/`filter_rows` will be used instead. The reason for this is because `hl.filter_intervals` is faster, but when the number of intervals is too large, this can cause memory errors. Default is 3000.
```

gnomad/utils/filtering.py
```diff
@@ -460,12 +469,51 @@ def filter_to_gencode_cds(
         "This Gencode CDS interval filter does not filter by transcript! Please see the"
         " documentation for more details to confirm it's being used as intended."
     )
-    filter_expr = hl.is_defined(gencode_ht[t.locus])
+    if genes:
+        genes_upper = (
+            [genes.upper()] if isinstance(genes, str) else [g.upper() for g in genes]
+        )
```
jkgoodrich, line 475:
```suggestion
        genes = hl.literal(
            [genes.upper()] if isinstance(genes, str) else [g.upper() for g in genes]
        )
```

gnomad/utils/filtering.py
```diff
@@ -460,12 +469,51 @@ def filter_to_gencode_cds(
...
+        if by_gene_symbol:
+            gencode_ht = gencode_ht.filter(
+                hl.literal(genes_upper).contains(gencode_ht.gene_name)
+            )
+        else:
+            gencode_ht = gencode_ht.filter(
+                hl.literal(genes_upper).contains(gencode_ht.gene_id)
+            )
```
jkgoodrich, line 483:
```suggestion
        gene_field = "gene_name" if by_gene_symbol else "gene_id"
        gencode_ht = gencode_ht.filter(genes.contains(gencode_ht[gene_field]))
```

KoalaQin:
I used --no-verify to my commit to disable the precommit, but it's still like this for 2 files

gnomad/utils/filtering.py
```diff
@@ -460,12 +469,51 @@ def filter_to_gencode_cds(
...
+        gencode_ht = gencode_ht.annotate(
+            padded_interval=hl.locus_interval(
+                gencode_ht.interval.start.contig,
+                gencode_ht.interval.start.position - padding_bp,
+                gencode_ht.interval.end.position + padding_bp,
+                includes_start=gencode_ht.interval.includes_start,
+                includes_end=gencode_ht.interval.includes_end,
+                reference_genome=gencode_ht.interval.start.dtype.reference_genome,
+            )
+        )
```
jkgoodrich, line 495:
```suggestion
    interval_expr = gencode_ht.interval
    if padding_bp:
        interval_expr = hl.locus_interval(
            interval_expr.start.contig,
            interval_expr.start.position - padding_bp,
            interval_expr.end.position + padding_bp,
            includes_start=interval_expr.includes_start,
            includes_end=interval_expr.includes_end,
            reference_genome=interval_expr.start.dtype.reference_genome,
        )
```

gnomad/utils/filtering.py
```diff
@@ -460,12 +469,51 @@ def filter_to_gencode_cds(
...
+
+    # Collect intervals if there are more than `max_intervals` to avoid memory issues
```
jkgoodrich, line 497:
```suggestion
    # Only collect intervals if there are less than or equal to `max_intervals` to avoid memory issues.
```

gnomad/utils/filtering.py
```diff
@@ -460,12 +469,51 @@ def filter_to_gencode_cds(
...
+    if gencode_ht.count() < max_intervals:
```
jkgoodrich, line 498: num_intervals = gencode_ht.count()
```suggestion
    if num_intervals <= max_collect_intervals:
```

gnomad/utils/filtering.py
```diff
@@ -403,7 +403,12 @@ def filter_to_clinvar_pathogenic(
 
 
 def filter_to_gencode_cds(
-    t: Union[hl.MatrixTable, hl.Table], gencode_ht: Optional[hl.Table] = None
+    t: Union[hl.MatrixTable, hl.Table],
+    gencode_ht: Optional[hl.Table] = None,
+    genes: Optional[Union[str, List[str]]] = None,
+    by_gene_symbol: bool = True,
+    padding_bp: Optional[int] = 0,
+    max_intervals: Optional[int] = 3000,
```
jkgoodrich, line 411:
```suggestion
    max_collect_intervals: int = 3000,
```

gnomad/utils/filtering.py
```diff
@@ -403,7 +403,12 @@ def filter_to_clinvar_pathogenic(
 
 
 def filter_to_gencode_cds(
-    t: Union[hl.MatrixTable, hl.Table], gencode_ht: Optional[hl.Table] = None
+    t: Union[hl.MatrixTable, hl.Table],
+    gencode_ht: Optional[hl.Table] = None,
+    genes: Optional[Union[str, List[str]]] = None,
+    by_gene_symbol: bool = True,
+    padding_bp: Optional[int] = 0,
```
jkgoodrich, line 410:
```suggestion
    padding_bp: int = 0,
```

gnomad/utils/filtering.py
```diff
@@ -460,12 +469,51 @@ def filter_to_gencode_cds(
...
+                includes_end=gencode_ht.interval.includes_end,
+                reference_genome=gencode_ht.interval.start.dtype.reference_genome,
+            )
+        )
+
+    # Collect intervals if there are more than `max_intervals` to avoid memory issues
+    if gencode_ht.count() < max_intervals:
+        logger.info(
+            "Since %d is smaller than the max intervals that can be collected, collecting all intervals...",
+            gencode_ht.count(),
```
jkgoodrich, line 501:
```suggestion
            "Since %d is less than or equal to 'max_collect_intervals', collecting all intervals...",
            num_intervals,
```

gnomad/utils/filtering.py
```diff
@@ -460,12 +469,51 @@ def filter_to_gencode_cds(
...
+        )
+        cds_intervals = (
+            gencode_ht.padded_interval if padding_bp else gencode_ht.interval
+        )
+        cds_intervals = cds_intervals.collect()
```
jkgoodrich, line 506:
```suggestion
        cds_intervals = interval_expr.collect()
```

gnomad/utils/filtering.py
```diff
@@ -460,12 +469,51 @@ def filter_to_gencode_cds(
...
+        t = hl.filter_intervals(t, cds_intervals)
     else:
-        t = t.filter(filter_expr)
+        if padding_bp:
+            gencode_ht = gencode_ht.key_by("padded_interval")
```
jkgoodrich, line 510:
```suggestion
            gencode_ht = gencode_ht.key_by(padded_interval=interval_expr)
```

jkgoodrich (changes requested):

gnomad/utils/filtering.py
```diff
@@ -436,6 +441,15 @@ def filter_to_gencode_cds(
     :param gencode_ht: Gencode Table to use for filtering the input Table/MatrixTable
         to CDS regions. Default is None, which will use the default version of the
         Gencode Table resource.
+    :param genes: Optional gene(s) to filter to. Can be a single gene string or list of genes. Default is None.
+    :param by_gene_symbol: Whether to filter by gene symbol. Default is True. If False, will filter by gene ID.
+    :param padding_bp: Number of bases to pad the CDS intervals by. Default is 0.
+    :param max_collect_intervals: Maximum number of intervals for the use of
+         `hl.filter_intervals` for filtering. When the number of intervals to filter is
+         greater than this number, `filter`/`filter_rows` will be used instead. The
+         reason for this is that`hl.filter_intervals` is faster, but when the
```
jkgoodrich, line 450:
```suggestion
         reason for this is that `hl.filter_intervals` is faster, but when the
```

gnomad/utils/filtering.py
```diff
@@ -460,12 +474,44 @@ def filter_to_gencode_cds(
...
+            interval_expr.end.position + padding_bp,
+            includes_start=interval_expr.includes_start,
+            includes_end=interval_expr.includes_end,
+            reference_genome=interval_expr.start.dtype.reference_genome,
+        )
 
-    if isinstance(t, hl.MatrixTable):
-        t = t.filter_rows(filter_expr)
+    # Only collect intervals if there are less than or equal to
+    # `max_intervals` to avoid memory issues.
```
jkgoodrich, line 497:
```suggestion
    # `max_collect_intervals` to avoid memory issues.
```

jkgoodrich (approved):
two small things, but then lgtm!

PR #749, opened 2025-01-02
Add coverage release global in GRCh37

KoalaQin:
This is to add the versions of released coverage HT as globals.


ch-kr (changes requested):
thanks for adding this! only comment is to update the `coverage` function now that you've added these constants

ch-kr (approved):
LGTM!

PR #760, opened 2025-01-28
Adapt `hl.de_novo` function

KoalaQin:
This PR introduces modifications based on **Julia’s work** on integrating [hl.de_novo](https://hail.is/docs/0.2/methods/genetics.html#hail.methods.de_novo) with Kaitlin Samocha’s [de_novo_finder](https://github.com/ksamocha/de_novo_scripts/blob/master/de_novo_finder_3.py). 

**Key Updates in This Implementation**

Compared to previous versions (`hl.de_novo` and Kaitlin’s implementation), this approach introduces **notable differences**:
1.	**De Novo Calling for XY Individuals on ChrX Non-PAR and ChrY Non-PAR**
	•	The **proband must be homozygous alternate (hom_alt**) and the **parent must be homozygous reference (hom_ref)** in these regions.
	•	The **PL values used for posterior probability calculation** follow this same logic.
	•	This differs from `hl.de_novo` but is consistent with Kaitlin’s implementation.
2.	**Confidence Level Assignments (HIGH, MEDIUM, LOW)**
	•	There are **discrepancies** in confidence level definitions across `hl.de_novo`, Kaitlin’s code, and this implementation.
	•	A **comparison summary** is provided in the attached screenshot for reference.
Screenshot 2025-02-12 at 1 55 37 PM

3.	**More Permissive Approach**
	•	This implementation is **generally more permissive**, particularly for **SNVs.**
	•	**No AC criteria is applied at this step**, unlike `hl.de_novo` and Kaitlin’s code.
	•	This is intentional, as the method is designed for a cohort that may contain **multiple trios per family**, where **recurrent de novo mutations (DNMs) are possible**.


gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+        hl.case()
+        .when(not_hemi_expr, father_expr.DP + mother_expr.DP)
+        .when(hemi_x_expr, mother_expr.DP)
+        .when(hemi_y_expr, father_expr.DP)
+        .or_missing()
+    )
+    dp_ratio = proband_expr.DP / parent_dp
+
+    # Key metrics
+    proband_ab = proband_expr.AD[1] / hl.sum(proband_expr.AD)
```
ch-kr, line 1614: this function is assuming that multiallelics have been split, right? if yes, that assumption should be noted in the docstring

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+    parent_dp = (
+        hl.case()
+        .when(not_hemi_expr, father_expr.DP + mother_expr.DP)
+        .when(hemi_x_expr, mother_expr.DP)
+        .when(hemi_y_expr, father_expr.DP)
+        .or_missing()
+    )
+    dp_ratio = proband_expr.DP / parent_dp
+
+    # Key metrics
```
ch-kr, line 1613: should this comment be something more descriptive, like "Calculate proband AB and assign variant type"?

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+                & (
+                    (proband_expr.DP > dp_threshold_snp)
+                    | (dp_ratio > high_conf_dp_ratio)
+                )
+            )
+            | (~is_snp & (p_de_novo > high_conf_p) & (proband_ab > high_med_conf_ab)),
+            "HIGH",
+        )
+        .when((p_de_novo > med_conf_p) & (proband_ab > high_med_conf_ab), "MEDIUM")
+        .when((p_de_novo > low_conf_p) & (proband_ab >= low_conf_ab), "LOW")
```
ch-kr, line 1634: I didn't dig into Kaitlin's code, but her documentation uses `>` and not `>=` for AB
```
p_dn > 0.05 and child_AD > 0.2
```

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+        father_expr.PL,
+        mother_expr.PL,
+        locus_expr,
+        is_female_expr,
+        freq_prior_expr,
+        min_pop_prior=min_pop_prior,
+        de_novo_prior=de_novo_prior,
+    )
+
+    # Determine genomic context
```
ch-kr, line 1587: rather than running this in this function and in `calculate_de_novo_post_prob`, why not switch the order and run `get_genomic_context` first and pass in the three return values to `calculate_de_novo_post_prob` as function arguments?

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+        .when((p_de_novo > med_conf_p) & (proband_ab > high_med_conf_ab), "MEDIUM")
+        .when((p_de_novo > low_conf_p) & (proband_ab >= low_conf_ab), "LOW")
+        .when(
+            (p_de_novo >= min_de_novo_p),
+            "VERY LOW",
+        )
+        .or_missing()
+    )
+
+    parent_sum_ad_0 = (
```
ch-kr, line 1642: these aren't sums; should this be renamed `parent_ad_0`, `parent_ad_0_check`, or something else that's similar?

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+        .when(hemi_y_expr, father_expr.AD[1] / hl.sum(father_expr.AD) > max_parent_ab)
+        .or_missing()
+    )
+
+    # Fail checks
+    fail_checks = {
+        "min_dp_ratio": dp_ratio < min_dp_ratio,
+        "parent_sum_ad_0": parent_sum_ad_0,
+        "max_parent_ab": fail_max_parent_ab,
+        "min_proband_ab": proband_ab < min_proband_ab,
```
ch-kr, line 1667: should this be `<=` instead of `<`?

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+        min_pop_prior=min_pop_prior,
+        de_novo_prior=de_novo_prior,
+    )
+
+    # Determine genomic context
+    not_hemi_expr, hemi_x_expr, hemi_y_expr = get_genomic_context(
+        locus_expr, is_female_expr
+    )
+
+    is_de_novo = (
```
ch-kr, line 1592: the current setup means you're calculating this probability on all variants, right? could you filter to variants that are eligible for being de novos first and then calculate probabilities? or is there a reason you don't want to do this filter upfront?

ch-kr (changes requested):
thanks for adding these functions! I have some questions (many because I'm not that familiar with this work) and suggestions

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,373 @@ def _get_alt_count(locus, gt, is_female):
     )
 
     return sib_stats
+
+
+def get_freq_prior(freq_prior_expr: hl.expr.Float64Expression, min_pop_prior=100 / 3e7):
```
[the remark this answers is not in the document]
KoalaQin: It's like this in the original code, I was thinking, maybe they want to differentiate from the 1/3e7 for the de novo prior?

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,373 @@ def _get_alt_count(locus, gt, is_female):
...
+    """
+    Get the population frequency prior for a de novo mutation.
+
+    :param freq_prior_expr: The population frequency prior for the variant.
+    :param min_pop_prior: The minimum population frequency prior.
+    """
+    return hl.max(
+        hl.or_else(
```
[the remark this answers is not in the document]
KoalaQin: Yeah, it's equivalent, I merged two steps from Julia to get this one.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,373 @@ def _get_alt_count(locus, gt, is_female):
...
+                    freq_prior_expr,
+                )
+            ),
+            0.0,
+        ),
+        min_pop_prior,
+    )
+
+
+def transform_pl_to_pp(pl_expr: hl.expr.ArrayExpression) -> hl.expr.ArrayExpression:
```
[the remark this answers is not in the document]
KoalaQin: It's the conditional probabilities of each geneotype given the data calculated by HaplotypeCaller, we're transforming it back.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,373 @@ def _get_alt_count(locus, gt, is_female):
...
+
+    :param pl_expr: ArrayExpression of PL.
+    :return: ArrayExpression of the probability of observing each genotype.
+    """
+    return hl.bind(lambda x: x / hl.sum(x), 10 ** (-pl_expr / 10))
+
+
+def get_genomic_context(
+    locus_expr: hl.expr.LocusExpression,
+    is_female_expr: hl.expr.BooleanExpression,
```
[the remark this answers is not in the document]
KoalaQin: The Hail trio_matrix function will return is_female in the MT, what do we do with that?

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,373 @@ def _get_alt_count(locus, gt, is_female):
...
+
+    :param locus_expr: LocusExpression of the variant.
+    :param is_female_expr: BooleanExpression indicating whether the proband is female.
+    :return: A tuple of BooleanExpressions:
+        - not_hemi_expr: True if the variant is in autosomes or PAR regions.
+        - hemi_x_expr: True if the variant is in the X non-PAR region for males.
+        - hemi_y_expr: True if the variant is in the Y non-PAR region for males.
+    """
+    not_hemi_expr = locus_expr.in_autosome_or_par() | (
+        locus_expr.in_x_nonpar() & is_female_expr
```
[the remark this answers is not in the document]
KoalaQin: I think this is for XX individuals in the nonPAR regions.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+
+    This function computes the posterior probability of a de novo mutation (P_dn)
+    using the likelihoods of the proband and parents' genotypes and the population
+    frequency prior for the variant.
+
+    Based on Kaitlin Samocha's de novo caller (
+    https://github.com/ksamocha/de_novo_scripts),
+    the posterior probability of a de novo mutation (P_dn) is computed as:
+
+        P_dn = P(DN | data) / (P(DN | data) + P(missed het in parent(s) | data))
```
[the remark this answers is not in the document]
KoalaQin: Kaitlin's README or docstring didn't include the hemi calculation, maybe that's why Hail didn't make that part right. I saw the Hail docstring for math, it's just hard to read for human eyes. I could convert it to LaTex format for Sphinx if you think that's the standard.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+        P(DN | data) = P(data | DN) * P(DN)
+
+    P(data | DN): The probability of observing the data under the assumption of a de novo mutation:
+       * Autosomes and PAR regions:
+
+          P(data | DN) = P(hom_ref in father) * P(hom_ref in mother) * P(het in proband)
+
+       * X non-PAR regions (males only):
+
+          P(data | DN) = P(hom_ref in mother) * P(het in proband)
```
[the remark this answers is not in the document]
KoalaQin: You're right, it should be hom_alt or hom_var. Yes, we do have het in chrX non-PAR for XY samples. My version and Julia's versions are wrong.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+
+    P(missed het in parent(s) | data): The probability of observing missed het in
+    parent(s) given the data. This is computed as:
+
+        P(missed het in parent(s) | data) = P(data | missed het in parent(s)) * P(missed het in parent(s))
+
+    P(data | missed het in parent(s)): The probability of observing the data under the assumption of a missed het in parent(s):
+       * Autosomes and PAR regions:
+
+          P(data | missed het in parents) = (P(het in father) * P(hom_ref in mother) + P(hom_ref in father) * P(het in mother)) * P(het in proband) * P(het in one parent)
```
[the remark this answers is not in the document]
KoalaQin: Sorry, that was added unintentionally while I was adjusting the format, it shouldn't be there.

gnomad/sample_qc/relatedness.py
KoalaQin: yeah I knew that, but <0.2 is failing, what. happens with AD = 0.2?

gnomad/sample_qc/relatedness.py
KoalaQin: Okay, I will change it here.

gnomad/sample_qc/relatedness.py
KoalaQin: great catch, it should be hl.sum()

KoalaQin:
I just saw your comment on adding tests, I will work on that now.

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
KoalaQin: This default(0.0) won't work, so I'm changing it back. 
error: 
```
AttributeError: 'StringExpression' object has no attribute 'default'
```

KoalaQin:
Back to you!

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
ch-kr: makes sense. can you note this was taken directly from Kaitlin's original code in the docstring?

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
ch-kr: can you add this note to the docstring?

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
ch-kr: can you transmute that to `is_xx`?

gnomad/sample_qc/relatedness.py
ch-kr: can you document this here that this is a change from Kaitlin's code?

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,230 @@
...
+        :param freq_prior: Frequency prior value.
+        :param min_pop_prior: Minimum population prior.
+        :param expect_error: Whether an error is expected.
+        :param expected: Expected frequency prior.
+        :return: None.
+        """
+        if expect_error:
+            with pytest.raises(
+                hl.utils.java.HailUserError,
+                match="de_novo: expect 0 <= freq_prior_expr <= 1",
```
ch-kr, line 80: naive question, does this `match` work with the error thrown by the actual function since that error message will include the user-provided freq prior? `de_novo: expect 0 <= freq_prior_expr <= 1, found %.3e`

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,230 @@
...
+            ):
+                expr = get_freq_prior(hl.literal(freq_prior), min_pop_prior)
+                hl.eval(expr)  # Hail will throw an error here
+        else:
+            expr = get_freq_prior(hl.literal(freq_prior), min_pop_prior)
+            result = hl.eval(expr)  # Evaluate the expression
```
ch-kr, line 86:
```suggestion
            result = hl.eval(expr)
```

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,230 @@
...
+            assert result == pytest.approx(
+                expected, rel=1e-6
+            )  # Compare floating point values safely
```
ch-kr, line 89:
```suggestion
            )
```

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,230 @@
...
+        :param expect_error: Whether an error is expected.
+        :param expected: Expected frequency prior.
+        :return: None.
+        """
+        if expect_error:
+            with pytest.raises(
+                hl.utils.java.HailUserError,
+                match="de_novo: expect 0 <= freq_prior_expr <= 1",
+            ):
+                expr = get_freq_prior(hl.literal(freq_prior), min_pop_prior)
```
ch-kr, line 82: naive question, is there a reason you convert `freq_prior` to a literal here rather than setting it as a Hail literal above? e.g.,
```
(hl.literal(0.05), 100 / 3e7, False, 0.05),
```

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,230 @@
...
+        locus_expr = self.loci[locus_key]
+        proband_expr = hl.struct(
+            GT=hl.call(*proband_gt) if proband_gt else hl.missing(hl.tcall)
+        )
+        father_expr = hl.struct(
+            GT=hl.call(*father_gt) if father_gt else hl.missing(hl.tcall)
+        )
+        mother_expr = hl.struct(
+            GT=hl.call(*mother_gt) if mother_gt else hl.missing(hl.tcall)
+        )
```
ch-kr, line 190: this is a bit repetitive -- can you condense this?

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @pytest.mark.parametrize(
+        "locus_key, proband_gt, father_gt, mother_gt, is_xx, expected",
+        [
+            ("autosomal", (0, 1), (0, 0), (0, 0), False, True),
+            ("autosomal", (1, 1), (0, 0), (0, 0), False, False),
+            ("x_nonpar", (1, 1), None, (0, 0), False, True),
+            ("x_nonpar", (1, 1), (0, 0), None, False, None),
+            ("y_nonpar", (1, 1), None, None, False, None),
```
ch-kr, line 164: these are testing for `None` as a return value when the function docstring specifies that it only returns boolean expressions. can you add to the docstring that it will return None if there is a missing genotype?

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,230 @@
...
+        expr = calculate_de_novo_post_prob(
+            self.proband_pl,
+            self.father_pl,
+            self.mother_pl,
+            self.locus_expr,
+            self.is_xx_expr,
+            self.freq_prior_expr,
+        )
+        result = hl.eval(expr)
+        assert 0 <= result <= 1  # Posterior probability should be within valid range
```
ch-kr, line 213: why are you checking a range here rather than the exact value?

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,230 @@
...
+            self.father_expr,
+            self.mother_expr,
+            self.is_xx_expr,
+            self.freq_prior_expr,
+        )
+        result = hl.eval(expr)
+
+        assert "p_de_novo" in result
+        assert "confidence" in result
+        assert 0 <= result.p_de_novo <= 1  # Probability must be valid
```
ch-kr, line 230: similar question to above: why not check the actual value produced by your test values?

ch-kr (commented):
thanks for adding tests! I have some more questions and suggestions

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,398 @@ def _get_alt_count(locus, gt, is_female):
...
+       .. math::
+          PL = -10 \\times \\log_{10}{P(\\text{Genotype} | \\text{Data})}
+
+    :param pl_expr: ArrayExpression of PL values.
+    :return: ArrayExpression of the probability of observing each genotype (PP).
+    """
+    return hl.bind(lambda x: x / hl.sum(x), 10 ** (-pl_expr / 10))
+
+
+def calculate_de_novo_post_prob(
```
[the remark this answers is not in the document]
KoalaQin: I want to have this one as an independent function, maybe we'll want to get P for some variants.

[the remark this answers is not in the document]
ch-kr: can you specify when you think this function would be called independently of `get_de_novo_expr`?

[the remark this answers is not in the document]
KoalaQin: When I was testing, I want to see if it's getting the right P values before using it with other metrics for confidence.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,366 @@ def _get_alt_count(locus, gt, is_female):
...
+    :param low_conf_p: P(de novo) threshold for low confidence, default to 0.2.
+
+    :return: A StructExpression with the de novo status and confidence.
+    """
+    # Determine genomic context
+    diploid_expr, hemi_x_expr, hemi_y_expr = get_copy_state_by_sex(
+        locus_expr, is_xx_expr
+    )
+
+    p_de_novo = calculate_de_novo_post_prob(
```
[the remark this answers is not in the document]
KoalaQin: but I need it also for the DP and other metrics.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,366 @@ def _get_alt_count(locus, gt, is_female):
...
+    :param min_proband_ab: Minimum allele balance for the proband, default to 0.2.
+    :param max_parent_ab: Maximum allele balance for parents, default to 0.05.
+    :param min_de_novo_p: Minimum probability for variant to be called de novo, default to 0.05.
+    :param high_conf_dp_ratio: DP ratio threshold of proband DP to combined DP in parents for high confidence, default to 0.2.
+    :param dp_threshold_snp: Minimum depth for high-confidence SNPs, default to 10.
+    :param high_med_conf_ab: AB threshold for high/medium confidence, default to 0.3.
+    :param low_conf_ab: AB threshold for low confidence, default to 0.2.
+    :param high_conf_p: P(de novo) threshold for high confidence, default to 0.99.
+    :param med_conf_p: P(de novo) threshold for medium confidence, default to 0.5.
+    :param low_conf_p: P(de novo) threshold for low confidence, default to 0.2.
```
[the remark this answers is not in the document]
KoalaQin: that's the FAIL p threshold, not for LOW.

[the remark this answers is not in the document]
ch-kr: I didn't read Kaitlin's code, but her documentation says this
```
LOW_SNV:
p_dn > 0.05 and child_AD > 0.2
```
does she use 0.2 in her code?

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,366 @@ def _get_alt_count(locus, gt, is_female):
...
+    Thresholds:
+
+    +----------------+------------+----------------------+------+------+------+------+------+
+    |   Category     | P(de novo) | AB*                  | AD*  | DP*  | DR*  | GQ*  | AC*  |
+    +----------------+------------+----------------------+------+------+------+------+------+
+    | FAIL           | < 0.05     | AB(parents) > 0.05   |  0   |      | <0.1 | <20  |      |
+    |                |            | OR AB(proband) < 0.2 |      |      |      |      |      |
+    | HIGH (Indel)   | > 0.99     | > 0.3                |      |      |      |      |  =1  |
+    | HIGH (SNV) 1   | > 0.99     | > 0.3                |      | >10  |      |      |      |
+    | HIGH (SNV) 2   | > 0.5      | > 0.3                |      |      | >0.2 |      | <10  |
```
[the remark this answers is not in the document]
KoalaQin: Great catch, the table is a bit mixed up, based on https://github.com/ksamocha/de_novo_scripts/blob/bde3e40cba46e02d5b45bc4d780de7761e02aee6/de_novo_finder_3.py#L100
Kaitlin wanted to rescue some MEDIUM SNVs by their DP. Given we're having too many de novos, should we ignore this logic?

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
ch-kr: sorry, I'm not following. I was suggesting that you run the `get_copy_state_by_sex` function only in this function and then take `diploid_expr`, `hemi_x_expr`, `hemi_y_expr` and pass them to `calculate_de_novo_post_prob` rather than calling `get_copy_state_by_sex` a second time in `calculate_de_novo_post_prob` to get those values again.

what do you mean by "I need it also for the DP and other metrics"?

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
ch-kr: that makes sense. it's fine to leave this as a separate function, given its size, but given its only application is for de novo variants, could you rename it `_calculate_de_novo_post_prob` to avoid adding another function name to the namespace of gnomad_methods?

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
ch-kr: I think it's fine to keep Kaitlin's original flags. we'll only release the high confidence de novos in a TSV and full set of de novos in the Hail Table (along with documentation of the Hail Table and how to filter it)

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
KoalaQin: No, she wasn't. Sorry, I got NAs, I think it may due to lack of "=" in LOW.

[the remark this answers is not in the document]
KoalaQin: Let me test by removing "VERY LOW" to fill the logic.

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
KoalaQin: I misunderstood, I thought you wanted me to use `get_copy_state_by_sex` only in `calculate_de_novo_post_prob`, yeah, I can try to pass the return to `calculate_de_novo_post_prob` as arguments.

gnomad/sample_qc/relatedness.py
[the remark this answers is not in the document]
KoalaQin: I found that once we makes it to a helper function, the docstring for the function will not be shown in the API rendering, all the time I spent on formatting it to math blocks is kind of a waste...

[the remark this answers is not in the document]
ch-kr: oh, that is a bummer. maybe leave it as is then. we could consider renaming `get_de_novo_expr` to `default_get_de_novo_expr` or something similar (like we did for [`default_compute_info`](https://github.com/broadinstitute/gnomad_methods/blob/main/gnomad/utils/sparse_mt.py#L593)) to indicate it's the wrapper function that users can call

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,366 @@ def _get_alt_count(locus, gt, is_female):
...
+        .or_missing()
+    )
+
+    # Calculate posterior probability of de novo mutation
+    prob_dn_given_data = prob_data_given_dn * de_novo_prior
+    p_dn = prob_dn_given_data / (prob_dn_given_data + prob_data_missed_het)
+    return p_dn
+
+
+def call_de_novo(
```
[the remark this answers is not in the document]
KoalaQin: It's called in gnomad_qc, for me to filter out the variants that not de novos, I only want to assign a confidence on de novo variants.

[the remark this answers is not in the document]
ch-kr: thanks -- yep, I saw it in the other PR after I finished reviewing this one. I don't think this needs to be a separate function, since this function is specific to calling de novos (it clutters the namespace, and it's confusing from a user perspective to have this function in addition to `get_de_novo_expr`). could you instead put this functionality back into `get_de_novo_expr` and then doing any extra filtering on the resulting table in the pipeline script in gnomAD QC?

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,366 @@ def _get_alt_count(locus, gt, is_female):
...
+    * DP: DP for the proband.
+
+    * DR: Defined as DP(proband) / DP(parent(s)).
+
+    * GQ: GQ for the proband.
+
+    * AC: Supposed to be the sum of alternate alleles in the proband and parents. This has not been implemented yet due to multiple trios in one family, where an allele might be de novo in a parent and transmitted to a child in the dataset.
+
+    :param locus_expr: Variant locus.
+    :param alleles_expr: Variant alleles. It assumes bi-allelic variants, meaning
```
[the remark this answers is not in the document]
KoalaQin: I added an error message, does that work?

KoalaQin:
I rewrote the test simpler, I could parametrize more, but I want you to have a look first.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,415 @@ def _get_alt_count(locus, gt, is_female):
...
+            P(\text{data} \mid \text{missed het in mother}) = (P(\text{het in mother}) + P(\text{hom_alt in mother})) \times P(\text{hom_alt in proband})
+
+      - **Y non-PAR regions**:
+
+        .. math::
+
+            P(\text{data} \mid \text{missed het in father}) = (P(\text{het in father}) + P(\text{hom_alt in father})) \times P(\text{hom_alt in proband})
+
+    - :math:`P(\text{missed het in parent(s)` equals the **probability for at least one heterozygous parent**:
+
```
ch-kr, line 1396: adding this here as well since you mentioned you are going to fix the formatting change, and I don't want the rewording to get lost: 
image

```
Prior that at least one parent is heterozygous. Depends on alternate allele frequency
```

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,415 @@ def _get_alt_count(locus, gt, is_female):
...
+                )
+            ),
+            "HIGH",
+        )
+        .when((p_de_novo > med_conf_p) & (proband_ab > high_med_conf_ab), "MEDIUM")
+        .when((p_de_novo >= min_de_novo_p) & (proband_ab >= low_conf_ab), "LOW")
+        .or_missing()
+    )
+
+    parent_sum_ad_0_expr = (
```
KoalaQin, line 1674: Now I'm looking at the code and data, isn't hl.sum(AD)=DP? Especially it's split. Can we just use parent DP here?

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class TestDeNovoMutation:
+    """Test suite for de novo mutation functions."""
+
+    @pytest.mark.parametrize(
+        "proband_pl, father_pl, mother_pl, diploid, hemi_x, hemi_y, freq_prior, min_pop_prior, expected_p_dn",
+        [
+            (
+                [73, 0, 161],
```
ch-kr, line 19: rather than having so many examples, could you have just a few with the following conditions:
- two examples where `calculate_de_novo_post_prob` should return the expected `P(dn)` values
- one example where `calculate_de_novo_post_prob` will throw an error because `freq_prior_expr` is out of the expected range?

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,168 @@
...
+            hl.literal(hemi_x),
+            hl.literal(hemi_y),
+            hl.literal(freq_prior),
+            min_pop_prior,
+        )
+
+        # Assert with floating-point tolerance
+        assert round(hl.eval(p_dn_expr), 3) == expected_p_dn
+
+    def test_default_get_de_novo_expr_fail_conditions(self):
```
ch-kr, line 125:
```suggestion
    def test_default_get_de_novo_expr(self):
```
since this is a test function, I don't think it needs to include "fail_conditions" explicitly in the name. also, similar to the comment above, I think this should test a few scenarios:
- a case like the one you've included, where multiple fail conditions apply
- a case where only one fail condition is true
- a case where the variant has `False` for `is_de_novo`
- a passing case

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,168 @@
...
+        """Test default_get_de_novo_expr with a failing case where multiple fail conditions apply."""
+        # Define locus and alleles (Autosomal)
+        locus = hl.locus("1", 10000)
```
ch-kr, line 128: this should specify the reference genome build (see example [here](https://github.com/broadinstitute/gnomad_methods/blob/main/tests/assessment/test_validity_checks.py#L266))

ch-kr (changes requested):
thanks for doing the suggested restructuring! the current structure and formatting looks nice. I have some more minor suggestions and some suggestions for the tests

gnomad/sample_qc/relatedness.py
ch-kr: this is a good question. I thought I remembered hearing that sum(AD) isn't exactly equal to DP and found this note:
> The AD calculation as performed by HaplotypeCaller may not yield exact results because only reads that statistically favor one allele over the other are counted. Due to this fact, the sum of AD may be different than the individual sample depth, especially when there are many non-informative reads.

https://gatk.broadinstitute.org/hc/en-us/articles/360037592411-DepthPerAlleleBySample

could you test to see if `sum(AD) == DP` here?

KoalaQin: I think the problem is that we don't have AD for hom_ref stored, and we're approximate it back by: 
```
AD=hl.or_else(mt.AD, [mt.DP, 0])
```
The sum(AD) would be DP for those hom ref parent(s) for sure. Do you know where to find the trace (or documentation) that we intentionally removed AD and PL?

ch-kr: oh that makes sense. given this function depends on the presence of AD/PL, let's stick with AD anyway, even if DP would be equivalent for our dataset. other people using this function might still have AD and PL in their datasets.

finding the documentation or trail on this topic might take a while -- my instincts are checking slack (both Broad and ATGU), project meeting notes, and joint calling meeting notes

KoalaQin:
I'll leave you to resolve the conversations in case I didn't address them as you suggested.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,423 @@ def _get_alt_count(locus, gt, is_female):
...
+        :param pl_expr: ArrayExpression of PL values.
+        :return: ArrayExpression of the probability of observing each genotype (PP).
+        """
+        return hl.bind(lambda x: x / hl.sum(x), 10 ** (-pl_expr / 10))
+
+    # Adjust frequency prior
+    freq_prior_expr = _get_freq_prior(freq_prior_expr, min_pop_prior)
+    prior_one_parent_het = 1 - (1 - freq_prior_expr) ** 4
+
+    # Convert PL to probabilities
```
KoalaQin, line 1465: I changed this back because your suggestion will need more changes below.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,381 @@ def _get_alt_count(locus, gt, is_female):
...
+
+    Returns
+    -------
+    hl.expr.StructExpression
+        A struct containing:
+        - `confidence`: Confidence level ("HIGH", "MEDIUM", "LOW", or missing).
+        - `fail`: Boolean indicating if the variant fails any checks.
+        - `fail_reason`: Set of strings with reasons for failure.
+    """
+    p_de_novo = calculate_de_novo_post_prob(
```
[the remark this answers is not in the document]
KoalaQin: I just saw this one and added now.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,415 @@ def _get_alt_count(locus, gt, is_female):
...
+    |   Category     | P(de novo) | AB                    | AD   | DP  | DR   | GQ  |
+    +================+============+=======================+======+=====+======+=====+
+    | FAIL           | < 0.05     | AB(parents) > 0.05 OR |  0   |     | <0.1 | <20 |
+    |                |            | AB(proband) < 0.2     |      |     |      |     |
+    +----------------+------------+-----------------------+------+-----+------+-----+
+    | HIGH (Indel)   | > 0.99     | > 0.3                 |      |     |      |     |
+    +----------------+------------+-----------------------+------+-----+------+-----+
+    | HIGH (SNV) 1   | > 0.99     | > 0.3                 |      |     | >0.2 |     |
+    +----------------+------------+-----------------------+------+-----+------+-----+
+    | HIGH (SNV) 2   | > 0.5      | > 0.3                 |      | >10 |      |     |
```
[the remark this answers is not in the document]
KoalaQin: Can you still see this one? I can't see it under "Files changed" and it's folded as Outdated here...

gnomad/sample_qc/relatedness.py
```diff
@@ -1524,30 +1528,30 @@ def default_get_de_novo_expr(
...
+    +----------------+--------------+-----------------------+------+-----+------+-----+
+    | HIGH (Indel)   | > 0.99       | > 0.3                 |      |     |      |     |
+    +----------------+--------------+-----------------------+------+-----+------+-----+
+    | HIGH (SNV) 1   | > 0.99       | > 0.3                 |      |     | >0.2 |     |
+    +----------------+--------------+-----------------------+------+-----+------+-----+
+    | HIGH (SNV) 2   | > 0.5        | > 0.3                 |      | >10 |      |     |
+    +----------------+--------------+-----------------------+------+-----+------+-----+
+    | MEDIUM         | > 0.5        | > 0.3                 |      |     |      |     |
+    +----------------+--------------+-----------------------+------+-----+------+-----+
+    | LOW            | >= 0.05      | >= 0.2                |      |     |      |     |
```
ch-kr, line 1549: just to revive the comment thread that became outdated -- I commented here that Kaitlin didn't use `>=`, and you were aware of this.

to respond to your question (`but where do those =0.05 and =0.2 belong?`): 
it doesn't seem like it would make a big difference to mark these variants as failing or to allow them to pass with low confidence. the more important piece is that this is well documented -- please add a note to the docstring that this is new behavior (this is a change from Kaitlin's original code)

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,415 @@ def _get_alt_count(locus, gt, is_female):
...
+        "parent_sum_ad_0": parent_sum_ad_0_expr,
+        "max_parent_ab": fail_max_parent_ab_expr,
+        "min_proband_ab": proband_ab < min_proband_ab,
+        "min_proband_gq": proband_expr.GQ < min_gq,
+        "min_de_novo_p": p_de_novo <= min_de_novo_p,
+    }
+
+    fail = hl.any(list(fail_checks_expr.values()))
+    result_expr = hl.struct(
+        is_de_novo=is_de_novo,
```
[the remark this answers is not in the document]
ch-kr: the latter -- the name `is_de_novo`. though I actually think, upon re-review, that this naming is fine in the context of the overall struct, so you can resolve this comment thread

gnomad/sample_qc/relatedness.py
```diff
@@ -1708,7 +1713,10 @@ def default_get_de_novo_expr(
         is_de_novo=is_de_novo,
         p_de_novo=hl.if_else(~is_de_novo | fail, hl.missing(hl.tfloat64), p_de_novo),
         confidence=hl.if_else(~is_de_novo | fail, hl.missing(hl.tstr), confidence_expr),
-        fail_reason=add_filters_expr(filters=fail_checks_expr),
+        fail_reason=hl.if_else(
```
ch-kr, line 1730: could you use `or_missing` here instead of `if_else` + `hl.empty_set()`?

tests/sample_qc/test_de_novo.py
```diff
@@ -13,8 +13,9 @@ class TestDeNovoMutation:
     """Test suite for de novo mutation functions."""
 
     @pytest.mark.parametrize(
-        "proband_pl, father_pl, mother_pl, diploid, hemi_x, hemi_y, freq_prior, min_pop_prior, expected_p_dn",
+        "proband_pl, father_pl, mother_pl, diploid, hemi_x, hemi_y, freq_prior, min_pop_prior, expected",
         [
+            # ✅ Valid test cases (should return numeric values)
```
ch-kr, line 18:
```suggestion
            # Valid test cases (should return expected numeric values)
```
I enjoy this emoji but don't think it's necessary here

tests/sample_qc/test_de_novo.py
```diff
@@ -37,60 +38,17 @@ class TestDeNovoMutation:
                 100 / 3e7,
                 0.198,
             ),
+            # ❌ Invalid `freq_prior` case (should raise `HailUserError`)
```
ch-kr, line 41:
```suggestion
            # Invalid `freq_prior` case (should raise `HailUserError`)
```

tests/sample_qc/test_de_novo.py
ch-kr: I chatted with Mike, and he pointed me to Kristen's recently written tests as a good example of what our tests should do: https://github.com/broadinstitute/gnomad_methods/blob/main/tests/assessment/test_validity_checks.py. 

with my comment above: we should test for valid uses of the function (making sure valid inputs return expected outputs) and incorrect uses of the function (checking that the function returns errors as expected). you've added these suggestions, but it would also be a good idea to add tests for any edge cases you can think of (if possible).

Mike also mentioned that pytest has a built in `caplog` functionality to capture logger information, which made me realize these functions don't have any loggers. I'm not sure they're necessary, but I wanted to note it since I thought of it

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1300,415 @@ def _get_alt_count(locus, gt, is_female):
...
+        .when(hemi_y_expr, father_expr.DP)
+        .or_missing()
+    )
+    dp_ratio = proband_expr.DP / parent_dp
+
+    # Calculate proband AB and assign variant type
+    proband_ab = proband_expr.AD[1] / hl.sum(proband_expr.AD)
+    is_snp = hl.is_snp(alleles_expr[0], alleles_expr[1])
+
+    is_de_novo = (
```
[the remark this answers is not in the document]
ch-kr: ah yes, sorry, that wasn't clear. this suggestion was relevant when you wanted to filter de novos from the input dataset upfront. in that scheme, you would pass an argument to this function (e.g., `filter_to_candidate_dnms` or something similar), and the default for that argument would be `True`, which means the function would first filter the input dataset to candidate de novo variants before calculating p-values and confidence levels.

however, in your current function design, where you pass in and return only expressions, I don't think this behavior would make sense; I more wanted to suggest this as an alternative option to your previous structure  (filtering the HT to variants with GTs that indicate possibility of being a de novo variant).

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,208 @@
...
+        mother_pl,
+        diploid,
+        hemi_x,
+        hemi_y,
+        freq_prior,
+        min_pop_prior,
+        expected,
+    ):
+        """Test `calculate_de_novo_post_prob` function."""
+        # Case where we expect an error (freq_prior is out of range)
```
ch-kr, line 68:
```suggestion
        # Case where we expect an error (`freq_prior` is out of range)
```

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,208 @@
...
+            ):
+                hl.eval(
+                    calculate_de_novo_post_prob(
+                        hl.literal(proband_pl),
+                        hl.literal(father_pl),
+                        hl.literal(mother_pl),
+                        hl.literal(diploid),
+                        hl.literal(hemi_x),
+                        hl.literal(hemi_y),
+                        hl.literal(freq_prior),  # Invalid frequency prior
```
ch-kr, line 82:
```suggestion
                        hl.literal(freq_prior),
```

ch-kr (changes requested):
a few more minor suggestions

tests/sample_qc/test_de_novo.py
KoalaQin: Thank you for sending the example, the `pytest.fixture` seems to be very useful. I'm still trying to think of some edge cases to test, for now I want to send back my modifications.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,440 @@ def _get_alt_count(locus, gt, is_female):
...
+        "min_proband_ab": proband_ab < min_proband_ab,
+        "min_proband_gq": proband_expr.GQ < min_gq,
+        "min_de_novo_p": p_de_novo <= min_de_novo_p,
+    }
+
+    fail = hl.any(list(fail_checks_expr.values()))
+    result_expr = hl.struct(
+        is_de_novo=is_de_novo,
+        p_de_novo=hl.if_else(~is_de_novo | fail, hl.missing(hl.tfloat64), p_de_novo),
+        confidence=hl.if_else(~is_de_novo | fail, hl.missing(hl.tstr), confidence_expr),
```
ch-kr, line 1734: these `if_else` could also be replaced with `or_missing`, right?

gnomad/utils/annotations.py
```diff
@@ -2780,3 +2780,28 @@ def _create_group_dicts(
...
+) -> Tuple[
+    hl.expr.BooleanExpression, hl.expr.BooleanExpression, hl.expr.BooleanExpression
+]:
+    """
+    Determine the copy state of a variant by its locus and the sex karotype of a sample.
+
+    :param locus_expr: LocusExpression of the variant.
+    :param is_xx_expr: BooleanExpression indicating whether the sample has an XX sex
+       karyotype.
+    :return: A tuple of BooleanExpressions:
```
ch-kr, line 2797:
```suggestion
    :return: Tuple of BooleanExpressions:
```

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,244 @@
...
+                False,
+                True,
+                False,
+                1.2,
+                100 / 3e7,
+                None,
+            ),
+        ],
+    )
+    def test_calculate_de_novo_post_prob(
```
ch-kr, line 55: looking at this again, I wonder if this function or the other function should test for missingness in the PL expressions. what do you think?

tests/sample_qc/test_de_novo.py
```diff
@@ -0,0 +1,244 @@
...
+                hl.literal(["C", "G"]),
+                hl.struct(GT=hl.call(0, 1), AD=[5, 30], DP=35, GQ=99, PL=[99, 0, 1]),
+                hl.struct(GT=hl.call(0, 0), AD=[20, 0], DP=20, GQ=60, PL=[0, 60, 120]),
+                hl.struct(GT=hl.call(0, 0), AD=[25, 0], DP=25, GQ=80, PL=[0, 80, 150]),
+                hl.literal(True),
+                hl.literal(1e-5),
+                False,
+                hl.struct(
+                    is_de_novo=True,
+                    p_de_novo=0.999,  # High confidence P(de novo)
```
ch-kr, line 175:
```suggestion
                    p_de_novo=0.999,
```

ch-kr: it might be worth checking MEDIUM and LOW as well, and also including an indel test case. what do you think about including all three confidence levels?

ch-kr (changes requested):
more nitpicks. can you also add a test for the copy state function?

KoalaQin (commented):
I tried to use `@pytest.fixture` to get a table, but it's not very convenient for the error cases, so I changed it back by adding more cases.

KoalaQin:
@ch-kr I gathered all the thresholds of different implementations in a [sheet](https://docs.google.com/spreadsheets/d/1E8xSEvfsK-NjmO5kZ7DHhyvkTo7DDP9-C9NcPoA7lHk/edit?usp=sharing) here, so you can see clearly what was changed.

gnomad/utils/annotations.py
```diff
@@ -2794,7 +2794,7 @@ def get_copy_state_by_sex(
     :param locus_expr: LocusExpression of the variant.
     :param is_xx_expr: BooleanExpression indicating whether the sample has an XX sex
        karyotype.
-    :return: A tuple of BooleanExpressions:
+    :return: Tuple of BooleanExpressions:
```
ch-kr, line 2797: it just occurred to me that this is assuming the sample only contains XX and XY karyotypes (i.e., no samples have ambiguous sex or aneuploidies). can you add that into the docstring?

tests/sample_qc/test_de_novo.py
```diff
@@ -7,11 +7,60 @@
...
+            (
+                hl.locus("chrY", 10000000, reference_genome="GRCh38"),
+                False,
+                False,
+                False,
+                True,
+            ),
+        ],
+    )
+    def test_get_copy_state_by_sex(
```
ch-kr, line 49: this test should be in another file because the function is in a separate file, right? or is this (keeping related functions together) the convention

tests/sample_qc/test_de_novo.py
ch-kr: yeah, my thinking was that it might be a good idea to make sure that missing PLs returns a missing value for p_d_novo as expected. fine with me if you think this is covered already

tests/sample_qc/test_de_novo.py
```diff
@@ -102,55 +151,142 @@ def test_calculate_de_novo_post_prob(
...
-                    ),
+                    p_de_novo=0.999,
+                    confidence="HIGH",
+                    fail_reason=hl.missing(hl.tset(hl.tstr)),
+                ),
+            ),
+            # 2. Autosomal locus with MEDIUM confidence
+            (
+                hl.locus("chr1", 11000, reference_genome="GRCh38"),
+                hl.literal(["CT", "C"]),
```
ch-kr, line 174: this is the only test that has an indel -- could you switch at least one more test to be an insertion or deletion (just to balance the tests between SNVs and indels a bit more)

ch-kr (changes requested):
so close -- just two more minor suggestions!

KoalaQin (commented):
I didn't change the hemi calls to have both parent hom ref, because I found [her code](https://github.com/ksamocha/de_novo_scripts/blob/bde3e40cba46e02d5b45bc4d780de7761e02aee6/de_novo_finder_3.py#L938) was checking one parent only, I think our "adj" will do that for us.

gnomad/sample_qc/relatedness.py
```diff
@@ -1298,3 +1301,442 @@ def _get_alt_count(locus, gt, is_female):
...
+
+            P(\text{data} \mid \text{missed het in parents}) = ( P(\text{het in father}) \cdot P(\text{hom_ref in mother}) + P(\text{hom_ref in father}) \cdot P(\text{het in mother})) \cdot P(\text{het in proband})
+
+      - **X non-PAR regions**:
+
+        .. math::
+
+            P(\text{data} \mid \text{missed het in mother}) = (P(\text{het in mother}) + P(\text{hom_alt in mother})) \cdot P(\text{hom_alt in proband})
+
+      - **Y non-PAR regions**:
```
ch-kr, line 1393: these are also for XY samples only

gnomad/sample_qc/relatedness.py
```diff
@@ -1530,24 +1538,24 @@ def default_get_de_novo_expr(
...
-    +----------------+--------------+-----------------------+------+-----+------+-----+
+    Confidence thresholds (from Kaitlin Samocha's `de novo caller <https://github.com/ksamocha/de_novo_scripts>`_):
+
+    +----------------+--------------+-----------------------+------+------+-------+------+
+    |   Category     | P(*de novo*) | AB                    | AD   | DP   | DR    | GQ   |
+    +================+==============+=======================+======+======+=======+======+
+    | FAIL           | < 0.05       | AB(parents) > 0.05 OR |  0   |      | < 0.1 | < 20 |
+    |                |              | AB(proband) < 0.2     |      |      |       |      |
+    +----------------+--------------+-----------------------+------+------+-------+------+
+    | HIGH (Indel)   | > 0.99       | > 0.3                 |      |      | > 0.2 |      |
```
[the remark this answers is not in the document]
ch-kr: bumping this comment
> However, if both the Hail code and Julia's code don't include DR here, then I vote we also remove it for consistency

ch-kr (changes requested):
bumping one older question, then good to go

KoalaQin:
I asked Kaitlin about this, she [said](https://the-tgg.slack.com/archives/C08C8DU0UR2/p1739224747455309).

KoalaQin:
I cited the points and added a summary for this PR, could you have a look at that?

gnomad/sample_qc/relatedness.py
ch-kr: can you add a note here reflecting this for consistency with the documentation above?

ch-kr (approved):
following up on minor docstring suggestion, then LGTM!

KoalaQin:
> following up on minor docstring suggestion, then LGTM!

ah, sorry, forgot to push my commit.

PR #761, opened 2025-01-31
Add public resource for v2 MNV Hail Tables

ch-kr:

### DIFF
--- a/gnomad/resources/grch37/gnomad.py
+++ b/gnomad/resources/grch37/gnomad.py
@@ -156,6 +156,22 @@ def _public_browser_gene_ht_path() -> str:
     return "gs://gnomad-public-requester-pays/resources/grch37/browser/gnomad.genes.GRCh37.GENCODEv19.pext.ht"
 
 
+def _public_mnv_path(distance: str) -> str:
+    """
+    Get public multinucleotide variant (MNV) data.
+
+    :param distance: Distance between two SNVs in MNV.
+    :return: Path to MNV data.
+    :raises DataException: If the provided distance is invalid (outside 1-10).
+    """
+    distances = set(map(str, range(1, 10)))
+    if distance not in distances:
+        raise DataException(
+            f"Invalid distance: '{distance}'. Valid options are {distances}."
+        )
+    return "gs://gnomad-public-requester-pays/release/2.1/mnv/gnomad_mnv_genome_d{distance}.ht"
+
+
 def public_release(data_type: str) -> VersionedTableResource:
     """
     Retrieve publicly released versioned table resource.
@@ -290,6 +306,16 @@ def constraint() -> GnomadPublicTableResource:
     :return: Gene constraint Table.
     """
     return GnomadPublicTableResource(path=_public_constraint_ht_path())
+
+
+def mnv(distance: str) -> GnomadPublicTableResource:
+    """
+    Retrieve multinucleotide variant table.
+
+    :param distance: Distance between two SNVs in MNV.
+    :return: MNV Table.
+    """
+    return GnomadPublicTableResource(path=_public_mnv_path())
 
 
 def browser_gene() -> GnomadPublicTableResource:


gnomad/resources/grch37/gnomad.py
```diff
@@ -156,6 +156,22 @@ def _public_browser_gene_ht_path() -> str:
     return "gs://gnomad-public-requester-pays/resources/grch37/browser/gnomad.genes.GRCh37.GENCODEv19.pext.ht"
 
 
+def _public_mnv_path(distance: str = "1") -> str:
+    """
+    Get public multinucleotide variant (MNV) data.
```
klaricch, line 161:
```suggestion
    Get path to public multinucleotide variant (MNV) data.
```

gnomad/resources/grch37/gnomad.py
```diff
@@ -156,6 +156,22 @@ def _public_browser_gene_ht_path() -> str:
...
+
+    :param distance: Distance between two SNVs in MNV. Default is '1'.
+    :return: Path to MNV data.
+    :raises DataException: If the provided distance is invalid (outside 1-10).
+    """
+    distances = set(map(str, range(1, 10)))
```
klaricch, line 167:
```suggestion
    distances = set(map(str, range(1, 11)))
```

gnomad/resources/grch37/gnomad.py
```diff
@@ -156,6 +156,22 @@ def _public_browser_gene_ht_path() -> str:
     return "gs://gnomad-public-requester-pays/resources/grch37/browser/gnomad.genes.GRCh37.GENCODEv19.pext.ht"
 
 
+def _public_mnv_path(distance: str = "1") -> str:
+    """
+    Get public multinucleotide variant (MNV) data.
+
+    :param distance: Distance between two SNVs in MNV. Default is '1'.
+    :return: Path to MNV data.
+    :raises DataException: If the provided distance is invalid (outside 1-10).
```
klaricch, line 165: is this something we're adding now or can the "raises" part here be dropped?

gnomad/resources/grch37/gnomad.py
```diff
@@ -156,6 +156,22 @@ def _public_browser_gene_ht_path() -> str:
...
+    """
+    distances = set(map(str, range(1, 10)))
+    if distance not in distances:
+        raise DataException(
+            f"Invalid distance: '{distance}'. Valid options are {distances}."
+        )
+    return "gs://gnomad-public-requester-pays/release/2.1/mnv/gnomad_mnv_genome_d{distance}.ht"
```
klaricch, line 172:
```suggestion
    return f"gs://gnomad-public-requester-pays/release/2.1/mnv/gnomad_mnv_genome_d{distance}.ht"
```

klaricch, line 172: are there supposed to be files here? it looks like there should be another subdir in the path, although there's both "genome" and "genomes"?

gnomad/resources/grch37/gnomad.py
```diff
@@ -156,6 +156,22 @@ def _public_browser_gene_ht_path() -> str:
     return "gs://gnomad-public-requester-pays/resources/grch37/browser/gnomad.genes.GRCh37.GENCODEv19.pext.ht"
 
 
+def _public_mnv_path(distance: str = "1") -> str:
```
klaricch, line 159: here and in below function i think it would make sense to make distance an int rather than a string

klaricch (changes requested):

gnomad/resources/grch37/gnomad.py
ch-kr: ah I copied the structure from pext above and am happy to remove (maybe cc @mike-w-wilson for opinions?)

gnomad/resources/grch37/gnomad.py
ch-kr: ah oops you're right --  I missed "genome" in the path. not sure why we have "genome" and "genomes", but the downloads page points to "genome" so will put that one here

gnomad/resources/grch37/gnomad.py
```diff
@@ -156,6 +156,22 @@ def _public_browser_gene_ht_path() -> str:
...
+def _public_mnv_path(distance: int = 1) -> str:
+    """
+    Get path to public multinucleotide variant (MNV) data.
+
+    :param distance: Distance between two SNVs in MNV. Default is 1.
+    :return: Path to MNV data.
+    :raises DataException: If the provided distance is invalid (outside 1-10).
+    """
+    distances = set(map(str, range(1, 11)))
+    if str(distance) not in distances:
```
klaricch, line 168:
```suggestion
    distances = set(range(1, 11))
    if distance not in distances:
```

klaricch (changes requested):

klaricch (approved):
lgtm!

PR #763, opened 2025-02-10
Make folders in RE for pre-commit config

KoalaQin:
Before the fix, black, autoped8 and isort would be skipped in pre-commit hooks, says: 
```
check yaml...........................................(no files to check)Skipped
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
black................................................(no files to check)Skipped
autopep8.............................................(no files to check)Skipped
pydocstyle...............................................................Passed
isort................................................(no files to check)Skipped
```
Kristen suggested it might need to be regular expression format for the two folders, and I tested, it's working now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,13 +11,13 @@ repos:
     hooks:
       - id: black
         language_version: python3
-        files: gnomad tests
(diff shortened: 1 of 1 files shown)


klaricch (approved):
lgtm!

PR #765, opened 2025-02-13
Edits to support reformatted validity check output

klaricch:


gnomad/assessment/validity_checks.py
```diff
@@ -46,30 +80,127 @@ def generic_field_check(
...
+            function_name=function_name,
+        )
+
+
+def generate_field_check_expr(
+    left_expr: Union[hl.expr.NumericExpression, hl.expr.StringExpression],
+    right_expr: Union[hl.expr.NumericExpression, hl.expr.StringExpression],
+    operator: str,
+) -> hl.expr.BooleanExpression:
+    """Generate a Hail expression to check field comparisons while handling missing values.
```
mike-w-wilson, line 174:
```suggestion
    """
    Generate a Hail expression to check field comparisons while handling missing values.
```

gnomad/assessment/validity_checks.py
```diff
@@ -46,30 +80,127 @@ def generic_field_check(
...
+
+    If both fields are missing, the retured expression will be False. If only one field is missing, the expression will be True. If both fields are defined and not equal, the expression will be True.
```
mike-w-wilson, line 176:
```suggestion
    If both fields are missing, the retured expression will be False. If only one field 
    is missing, the expression will be True. If both fields are defined and not equal,
    the expression will be True.
```

gnomad/assessment/validity_checks.py
```diff
@@ -501,6 +601,7 @@ def sum_group_callstats(
     :return: None
     """
     # TODO: Add support for subpop sums
+    # set_first_function()
```
mike-w-wilson, line 604:
```suggestion
```

mike-w-wilson (approved):
Just some slight comment changes but this LGTM!

PR #781, opened 2025-05-20
Add May 2025 clinvar resource

jkgoodrich:


gnomad/resources/grch38/reference_data.py
```diff
@@ -46,7 +50,12 @@ def _import_clinvar(**kwargs) -> hl.Table:
     clinvar = clinvar.filter(
         hl.len(clinvar.alleles) > 1
     )  # Get around problematic single entry in alleles array in the clinvar vcf
+    _curr_source = gnomad_public_resource_configuration.source
+    gnomad_public_resource_configuration.source = (
+        GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS
+    )
     clinvar = vep_or_lookup_vep(clinvar, reference="GRCh38")
+    gnomad_public_resource_configuration.source = _curr_source
```
ch-kr, line 58: two basic questions:
- Do you need to set the source because you're running this locally? I thought the default source would be found automatically (if running on a cluster) based on the docstring for `get_default_public_resource_source`
- what's the reason for creating `_curr_source` rather than directly setting `gnomad_public_resource_configuration.source` to `GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS`?

ch-kr (commented):
a couple questions

gnomad/resources/grch38/reference_data.py
jkgoodrich: https://github.com/broadinstitute/gnomad_methods/blob/6d0b27b0432d8c999b829c714ebac747c6b60876/gnomad/resources/import_resources.py#L91

The import script sets it to GNOMAD since that is where we need to write the HT to. But the VEP function needs to use the public VEP context HT.

ch-kr: ah got it, thanks!

ch-kr (approved):
LGTM!

PR #786, opened 2025-06-26
Update sex terms repo-wide

ch-kr:
PR updates terms to use karyotype terms (XX/xx, XY/xy) instead of gender terms. PR also updates a few instances of "population"/"pop" to "genetic ancestry"/"gen_anc"

### DIFF
--- a/gnomad/sample_qc/sex.py
+++ b/gnomad/sample_qc/sex.py
@@ -14,7 +14,7 @@ logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
-SEXES = {"Male": "Male", "Female": "Female"}
+SEXES = {"XY": "XY", "XX": "XX"}
 
 
 def adjusted_sex_ploidy_expr(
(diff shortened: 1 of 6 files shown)


gnomad/sample_qc/sex.py
```diff
@@ -61,20 +61,20 @@ def adjusted_sex_ploidy_expr(
 def adjust_sex_ploidy(
     mt: hl.MatrixTable,
     sex_expr: hl.expr.StringExpression,
-    male_str: str = "male",
-    female_str: str = "female",
+    xy_str: str = "xy",
```
jkgoodrich: I think we have been going with capitals

gnomad/utils/vcf.py
```diff
@@ -1482,29 +1482,29 @@ def make_hist_dict(
...
     """
-    Set AC, AN, and nhomalt chrY variant annotations for females to NA (instead of 0).
+    Set AC, AN, and nhomalt chrY variant annotations for XX samples to NA (instead of 0).
 
-    :param t: Table/MatrixTable containing female variant annotations.
+    :param t: Table/MatrixTable containing XX variant annotations.
     :return: Dictionary with reset annotations
     """
     metrics = list(t.row.info)
-    female_metrics = [x for x in metrics if "_female" in x or "_XX" in x]
```
jkgoodrich, line 1495: This one already handles both, but I'm fine if we want to only support XX now

jkgoodrich:
Also, it might be good to add which functions are specifically impacted and/or renamed in the PR name (I know it will be a little long) so it shows up in the change log

gnomad/sample_qc/sex.py
ch-kr: ah oops, this may have been from poor prompting, will fix

gnomad/utils/vcf.py
ch-kr: this might have been a tab completion that I missed. I'll add the old code back just in case

jkgoodrich (approved):
lgtm, thanks for doing!

PR #789, opened 2025-07-03
Add function for exploding intervals to loci

Ruchit10:
Feature Addition:

Added a function in `intervals.py` which explodes hail intervals into per base loci, keeping same reference genome as original interval. 

Optionally it can return the original interval that the locus belongs to or drop it from the exploded table.

### DIFF
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -115,7 +115,7 @@ def pad_intervals(
 
 
 def explode_intervals_to_loci(
-    obj: Union[hl.Table, hl.MatrixTable],
+    ht: hl.Table,
     interval_field: str = "interval",
     keep_intervals: bool = False,
 ) -> Union[hl.Table, hl.MatrixTable]:
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -134,7 +134,9 @@ def explode_intervals_to_loci(
     includes_start = interval.includes_start.take(1)[0]
     includes_end = interval.includes_end.take(1)[0]
 
-    interval_start = interval.start.position if includes_start else interval.start.position + 1
+    interval_start = (
+        interval.start.position if includes_start else interval.start.position + 1
+    )
     interval_end = interval.end.position + 1 if includes_end else interval.end.position
 
     ht = ht.annotate(pos=hl.range(interval_start, interval_end)).explode("pos")
@@ -142,7 +144,7 @@ def explode_intervals_to_loci(
         locus=hl.locus(
             ht[interval_field].start.contig,
             ht.pos,
-            reference_genome=str(interval.start.take(1)[0].reference_genome)
+            reference_genome=str(interval.start.take(1)[0].reference_genome),
         )
     ).key_by("locus")
 
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -125,7 +125,7 @@ def explode_intervals_to_loci(
     :param obj: Hail Table with intervals to be exploded.
     :param interval_field: Name of the interval field. Default is 'interval'.
     :param keep_intervals: If True, keep the original intervals as a column in output.
-    :return: Hail Table or MatrixTable with interval exploded to loci.
+    :return: Hail Table with intervals exploded to loci.
     """
     is_matrix = isinstance(obj, hl.MatrixTable)
     ht = obj.rows() if is_matrix else obj
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -124,7 +124,7 @@ def explode_intervals_to_loci(
 
     :param obj: Hail Table with intervals to be exploded.
     :param interval_field: Name of the interval field. Default is 'interval'.
-    :param keep_intervals: If True, keep the original interval as a column in output.
+    :param keep_intervals: If True, keep the original intervals as a column in output.
     :return: Hail Table or MatrixTable with interval exploded to loci.
     """
     is_matrix = isinstance(obj, hl.MatrixTable)
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -3,6 +3,7 @@
 from typing import List, Union
 
 import hail as hl
+
 from gnomad.utils.reference_genome import get_reference_genome
 
 
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -129,7 +129,11 @@ def explode_intervals_to_loci(
     :param keep_intervals: If True, keep the original intervals as a column in output. Only applies if input is a Hail Table. Default is False.
     :return: If input is a Hail Table, returns exploded Table keyed by locus. If input is an IntervalExpression, returns position array expression.
     """
-    if isinstance(obj, hl.expr.IntervalExpression):
+    assert isintance(intervals, hl.Table) or isinstance(intervals, hl.expr.IntervalExpression), "Input must be a Table or IntervalExpression!"
+
+    if isinstance(intervals, hl.Table) and (not interval_field or keep_intervals is None):
+        raise ValueError("`interval_field` and `keep_intervals` must be defined if input is a Table!")
+    assert interval_field in intervals.row, "`interval_field` must be an annotation present on input Table!"
         interval = obj
         interval_start_expr = hl.if_else(
             interval.includes_start,
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -115,25 +115,48 @@ def pad_intervals(
 
 
 def explode_intervals_to_loci(
-    ht: hl.Table,
+    obj: Union[hl.Table, hl.MatrixTable],
+    interval_field: str = "interval",
     keep_intervals: bool = False,
-) -> hl.Table:
+) -> Union[hl.Table, hl.MatrixTable]:
     """
     Expand intervals to loci.
 
-    :param ht: Hail Table with an interval field.
+    :param obj: Hail Table or MatrixTable with an interval field.
+    :param interval_field: Name of the interval field. Default is 'interval'.
     :param keep_intervals: If True, keep the original interval as a column in output.
-        Default is False.
-    :return: Hail Table keyed by loci and intervals as optional field.
+    :return: Hail Table or MatrixTable with interval exploded to loci.
     """
-    ht = ht.annotate(
-        pos=hl.range(ht.interval.start.position, ht.interval.end.position + 1),
-    ).explode("pos")
+    is_matrix = isinstance(obj, hl.MatrixTable)
+    ht = obj.rows() if is_matrix else obj
+
+    interval = ht[interval_field]
+    includes_start = interval.includes_start.take(1)[0]
+    includes_end = interval.includes_end.take(1)[0]
+
+    interval_start = interval.start.position if includes_start else interval.start.position + 1
+    interval_end = interval.end.position + 1 if includes_end else interval.end.position
+
+    ht = ht.annotate(pos=hl.range(interval_start, interval_end)).explode("pos")
     ht = ht.annotate(
         locus=hl.locus(
-            ht.interval.start.contig,
+            ht[interval_field].start.contig,
             ht.pos,
-            reference_genome=ht.interval.start.reference_genome,
-        ),
+            reference_genome=str(interval.start.take(1)[0].reference_genome)
+        )
     ).key_by("locus")
-    return ht.drop("interval", "pos") if not keep_intervals else ht.drop("pos")
+
+    fields_to_drop = ["pos"]
+    if not keep_intervals:
+        fields_to_drop.append(interval_field)
+
+    ht = ht.drop(*fields_to_drop)
+
+    if is_matrix:
+        mt = obj
+        ht = ht.select_globals()
+        mt = mt.annotate_rows(**ht[mt.row_key])
+        mt = mt.filter_rows(hl.is_defined(ht[mt.row_key]))
+        return mt
+    else:
+        return ht
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -125,7 +125,7 @@ def explode_intervals_to_loci(
     Expand intervals to loci and key by loci, or return loci range expression.
 
     :param obj: Hail Table or Interval Expression.
-    :param interval_field: Name of the interval field if `obj` is a Hail Table.
+    :param interval_field: Name of the interval field. Only required if input is a Hail Table. Default is None.
     :param keep_intervals: If True, keep the original intervals as a column in output, if `obj` is a Hail Table. Default is False.
     :return: If input is a Hail Table, returns exploded Table keyed by locus. If input is an IntervalExpression, returns position array expression.
     """
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -145,7 +145,26 @@ def explode_intervals_to_loci(
           intervals_expr.end.position + 1,
           intervals_expr.end.position
       )
-        return hl.range(interval_start_expr, interval_end_expr)
+    if isinstance(intervals, hl.Table):
+        intervals = intervals.annotate(pos=hl.range(intervals_start_expr, intervals_end_expr)).explode(
+            "pos"
+        )
+        intervals = intervals.key_by(
+            locus=hl.locus(
+                intervals[interval_field].start.contig,
+                intervals.pos,
+                reference_genome=get_reference_genome(intervals[interval_field]),
+            )
+        )
+
+        fields_to_drop = ["pos"]
+        if not keep_intervals:
+            fields_to_drop.append(interval_field)
+
+        return intervals.drop(*fields_to_drop)
+        
+    logger.warning("Input is an IntervalExpression, so function will return ArrayExpression of positions  within input intervals. To fully explode intervals to loci, we recommend annotating your dataset with the returned ArrayExpression, exploding the array, and converting the positions to loci!")
+    return hl.range(intervals_start_expr, intervals_end_expr)
 
     elif isinstance(obj, hl.Table):
         ht = obj
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -134,15 +134,17 @@ def explode_intervals_to_loci(
     if isinstance(intervals, hl.Table) and (not interval_field or keep_intervals is None):
         raise ValueError("`interval_field` and `keep_intervals` must be defined if input is a Table!")
     assert interval_field in intervals.row, "`interval_field` must be an annotation present on input Table!"
-        interval = obj
-        interval_start_expr = hl.if_else(
-            interval.includes_start,
-            interval.start.position,
-            interval.start.position + 1,
-        )
-        interval_end_expr = hl.if_else(
-            interval.includes_end, interval.end.position + 1, interval.end.position
-        )
+      intervals_expr = intervals if isinstance(intervals, hl.expr.IntervalExpression) else intervals[interval_field]
+      intervals_start_expr = hl.if_else(
+          intervals_expr.includes_start,
+          intervals_expr.start.position,
+          intervals_expr.start.position + 1,
+      )
+      intervals_end_expr = hl.if_else(
+          intervals_expr.includes_end,
+          intervals_expr.end.position + 1,
+          intervals_expr.end.position
+      )
         return hl.range(interval_start_expr, interval_end_expr)
 
     elif isinstance(obj, hl.Table):
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -118,8 +118,8 @@ def pad_intervals(
 
 def explode_intervals_to_loci(
     obj: Union[hl.Table, hl.expr.IntervalExpression],
-    interval_field: str = "interval",
-    keep_intervals: bool = False,
+    interval_field: Optional[str] = None,
+    keep_intervals: Optional[bool] = None,
 ) -> Union[hl.Table, hl.expr.ArrayExpression]:
     """
     Expand intervals to loci and key by loci, or return loci range expression.
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -144,13 +144,5 @@ def explode_intervals_to_loci(
     if not keep_intervals:
         fields_to_drop.append(interval_field)
 
-    ht = ht.drop(*fields_to_drop)
+    return ht.drop(*fields_to_drop)
 
-    if is_matrix:
-        mt = obj
-        ht = ht.select_globals()
-        mt = mt.annotate_rows(**ht[mt.row_key])
-        mt = mt.filter_rows(hl.is_defined(ht[mt.row_key]))
-        return mt
-    else:
-        return ht
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -128,13 +128,8 @@ def explode_intervals_to_loci(
     :return: Hail Table with intervals exploded to loci.
     """
     interval_expr = ht[interval_field]
-    includes_start = interval.includes_start.take(1)[0]
-    includes_end = interval.includes_end.take(1)[0]
-
-    interval_start = (
-        interval.start.position if includes_start else interval.start.position + 1
-    )
-    interval_end = interval.end.position + 1 if includes_end else interval.end.position
+    interval_start_expr = hl.if_else(interval_expr.includes_start, interval_expr.start.position, interval_expr.start.position + 1)
+    interval_end_expr = hl.if_else(interval_expr.includes_end, interval_expr.end.position + 1, interval_expr.end.position)
 
     ht = ht.annotate(pos=hl.range(interval_start, interval_end)).explode("pos")
     ht = ht.key_by(
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -3,6 +3,7 @@
 from typing import List, Union
 
 import hail as hl
+from gnomad.utils.reference_genome import get_reference_genome
 
 
 def sort_intervals(intervals: List[hl.Interval]):
@@ -115,34 +116,60 @@ def pad_intervals(
 
 
 def explode_intervals_to_loci(
-    ht: hl.Table,
+    obj: Union[hl.Table, hl.expr.IntervalExpression],
     interval_field: str = "interval",
     keep_intervals: bool = False,
-) -> hl.Table,
+) -> Union[hl.Table, hl.expr.ArrayExpression]:
     """
-    Expand intervals to loci and key by loci.
+    Expand intervals to loci and key by loci, or return loci range expression.
 
-    :param obj: Hail Table with intervals to be exploded.
-    :param interval_field: Name of the interval field. Default is 'interval'.
-    :param keep_intervals: If True, keep the original intervals as a column in output.
-    :return: Hail Table with intervals exploded to loci.
+    :param obj: Hail Table or Interval Expression.
+    :param interval_field: Name of the interval field if `obj` is a Hail Table.
+    :param keep_intervals: If True, keep the original intervals as a column in output, if `obj` is a Hail Table. Default is False.
+    :return: If input is a Hail Table, returns exploded Table keyed by locus. If input is an IntervalExpression, returns position array expression.
     """
-    interval_expr = ht[interval_field]
-    interval_start_expr = hl.if_else(interval_expr.includes_start, interval_expr.start.position, interval_expr.start.position + 1)
-    interval_end_expr = hl.if_else(interval_expr.includes_end, interval_expr.end.position + 1, interval_expr.end.position)
-
-    ht = ht.annotate(pos=hl.range(interval_start, interval_end)).explode("pos")
-    ht = ht.key_by(
-        locus=hl.locus(
-            ht[interval_field].start.contig,
-            ht.pos,
-            reference_genome=get_reference_genome(ht[interval_field])
+    if isinstance(obj, hl.expr.IntervalExpression):
+        interval = obj
+        interval_start_expr = hl.if_else(
+            interval.includes_start,
+            interval.start.position,
+            interval.start.position + 1,
         )
-    )
+        interval_end_expr = hl.if_else(
+            interval.includes_end, interval.end.position + 1, interval.end.position
+        )
+        return hl.range(interval_start_expr, interval_end_expr)
 
-    fields_to_drop = ["pos"]
-    if not keep_intervals:
-        fields_to_drop.append(interval_field)
+    elif isinstance(obj, hl.Table):
+        ht = obj
+        interval_expr = ht[interval_field]
+        interval_start_expr = hl.if_else(
+            interval_expr.includes_start,
+            interval_expr.start.position,
+            interval_expr.start.position + 1,
+        )
+        interval_end_expr = hl.if_else(
+            interval_expr.includes_end,
+            interval_expr.end.position + 1,
+            interval_expr.end.position,
+        )
 
-    return ht.drop(*fields_to_drop)
+        ht = ht.annotate(pos=hl.range(interval_start_expr, interval_end_expr)).explode(
+            "pos"
+        )
+        ht = ht.key_by(
+            locus=hl.locus(
+                ht[interval_field].start.contig,
+                ht.pos,
+                reference_genome=get_reference_genome(ht[interval_field]),
+            )
+        )
 
+        fields_to_drop = ["pos"]
+        if not keep_intervals:
+            fields_to_drop.append(interval_field)
+
+        return ht.drop(*fields_to_drop)
+
+    else:
+        raise TypeError("Input must be a Hail Table or a Hail Interval Expression.")
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -112,3 +112,27 @@ def pad_intervals(
         return [_add_padding(i) for i in intervals]
     else:
         return _add_padding(intervals)
+
+
+def explode_intervals_to_loci(
+    ht: hl.Table,
+    keep_intervals: bool = False,
+) -> hl.Table:
+    """
+    Expand intervals to loci.
+
+    :param ht: Hail Table with an interval field.
+    :param keep_intervals: If True, keep the original interval as a column in output.
+    :return: Hail Table keyed by loci and intervals as optional field.
+    """
+    ht = ht.annotate(
+        pos=hl.range(ht.interval.start.position, ht.interval.end.position + 1),
+    ).explode("pos")
+    ht = ht.annotate(
+        locus=hl.locus(
+            ht.interval.start.contig,
+            ht.pos,
+            reference_genome=ht.interval.start.dtype.reference_genome,
+        ),
+    ).key_by("locus")
+    return ht.drop("interval", "pos") if not keep_intervals else ht.drop("pos")
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -140,13 +140,13 @@ def explode_intervals_to_loci(
     interval_end = interval.end.position + 1 if includes_end else interval.end.position
 
     ht = ht.annotate(pos=hl.range(interval_start, interval_end)).explode("pos")
-    ht = ht.annotate(
+    ht = ht.key_by(
         locus=hl.locus(
             ht[interval_field].start.contig,
             ht.pos,
-            reference_genome=str(interval.start.take(1)[0].reference_genome),
+            reference_genome=get_reference_genome(ht[interval_field])
         )
-    ).key_by("locus")
+    )
 
     fields_to_drop = ["pos"]
     if not keep_intervals:
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -117,7 +117,7 @@ def pad_intervals(
 
 
 def explode_intervals_to_loci(
-    obj: Union[hl.Table, hl.expr.IntervalExpression],
+    intervals: Union[hl.Table, hl.expr.IntervalExpression],
     interval_field: Optional[str] = None,
     keep_intervals: Optional[bool] = None,
 ) -> Union[hl.Table, hl.expr.ArrayExpression]:
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -127,9 +127,6 @@ def explode_intervals_to_loci(
     :param keep_intervals: If True, keep the original intervals as a column in output.
     :return: Hail Table with intervals exploded to loci.
     """
-    is_matrix = isinstance(obj, hl.MatrixTable)
-    ht = obj.rows() if is_matrix else obj
-
     interval_expr = ht[interval_field]
     includes_start = interval.includes_start.take(1)[0]
     includes_end = interval.includes_end.take(1)[0]
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -130,7 +130,7 @@ def explode_intervals_to_loci(
     is_matrix = isinstance(obj, hl.MatrixTable)
     ht = obj.rows() if is_matrix else obj
 
-    interval = ht[interval_field]
+    interval_expr = ht[interval_field]
     includes_start = interval.includes_start.take(1)[0]
     includes_end = interval.includes_end.take(1)[0]
 
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -118,7 +118,7 @@ def explode_intervals_to_loci(
     ht: hl.Table,
     interval_field: str = "interval",
     keep_intervals: bool = False,
-) -> Union[hl.Table, hl.MatrixTable]:
+) -> hl.Table,
     """
     Expand intervals to loci and key by loci.
 
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -166,36 +166,3 @@ def explode_intervals_to_loci(
     logger.warning("Input is an IntervalExpression, so function will return ArrayExpression of positions  within input intervals. To fully explode intervals to loci, we recommend annotating your dataset with the returned ArrayExpression, exploding the array, and converting the positions to loci!")
     return hl.range(intervals_start_expr, intervals_end_expr)
 
-    elif isinstance(obj, hl.Table):
-        ht = obj
-        interval_expr = ht[interval_field]
-        interval_start_expr = hl.if_else(
-            interval_expr.includes_start,
-            interval_expr.start.position,
-            interval_expr.start.position + 1,
-        )
-        interval_end_expr = hl.if_else(
-            interval_expr.includes_end,
-            interval_expr.end.position + 1,
-            interval_expr.end.position,
-        )
-
-        ht = ht.annotate(pos=hl.range(interval_start_expr, interval_end_expr)).explode(
-            "pos"
-        )
-        ht = ht.key_by(
-            locus=hl.locus(
-                ht[interval_field].start.contig,
-                ht.pos,
-                reference_genome=get_reference_genome(ht[interval_field]),
-            )
-        )
-
-        fields_to_drop = ["pos"]
-        if not keep_intervals:
-            fields_to_drop.append(interval_field)
-
-        return ht.drop(*fields_to_drop)
-
-    else:
-        raise TypeError("Input must be a Hail Table or a Hail Interval Expression.")
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -126,7 +126,7 @@ def explode_intervals_to_loci(
 
     :param obj: Hail Table or Interval Expression.
     :param interval_field: Name of the interval field. Only required if input is a Hail Table. Default is None.
-    :param keep_intervals: If True, keep the original intervals as a column in output, if `obj` is a Hail Table. Default is False.
+    :param keep_intervals: If True, keep the original intervals as a column in output. Only applies if input is a Hail Table. Default is False.
     :return: If input is a Hail Table, returns exploded Table keyed by locus. If input is an IntervalExpression, returns position array expression.
     """
     if isinstance(obj, hl.expr.IntervalExpression):
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -122,7 +122,7 @@ def explode_intervals_to_loci(
     """
     Expand intervals to loci.
 
-    :param obj: Hail Table or MatrixTable with an interval field.
+    :param obj: Hail Table with intervals to be exploded.
     :param interval_field: Name of the interval field. Default is 'interval'.
     :param keep_intervals: If True, keep the original interval as a column in output.
     :return: Hail Table or MatrixTable with interval exploded to loci.
--- a/gnomad/utils/intervals.py
+++ b/gnomad/utils/intervals.py
@@ -133,7 +133,7 @@ def explode_intervals_to_loci(
         locus=hl.locus(
             ht.interval.start.contig,
             ht.pos,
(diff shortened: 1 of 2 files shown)


gnomad/utils/intervals.py
```diff
@@ -112,3 +112,27 @@ def _add_padding(
...
+
+def explode_intervals_to_loci(
+    ht: hl.Table,
+    keep_intervals: bool = False,
+) -> hl.Table:
+    """
+    Expand intervals to loci.
+
+    :param ht: Hail Table with an interval field.
+    :param keep_intervals: If True, keep the original interval as a column in output.
```
ch-kr:
```suggestion
    :param keep_intervals: If True, keep the original interval as a column in output.
        Default is False.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,27 @@ def _add_padding(
...
+    :return: Hail Table keyed by loci and intervals as optional field.
+    """
+    ht = ht.annotate(
+        pos=hl.range(ht.interval.start.position, ht.interval.end.position + 1),
```
ch-kr: this should double check whether the interval includes the start and end positions and adjust accordingly

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,27 @@ def _add_padding(
...
+    ).explode("pos")
+    ht = ht.annotate(
+        locus=hl.locus(
+            ht.interval.start.contig,
+            ht.pos,
+            reference_genome=ht.interval.start.dtype.reference_genome,
```
ch-kr:
```suggestion
            reference_genome=ht.interval.start.reference_genome,
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,27 @@ def _add_padding(
         return [_add_padding(i) for i in intervals]
     else:
         return _add_padding(intervals)
+
+
+def explode_intervals_to_loci(
```
ch-kr, line 149: could you make this function slightly more general to also accept a MatrixTable as input?

ch-kr (changes requested):
thanks for adding! a few suggestions

gnomad/utils/intervals.py
Ruchit10: Done

gnomad/utils/intervals.py
Ruchit10: Done

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    obj: Union[hl.Table, hl.MatrixTable],
```
ch-kr: sorry, I know I suggested this, but on seeing the code/second thought, I think it doesn't make sense to support a MatrixTable here. an alternative path for flexibility would be to have the function accept an interval expression as the input and return a position expression (from `pos=hl.range(interval_start, interval_end)`), which would mean the user would have to run the annotate, explode, and `key_by` downstream. I'll add suggestions to drop the extra flexibility to hopefully save you some time
```suggestion
    ht: hl.Table,
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    interval_field: str = "interval",
+    keep_intervals: bool = False,
+) -> Union[hl.Table, hl.MatrixTable]:
+    """
+    Expand intervals to loci.
+
+    :param obj: Hail Table or MatrixTable with an interval field.
```
ch-kr:
```suggestion
    :param obj: Hail Table with intervals to be exploded.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    :param interval_field: Name of the interval field. Default is 'interval'.
+    :param keep_intervals: If True, keep the original interval as a column in output.
```
ch-kr:
```suggestion
    :param keep_intervals: If True, keep the original intervals as a column in output.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    :return: Hail Table or MatrixTable with interval exploded to loci.
```
ch-kr:
```suggestion
    :return: Hail Table with intervals exploded to loci.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    """
+    is_matrix = isinstance(obj, hl.MatrixTable)
+    ht = obj.rows() if is_matrix else obj
+
+    interval = ht[interval_field]
```
ch-kr:
```suggestion
    interval_expr = ht[interval_field]
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    interval_end = interval.end.position + 1 if includes_end else interval.end.position
+
+    ht = ht.annotate(pos=hl.range(interval_start, interval_end)).explode("pos")
+    ht = ht.annotate(
+        locus=hl.locus(
+            ht[interval_field].start.contig,
+            ht.pos,
+            reference_genome=str(interval.start.take(1)[0].reference_genome),
+        )
+    ).key_by("locus")
```
ch-kr:
```suggestion
    ht = ht.key_by(
        locus=hl.locus(
            ht[interval_field].start.contig,
            ht.pos,
            reference_genome=get_reference_genome(ht[interval_field])
        )
    )
```
this will need to be imported if it isn't already: https://github.com/broadinstitute/gnomad_methods/blob/main/gnomad/utils/reference_genome.py#L124

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
         return _add_padding(intervals)
+
+
+def explode_intervals_to_loci(
+    obj: Union[hl.Table, hl.MatrixTable],
+    interval_field: str = "interval",
+    keep_intervals: bool = False,
+) -> Union[hl.Table, hl.MatrixTable]:
+    """
+    Expand intervals to loci.
```
ch-kr:
```suggestion
    Expand intervals to loci and key by loci.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
         return [_add_padding(i) for i in intervals]
     else:
         return _add_padding(intervals)
+
+
+def explode_intervals_to_loci(
+    obj: Union[hl.Table, hl.MatrixTable],
+    interval_field: str = "interval",
+    keep_intervals: bool = False,
+) -> Union[hl.Table, hl.MatrixTable]:
```
ch-kr:
```suggestion
) -> hl.Table,
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    Expand intervals to loci.
+
+    :param obj: Hail Table or MatrixTable with an interval field.
+    :param interval_field: Name of the interval field. Default is 'interval'.
+    :param keep_intervals: If True, keep the original interval as a column in output.
+    :return: Hail Table or MatrixTable with interval exploded to loci.
+    """
+    is_matrix = isinstance(obj, hl.MatrixTable)
+    ht = obj.rows() if is_matrix else obj
+
```
ch-kr:
```suggestion
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    interval = ht[interval_field]
+    includes_start = interval.includes_start.take(1)[0]
+    includes_end = interval.includes_end.take(1)[0]
+
+    interval_start = (
+        interval.start.position if includes_start else interval.start.position + 1
+    )
+    interval_end = interval.end.position + 1 if includes_end else interval.end.position
```
ch-kr:
```suggestion
    interval_start_expr = hl.if_else(interval_expr.includes_start, interval_expr.start.position, interval_expr.start.position + 1)
    interval_end_expr = hl.if_else(interval_expr.includes_end, interval_expr.end.position + 1, interval_expr.end.position)
```
this avoids the conversion from Hail to python with `take` (though it might not be properly formatted)

gnomad/utils/intervals.py
```diff
@@ -112,3 +112,53 @@ def _add_padding(
...
+    ht = ht.drop(*fields_to_drop)
+
+    if is_matrix:
+        mt = obj
+        ht = ht.select_globals()
+        mt = mt.annotate_rows(**ht[mt.row_key])
+        mt = mt.filter_rows(hl.is_defined(ht[mt.row_key]))
+        return mt
+    else:
+        return ht
```
ch-kr:
```suggestion
    return ht.drop(*fields_to_drop)

```

ch-kr (commented):
a few more minor changes -- sorry about the back and forth on the MT part!

gnomad/utils/intervals.py
```diff
@@ -112,3 +114,63 @@ def _add_padding(
         return [_add_padding(i) for i in intervals]
     else:
         return _add_padding(intervals)
+
+
+def explode_intervals_to_loci(
+    obj: Union[hl.Table, hl.expr.IntervalExpression],
+    interval_field: str = "interval",
+    keep_intervals: bool = False,
```
ch-kr: thank you for adding the flexibility of accepting a Table or IntervalExpression! adding a suggestion to make these two fields Optional now that they only apply to a Table input
```suggestion
    interval_field: Optional[str] = None,
    keep_intervals: Optional[bool] = None,
```
you'll also need to import Optional above -- I can't comment on the line directly, but it'll be
```
from typing import List, Optional, Union
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +114,63 @@ def _add_padding(
...
+) -> Union[hl.Table, hl.expr.ArrayExpression]:
+    """
+    Expand intervals to loci and key by loci, or return loci range expression.
+
+    :param obj: Hail Table or Interval Expression.
+    :param interval_field: Name of the interval field if `obj` is a Hail Table.
```
ch-kr:
```suggestion
    :param interval_field: Name of the interval field. Only required if input is a Hail Table. Default is None.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +114,63 @@ def _add_padding(
...
+    :param keep_intervals: If True, keep the original intervals as a column in output, if `obj` is a Hail Table. Default is False.
```
ch-kr:
```suggestion
    :param keep_intervals: If True, keep the original intervals as a column in output. Only applies if input is a Hail Table. Default is False.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +114,63 @@ def _add_padding(
         return [_add_padding(i) for i in intervals]
     else:
         return _add_padding(intervals)
+
+
+def explode_intervals_to_loci(
+    obj: Union[hl.Table, hl.expr.IntervalExpression],
```
ch-kr:
```suggestion
    intervals: Union[hl.Table, hl.expr.IntervalExpression],
```
I think this should be named something more informative than `obj`, so I'll suggest intervals (but open to other ideas!)

gnomad/utils/intervals.py
```diff
@@ -112,3 +114,63 @@ def _add_padding(
...
+) -> Union[hl.Table, hl.expr.ArrayExpression]:
+    """
+    Expand intervals to loci and key by loci, or return loci range expression.
+
+    :param obj: Hail Table or Interval Expression.
+    :param interval_field: Name of the interval field if `obj` is a Hail Table.
+    :param keep_intervals: If True, keep the original intervals as a column in output, if `obj` is a Hail Table. Default is False.
+    :return: If input is a Hail Table, returns exploded Table keyed by locus. If input is an IntervalExpression, returns position array expression.
+    """
+    if isinstance(obj, hl.expr.IntervalExpression):
```
ch-kr:
```suggestion
    assert isintance(intervals, hl.Table) or isinstance(intervals, hl.expr.IntervalExpression), "Input must be a Table or IntervalExpression!"

    if isinstance(intervals, hl.Table) and (not interval_field or keep_intervals is None):
        raise ValueError("`interval_field` and `keep_intervals` must be defined if input is a Table!")
    assert interval_field in intervals.row, "`interval_field` must be an annotation present on input Table!"
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +114,63 @@ def _add_padding(
...
+        interval = obj
+        interval_start_expr = hl.if_else(
+            interval.includes_start,
+            interval.start.position,
+            interval.start.position + 1,
+        )
+        interval_end_expr = hl.if_else(
+            interval.includes_end, interval.end.position + 1, interval.end.position
+        )
```
ch-kr:
```suggestion
      intervals_expr = intervals if isinstance(intervals, hl.expr.IntervalExpression) else intervals[interval_field]
      intervals_start_expr = hl.if_else(
          intervals_expr.includes_start,
          intervals_expr.start.position,
          intervals_expr.start.position + 1,
      )
      intervals_end_expr = hl.if_else(
          intervals_expr.includes_end,
          intervals_expr.end.position + 1,
          intervals_expr.end.position
      )
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +114,63 @@ def _add_padding(
...
+        return hl.range(interval_start_expr, interval_end_expr)
```
ch-kr:
```suggestion
    if isinstance(intervals, hl.Table):
        intervals = intervals.annotate(pos=hl.range(intervals_start_expr, intervals_end_expr)).explode(
            "pos"
        )
        intervals = intervals.key_by(
            locus=hl.locus(
                intervals[interval_field].start.contig,
                intervals.pos,
                reference_genome=get_reference_genome(intervals[interval_field]),
            )
        )

        fields_to_drop = ["pos"]
        if not keep_intervals:
            fields_to_drop.append(interval_field)

        return intervals.drop(*fields_to_drop)
        
    logger.warning("Input is an IntervalExpression, so function will return ArrayExpression of positions  within input intervals. To fully explode intervals to loci, we recommend annotating your dataset with the returned ArrayExpression, exploding the array, and converting the positions to loci!")
    return hl.range(intervals_start_expr, intervals_end_expr)
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +114,63 @@ def _add_padding(
...
+        )
+
+        fields_to_drop = ["pos"]
+        if not keep_intervals:
+            fields_to_drop.append(interval_field)
+
+        return ht.drop(*fields_to_drop)
+
+    else:
+        raise TypeError("Input must be a Hail Table or a Hail Interval Expression.")
```
ch-kr:
```suggestion
```

ch-kr (commented):
thank you for adding additional flexibility! I've added some suggestions to streamline the code (and hopefully didn't break the formatting too much)

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
...
+
+def explode_intervals_to_loci(
+    intervals: Union[hl.Table, hl.expr.IntervalExpression],
+    interval_field: Optional[str] = None,
+    keep_intervals: Optional[bool] = False,
+) -> Union[hl.Table, hl.expr.ArrayExpression]:
+    """
+    Expand intervals to loci and key by loci, or return loci range expression.
+
+    :param intervals: Hail Table or Interval Expression.
```
ch-kr:
```suggestion
    :param intervals: Table or IntervalExpression.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
...
         return _add_padding(intervals)
+
+
+def explode_intervals_to_loci(
+    intervals: Union[hl.Table, hl.expr.IntervalExpression],
+    interval_field: Optional[str] = None,
+    keep_intervals: Optional[bool] = False,
+) -> Union[hl.Table, hl.expr.ArrayExpression]:
+    """
+    Expand intervals to loci and key by loci, or return loci range expression.
```
ch-kr: Suggested change
      
    
    
      
          
            
                Expand intervals to loci and key by loci, or return loci range expression.
          
          
            
                Expand interval(s) to loci.
          
          
            
                
          
          
            
                If input is a Table, function will expand intervals to loci and key Table by loci. If input is an IntervalExpression, function will return an ArrayExpression containing all loci within input interval.
          
          
            
                
          
          
            
                .. warning::
          
          
            
                
          
          
            
                    - Do not use this function on very large intervals (e.g., whole chromosomes), as it will create extremely large arrays, which may cause performance issues.
                    - This function will produce duplicate loci if there are overlapping intervals in the input Table.

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
...
+
+    :param intervals: Hail Table or Interval Expression.
+    :param interval_field: Name of the interval field. Only required if input is a Hail Table. Default is None.
+    :param keep_intervals: If True, keep the original intervals as a column in output. Only applies if input is a Hail Table. Default is False.
+    :return: If input is a Hail Table, returns exploded Table keyed by locus. If input is an IntervalExpression, returns position array expression.
```
ch-kr:
```suggestion
    :return: If input is a Hail Table, returns exploded Table keyed by locus. If input is an IntervalExpression, returns ArrayExpression containing loci within input interval.
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
...
+    ), "Input must be a Table or IntervalExpression!"
+
+    if isinstance(intervals, hl.Table) and (
+        not interval_field or keep_intervals is None
+    ):
+        raise ValueError(
+            "`interval_field` and `keep_intervals` must be defined if input is a Table!"
+        )
+    if isinstance(intervals, hl.Table):
+        assert (
```
ch-kr: can you raise an error here rather than using an assertion for consistency with the interval_field and keep_intervals check above?

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
         return [_add_padding(i) for i in intervals]
     else:
         return _add_padding(intervals)
+
+
+def explode_intervals_to_loci(
```
ch-kr, line 149: nitpick, but this is only plural ("intervals") if the input is a Table. could you update this function to accept a table or list of intervals?

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
...
+            )
+        )
+
+        fields_to_drop = ["pos"]
+        if not keep_intervals:
+            fields_to_drop.append(interval_field)
+
+        return intervals.drop(*fields_to_drop)
+
+    logger.warning(
```
ch-kr:
```suggestion
    logger.info(
```

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
...
+        "Input is an IntervalExpression, so function will return ArrayExpression of"
+        " positions  within input intervals. To fully explode intervals to loci, we"
```
ch-kr:
```suggestion
        " positions within input intervals. To fully explode intervals to loci, we"
```

gnomad/utils/intervals.py
ch-kr: for that second warning, we could consider adding a distinct() after the table rekey, though this won't change that there couple be multiple duplicate loci in the returned ArrayExpression if you change this function to allow an array of IntervalExpressions as input

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
...
+        " recommend annotating your dataset with the returned ArrayExpression,"
+        " exploding the array, and converting the positions to loci!"
+    )
+    return hl.range(intervals_start_expr, intervals_end_expr)
```
ch-kr: we should probably add to the docstring that intervals that cross chromosomes are not supported

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,72 @@ def _add_padding(
...
+        ).explode("pos")
+        intervals = intervals.key_by(
+            locus=hl.locus(
+                intervals[interval_field].start.contig,
+                intervals.pos,
+                reference_genome=get_reference_genome(intervals[interval_field]),
+            )
+        )
+
+        fields_to_drop = ["pos"]
```
ch-kr:
```suggestion
            _pos=hl.range(intervals_start_expr, intervals_end_expr)
        ).explode("_pos")
        intervals = intervals.key_by(
            locus=hl.locus(
                intervals[interval_field].start.contig,
                intervals._pos,
                reference_genome=get_reference_genome(intervals[interval_field]),
            )
        )

        fields_to_drop = ["_pos"]
```

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,385 @@
...
+        )
+
+    def test_explode_table_with_includes_start_and_end(
+        self, sample_interval_table: hl.Table
+    ) -> None:
+        """
+        Test exploding a table with intervals that include both start and end positions.
+
+        :param sample_interval_table: Sample Hail Table with intervals.
+        :return: None.
```
ch-kr: minor point: these test functions don't need to document `:param:` and `:return:`

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,385 @@
...
+
+
+class TestExplodeIntervalsToLoci:
+    """Test the explode_intervals_to_loci function."""
+
+    @pytest.fixture
+    def sample_interval_table(self):
+        """Fixture to create a sample Hail Table with intervals."""
+        intervals = [
+            hl.Interval(
```
ch-kr: should this be `hl.interval`?

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,385 @@
...
+            hl.tstruct(
+                interval=hl.tinterval(hl.tlocus("GRCh38")),
+                gene=hl.tstr,
+            ),
+        )
+
+    @pytest.fixture
+    def sample_interval_expr(self):
+        """Fixture to create a sample interval expression."""
+        return hl.literal(
```
ch-kr: do you need this literal? `hl.interval` returns an IntervalExpression

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,385 @@
...
+    def test_explode_table_with_includes_start_and_end(
+        self, sample_interval_table: hl.Table
+    ) -> None:
+        """
+        Test exploding a table with intervals that include both start and end positions.
+
+        :param sample_interval_table: Sample Hail Table with intervals.
+        :return: None.
+        """
+        # Explode the intervals to loci.
```
ch-kr: nitpick: most of the comments aren't necessary (the code is self-explanatory)

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,385 @@
...
+
+        # Expected positions (101-104, both excluded).
+        expected_positions = list(range(101, 105))
+
+        # Verify the result.
+        assert positions == expected_positions
+
+    def test_explode_table_single_position_interval(self) -> None:
+        """
+        Test exploding a table with a single-position interval.
```
ch-kr: should there also be a function that tests the case where both start and end are excluded for a single position interval?

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,385 @@
...
+
+        # Expected positions (100-105, both inclusive).
+        expected_positions = list(range(100, 106))
+
+        # Verify the result.
+        assert positions == expected_positions
+
+    def test_explode_interval_expression_excludes_start(self) -> None:
+        """
+        Test exploding an interval expression that excludes the start position.
```
ch-kr: should there also be inclusion/exclusion tests when the input is a Table?

tests/utils/test_interval_utils.py
ch-kr: also, should there be a test when the input is a table with this kind of empty single position interval (i.e., the return table will have 0 rows)?

tests/utils/test_interval_utils.py
ch-kr, line 1: should there be tests for when there are overlapping intervals?

ch-kr (commented):
thanks for adding tests! a few more thoughts

tests/utils/test_interval_utils.py
Ruchit10: Ah this is very silly on my end. I was using `hl.Interval` instead of `hl.interval` before so was wondering why the interval function wasn't working (which is reason for adding hl.literal). Will change this

tests/utils/test_interval_utils.py
Ruchit10: yes absolutely

gnomad/utils/intervals.py
Ruchit10: For first warning how about
```
- Caution when using this function on very large intervals (e.g. whole chromosomes), as it will create extremely large arrays, which may cause performance issues.
```
So we're not saying "don't use it" but more like "beware, it will run slow"?

For second, I can change it to accept array of IntervalExpression but was thinking of keeping this simple enough to use inside a lambda function if dealing with array of intervals (which might run slow but that wasn't the point of adding this utility, just that it supports it)

Ruchit10: Okay after reading further comments down I changed my mind and will add list of intervals as input and flatmap it to a single array so that we can de-duplicate

gnomad/utils/intervals.py
```diff
@@ -125,21 +125,41 @@ def _add_padding(
...
+
+    If input is an IntervalExpression or a list of IntervalExpressions, function will return an ArrayExpression containing all loci within the input interval(s).
+
+    .. warning::
+        - Overlapping intervals will produce duplicate loci. Use ``deduplicate=True`` (the default) to remove them.
+        - When ``keep_intervals=True`` on a Table input, deduplication is not possible because duplicate rows with different interval annotations may exist; a warning is displayed instead.
+        - Caution when using this function on very larg intervals (e.g. whole chromosomes), as it will create extremely large arrays, which may cause performance issues.
 
-    :param intervals: Hail Table or Interval Expression.
+    Note that Intervals that cross chromosomes is currently not supported.
```
ch-kr:
```suggestion
    Note that Intervals that cross chromosomes are currently not supported.
```
also, should this be in a  `.. note ::` block?

gnomad/utils/intervals.py
```diff
@@ -125,21 +125,41 @@ def _add_padding(
...
+    Expand interval(s) to loci.
+
+    If input is a Table, function will expand intervals to loci and key Table by loci.
+
+    If input is an IntervalExpression or a list of IntervalExpressions, function will return an ArrayExpression containing all loci within the input interval(s).
+
+    .. warning::
+        - Overlapping intervals will produce duplicate loci. Use ``deduplicate=True`` (the default) to remove them.
+        - When ``keep_intervals=True`` on a Table input, deduplication is not possible because duplicate rows with different interval annotations may exist; a warning is displayed instead.
+        - Caution when using this function on very larg intervals (e.g. whole chromosomes), as it will create extremely large arrays, which may cause performance issues.
```
ch-kr:
```suggestion
        - Caution when using this function on very large intervals (e.g., whole chromosomes), as it will create extremely large arrays, which may cause performance issues.
```

gnomad/utils/intervals.py
```diff
@@ -148,9 +168,45 @@ def explode_intervals_to_loci(
...
+            " ArrayExpression, exploding the array, and converting the positions to"
+            " loci!"
+        )
+        if not deduplicate:
+            logger.warning(
+                "Overlapping intervals in the input list may produce duplicate loci in"
+                " the returned ArrayExpression. Set `deduplicate=True` to remove them."
+            )
+
+        def _interval_to_range(interval_expr):
```
ch-kr: a few notes:
- add documentation for this function
- maybe this function should be moved outside of `explode_intervals_to_loci`
- should this be named interval_to_pos_range or something similar for clarity?

gnomad/utils/intervals.py
```diff
@@ -168,25 +224,38 @@ def explode_intervals_to_loci(
     )
     if isinstance(intervals, hl.Table):
         intervals = intervals.annotate(
-            pos=hl.range(intervals_start_expr, intervals_end_expr)
-        ).explode("pos")
+            _pos=hl.range(intervals_start_expr, intervals_end_expr)
+        ).explode("_pos")
```
ch-kr: this looks like it could be replaced with a call to the new `_interval_to_range` function above

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,141 @@ def _add_padding(
...
+            end = hl.if_else(
+                interval_expr.includes_end,
+                interval_expr.end.position + 1,
+                interval_expr.end.position,
+            )
+            return hl.range(start, end)
+
+        result = hl.array([_interval_to_range(i) for i in intervals]).flatmap(
+            lambda x: x
+        )
```
ch-kr: Hail has a `flatten` function that might be easier to read here. also, should this function support returning the nested array if preferred by the user?

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,141 @@ def _add_padding(
...
+    interval_field: Optional[str] = None,
+    keep_intervals: Optional[bool] = False,
+    deduplicate: bool = True,
+) -> Union[hl.Table, hl.expr.ArrayExpression]:
+    """
+    Expand interval(s) to loci.
+
+    If input is a Table, function will expand intervals to loci and key Table by loci.
+
+    If input is an IntervalExpression or a list of IntervalExpressions, function will return an ArrayExpression containing all loci within the input interval(s).
```
ch-kr: the function currently returns a list of positions, not loci; maybe it should return a list of loci for consistency with this comment and with the behavior when the user passes a table?

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,586 @@
...
+        assert len(result_loci) == len(expected_loci)
+        assert all(locus in result_loci for locus in expected_loci)
+
+    def test_explode_table_overlapping_intervals_keep_intervals_warns(
+        self, caplog
+    ) -> None:
+        """Test that a warning is emitted and duplicates remain when keep_intervals=True with overlapping intervals."""
+        # With keep_intervals=True and deduplicate=True, deduplication is skipped and
+        # a warning is emitted; overlapping positions appear as duplicate rows.
+        import logging
```
ch-kr: this should be imported at the top of the script

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,586 @@
...
+        # Expected loci for the first interval (100-105, both inclusive).
+        expected_loci_1 = [hl.Locus("chr1", pos, "GRCh38") for pos in range(100, 106)]
+        # Expected loci for the second interval (200-203, start inclusive, end
+        # exclusive).
+        expected_loci_2 = [hl.Locus("chr2", pos, "GRCh38") for pos in range(200, 203)]
+
+        # Get the loci from the result.
+        result_loci = [row.locus for row in result]
+
+        # Verify the result contains the expected loci.
```
ch-kr:
```suggestion
        result_ht = explode_intervals_to_loci(
            sample_interval_table, interval_field="interval", keep_intervals=False
        )

        result = result_ht.collect()
        
        expected_loci_1 = [hl.Locus("chr1", pos, "GRCh38") for pos in range(100, 106)]
        expected_loci_2 = [hl.Locus("chr2", pos, "GRCh38") for pos in range(200, 203)]

        result_loci = [row.locus for row in result]
```
general comment to remove these unnecessary comments from the tests

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,141 @@ def _add_padding(
...
+
+    if isinstance(intervals, hl.Table) and (
+        not interval_field or keep_intervals is None
+    ):
+        raise ValueError(
+            "`interval_field` and `keep_intervals` must be defined if input is a Table!"
+        )
+    if isinstance(intervals, hl.Table):
+        if interval_field not in intervals.row:
+            raise ValueError(
```
ch-kr, line 214: it's a little excessive, but maybe we should check to make sure that `interval_field` is an IntervalExpression after this check?

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,586 @@
...
+    def sample_interval_expr(self):
+        """Fixture to create a sample interval expression."""
+        return hl.interval(
+            hl.locus("chr1", 100, "GRCh38"),
+            hl.locus("chr1", 105, "GRCh38"),
+            includes_start=True,
+            includes_end=True,
+        )
+
+    def test_explode_table_with_includes_start_and_end(
```
ch-kr, line 52: can you add a test for the case where the table is exploded and loci are not deduplicated?

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,586 @@
...
+        )
+
+        # Collect results.
+        result = result_ht.collect()
+
+        # Expected: single locus at position 100.
+        assert len(result) == 1
+        assert result[0].locus == hl.Locus("chr1", 100, "GRCh38")
+
+    def test_explode_table_missing_interval_field_raises_error(
```
ch-kr, line 217: if you add the check above that looks for `interval_field` having the correct type, we should also add a test for that after this function

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,586 @@
...
+
+        # Collect results.
+        result = result_ht.collect()
+
+        # Verify that other fields are preserved.
+        assert all(row.gene == "GENE1" for row in result)
+        assert all(row.score == 42 for row in result)
+        assert len(result) == 3  # 3 positions: 100, 101, 102
+
+    def test_explode_invalid_input_type_raises_error(self) -> None:
```
ch-kr, line 314: can you also add a test for the case when an empty list? the function behaves differently when these two are passed:
```
explode_intervals_to_loci(hl.array()) # returns nothing
explode_intervals_to_loci([]) # throws TypeError
```

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,586 @@
...
+                includes_start=True,
+                includes_end=False,
+            ),
+        ]
+
+        result = explode_intervals_to_loci(intervals)
+        positions = hl.eval(result)
+
+        # deduplicate=True by default; set comparison since order is not guaranteed.
+        assert set(positions) == {100, 101, 102, 200, 201}
```
ch-kr: this is the expected behavior given the inputs, but what happens if there are duplicate positions from the different chromosomes (chr1:100-102 and chr2:100-102)?

ch-kr (commented):
a few more comments

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+    If input is a Table, function will expand intervals to loci and key Table by loci.
+
+    If input is an IntervalExpression or a list of IntervalExpressions, function will return an ArrayExpression containing all loci within the input interval(s).
+
+    .. warning::
+        - Overlapping intervals will produce duplicate loci. Use ``deduplicate=True`` (the default) to remove them.
+        - When ``keep_intervals=True`` on a Table input, deduplication is not possible because duplicate rows with different interval annotations may exist; a warning is displayed instead.
+        - Caution when using this function on very large intervals (e.g., whole chromosomes), as it will create extremely large arrays, which may cause performance issues.
+
+    Note that Intervals that cross chromosomes are currently not supported.
```
ch-kr:
```suggestion
    Note that intervals that cross chromosomes are currently not supported.
```
nit

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+        if not isinstance(intervals[interval_field], hl.expr.IntervalExpression):
+            raise ValueError(
+                f"`interval_field` '{interval_field}' must be a hail interval"
+                " expression in the input Table!"
+            )
+
+    if isinstance(intervals, list):
+        logger.info(
+            "Input is a list of IntervalExpressions, so function will return an"
+            " ArrayExpression of loci within all input intervals."
```
ch-kr: should these be consistent types? the current behavior means the user passes in a list but receives an ArrayExpression

gnomad/utils/intervals.py
ch-kr: two other points:
- I don't think this would be common, but should we add a test to make sure the user hasn't passed one of these types of intervals?
- this should be included in the warning block above

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+        )
+    if isinstance(intervals, hl.Table):
+        if interval_field not in intervals.row:
+            raise ValueError(
+                "`interval_field` must be an annotation present on input Table!"
+            )
+        if not isinstance(intervals[interval_field], hl.expr.IntervalExpression):
+            raise ValueError(
+                f"`interval_field` '{interval_field}' must be a hail interval"
+                " expression in the input Table!"
```
ch-kr:
```suggestion
                f"`interval_field` '{interval_field}' must be an IntervalExpression"
                " in the input Table!"
```
nit

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+        "Input is an IntervalExpression, so function will return an ArrayExpression of"
+        " loci within the input interval."
+    )
+    return interval_to_pos_range(intervals_expr).map(
+        lambda pos: hl.locus(
+            intervals_expr.start.contig,
+            pos,
+            reference_genome=intervals_expr.start.dtype.reference_genome,
+        )
+    )
```
ch-kr: this can be replaced with a call to `_make_loci_array`

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+            return interval_to_pos_range(interval_expr).map(
+                lambda pos: hl.locus(
+                    interval_expr.start.contig,
+                    pos,
+                    reference_genome=interval_expr.start.dtype.reference_genome,
+                )
+            )
+
+        loci_arrays = [_make_loci_array(i) for i in intervals]
+        if not flatten:
```
ch-kr: for consistency, this probably should be reordered to be `if flatten:`

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+        intervals = intervals.annotate(
+            _pos=interval_to_pos_range(intervals_expr)
+        ).explode("_pos")
+        intervals = intervals.key_by(
+            locus=hl.locus(
+                intervals[interval_field].start.contig,
+                intervals._pos,
+                reference_genome=get_reference_genome(intervals[interval_field]),
+            )
+        )
```
ch-kr: this can also be replaced by a call to `_make_loci_array`

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+        if deduplicate:
+            if keep_intervals:
+                logger.warning(
+                    "`deduplicate=True` has no effect when `keep_intervals=True`"
+                    " because rows with different interval annotations cannot be safely"
+                    " collapsed. Duplicate loci may be present in the output if"
+                    " intervals overlap."
+                )
+            else:
+                intervals = intervals.distinct()
```
ch-kr, line 280: should we add a warning here that `distinct()` arbitrarily deduplicates, which means that if the input table was annotated with something like gene or transcript ID, this will pick one at random for duplicated loci?

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+                )
+            )
+
+        loci_arrays = [_make_loci_array(i) for i in intervals]
+        if not flatten:
+            return hl.array(loci_arrays)
+        result = hl.flatten(hl.array(loci_arrays))
+        if deduplicate:
+            result = hl.array(hl.set(result))
+        return result
```
ch-kr: should there be a check to make sure the positions are sorted?

gnomad/utils/intervals.py
```diff
@@ -112,3 +122,160 @@ def _add_padding(
...
+
+    intervals_expr = (
```
ch-kr, line 254: it looks like this expression only gets used once (it doesn't get used if this is a Table below); is this code necessary?

ch-kr (commented):
a few more style comments

gnomad/utils/intervals.py
Ruchit10: I feel an ArrayExpression as input makes more sense in that case. That will be the most common input type in hail isn't it?

gnomad/utils/intervals.py
Ruchit10: It also gets used in a table to annotate `_pos`, but it gets wrapped into `interval_to_pos_range` either way so maybe I should make that a single call upstream

gnomad/utils/intervals.py
```diff
@@ -112,3 +120,175 @@ def _add_padding(
...
+    Expand interval(s) to loci.
+
+    If input is a Table, function will expand intervals to loci and key Table by loci.
+
+    If input is an IntervalExpression or an ArrayExpression of IntervalExpressions,
+    function will return an ArrayExpression containing all loci within the input
+    interval(s).
+
+    .. warning::
+        - Overlapping intervals will produce duplicate loci. Use ``deduplicate=True``
```
ch-kr, line 166:
```suggestion
    .. warning::
    
        - Overlapping intervals will produce duplicate loci. Use ``deduplicate=True``
```
nit

gnomad/utils/intervals.py
```diff
@@ -112,3 +120,175 @@ def _add_padding(
...
+          (default) to remove them.
+        - When ``keep_intervals=True`` on a Table input, deduplication is not possible
+          because duplicate rows with different interval annotations may exist; a warning
+          is displayed instead.
+        - Caution when using this function on very large intervals (e.g., whole
+          chromosomes), as it will create extremely large arrays, which may cause
+          performance issues.
+
+    Note that intervals that cross chromosomes are currently not supported.
```
ch-kr, line 175: maybe we should raise an error if a user passes an interval that crosses chromosomes?

gnomad/utils/intervals.py
```diff
@@ -112,3 +120,175 @@ def _add_padding(
...
+    assert (
+        isinstance(intervals, hl.Table)
+        or isinstance(intervals, hl.expr.IntervalExpression)
+        or isinstance(intervals, hl.expr.ArrayExpression)
+    ), (
+        "Input must be a Table, IntervalExpression, or ArrayExpression of"
+        " IntervalExpressions!"
+    )
+
+    if isinstance(intervals, hl.Table) and (
```
ch-kr, line 206: seeing this function again with fresh eyes, you could reorder how you tackle the different inputs (Table, IntervalExpression, ArrayExpression). the current function structure checks whether the input is a Table 3 separate times, and reordering would remove the extra checks.

by reordering, I mean this function could handle the case of an input IntervalExpression + return, then the case of an input ArrayExpression + return, and finally the case of an input Table

gnomad/utils/intervals.py
ch-kr: I think what I meant when I wrote this comment was that you could directly call `_make_loci_array` on `intervals[interval_field]`, since the only case that is handled below is `if isinstance(intervals, hl.Table)`. however, if you reorder the code as suggested above, you shouldn't need this if/else at all

gnomad/utils/intervals.py
```diff
@@ -112,3 +120,175 @@ def _add_padding(
...
+    )
+    return hl.range(start, end)
+
+
+def explode_intervals_to_loci(
+    intervals: Union[hl.Table, hl.expr.IntervalExpression, hl.expr.ArrayExpression],
+    interval_field: Optional[str] = None,
+    keep_intervals: Optional[bool] = False,
+    deduplicate: bool = True,
+    flatten: bool = True,
```
ch-kr, line 154: there doesn't seem to be a test covering `flatten=False` in the test suite

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,680 @@
...
+
+        The API now requires a Hail ``ArrayExpression`` instead of a Python list.
+        Passing a Python list (including an empty one) fails the input-type
+        assertion. For an empty typed array use
+        ``hl.literal([], hl.tarray(hl.tinterval(hl.tlocus(...))))``).
+        """
+        with pytest.raises(AssertionError):
+            explode_intervals_to_loci([])
+
+    def test_explode_list_cross_chromosome_same_position_deduplicates_correctly(
```
ch-kr, line 647: this test will get replaced if you update the function to raise an error for intervals that cross chromosomes

gnomad/utils/intervals.py
```diff
@@ -112,3 +120,175 @@ def _add_padding(
...
+                " in the returned ArrayExpression. Set `deduplicate=True` to remove"
+                " them."
+            )
+
+        loci_arrays = intervals.map(lambda i: _make_loci_array(i))
+        if flatten:
+            result = hl.flatten(loci_arrays)
+            if deduplicate:
+                result = hl.array(hl.set(result))
+            result = hl.sorted(result)
```
ch-kr, line 249: we should add a test checking that results are sorted when flatten is True

gnomad/utils/intervals.py
```diff
@@ -112,3 +120,175 @@ def _add_padding(
...
+        intervals = intervals.annotate(_loci=loci_array).explode("_loci")
+        intervals = intervals.key_by(locus=intervals._loci)
+
+        fields_to_drop = ["_loci"]
+        if not keep_intervals:
+            fields_to_drop.append(interval_field)
+
+        intervals = intervals.drop(*fields_to_drop)
+
+        if deduplicate:
```
ch-kr, line 271: should the other logic branch (deduplicate=False, keep_intervals=True) case also be covered in tests?

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,680 @@
...
+        assert loci == expected_loci
+
+    def test_explode_interval_expression_excludes_both(self) -> None:
+        """Test that both endpoints are excluded when includes_start and includes_end are False."""
+        interval = hl.interval(
+            hl.locus("chr1", 100, "GRCh38"),
+            hl.locus("chr1", 105, "GRCh38"),
+            includes_start=False,
+            includes_end=False,
+        )
```
ch-kr, line 174: this looks like the same interval that is built into the function above. could this be extracted into a fixture?

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,680 @@
...
+            start=hl.locus("chr1", 100, "GRCh38"),
+            end=hl.locus("chr1", 100, "GRCh38"),
+            includes_start=False,
+            includes_end=False,
+        )
+
+        ht = hl.Table.parallelize(
+            [{"interval": interval, "gene": "GENE1"}],
+            hl.tstruct(interval=hl.tinterval(hl.tlocus("GRCh38")), gene=hl.tstr),
+        )
```
ch-kr, line 351: there are a few `hl.Table.parallelize`s in the tests as well, could you update the fixture or create other fixtures to reduce redundancy?

tests/utils/test_interval_utils.py
```diff
@@ -0,0 +1,680 @@
...
+        result = result_ht.collect()
+
+        # Verify that the interval field is not present.
+        assert "interval" not in result_ht.row
+        assert all(not hasattr(row, "interval") for row in result)
+
+        # Verify that other fields are still present.
+        assert all(hasattr(row, "gene") for row in result)
+
+    def test_explode_interval_expression(
```
ch-kr, line 109: the multiple `test_explode_interval_expression*` tests that only change testing whether the start/end are included/excluded can be collapsed into a single`@pytest.mark.parametrized` function that uses a table with a few rows that vary the start/end inclusion. this is also true of the `test_explode_table_*excludes*` tests

ch-kr (changes requested):
a few more comments after revisiting

PR #792, opened 2025-07-22
Remove Azure resources from `gnomad/resources/config.py`, `gnomad/resources/resource_utils.py`, `tests/resources/test_resource_utils.py`, and `docs/resource_sources.rst`

ch-kr:
This PR removes support for [Azure Open Datasets-based gnomAD public resources](https://learn.microsoft.com/en-us/azure/open-datasets/dataset-gnomad). Access to gnomAD data in Azure Open Datasets will be deprecated in August.

Breaking changes in `gnomad/resources/config.py`:
- Removed `AZURE_OPEN_DATASETS` constant
- Removed hdinsight from in `default_resource_sources_by_provider` within `get_default_public_resource_source` function

### DIFF
--- a/tests/resources/test_resource_utils.py
+++ b/tests/resources/test_resource_utils.py
@@ -167,9 +167,7 @@ class TestDefaultPublicResourceSource:
         "cloud_spark_provider,expected_default_source",
         [
             ("dataproc", GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS),
-            ("hdinsight", GnomadPublicResourceSource.REGISTRY_OF_OPEN_DATA_ON_AWS),
             ("unknown", GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS),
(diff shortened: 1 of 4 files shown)


tests/resources/test_resource_utils.py
```diff
@@ -172,9 +167,7 @@ def test_get_default_source_from_environment(
         "cloud_spark_provider,expected_default_source",
         [
             ("dataproc", GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS),
-            ("hdinsight", GnomadPublicResourceSource.AZURE_OPEN_DATASETS),
             ("unknown", GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS),
-            (None, GnomadPublicResourceSource.GOOGLE_CLOUD_PUBLIC_DATASETS),
```
mike-w-wilson, line 177: Just curious why the None is being dropped too since it looks like we would default to google in this case?

ch-kr: oops, I thought this was something Cursor added -- will add back

mike-w-wilson (approved):
:shipit:

PR #793, opened 2025-07-24
Add the option to subtract histograms to the merge_histogram function

mike-w-wilson:
This adds the option to subtract histograms in the merge_histograms function as well as tests for the entire function

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -1633,7 +1633,7 @@ def merge_histograms(
         merged_n_smaller = _sum_or_diff_values(i.n_smaller, j.n_smaller, operation)
         merged_n_larger = _sum_or_diff_values(i.n_larger, j.n_larger, operation)
 
-        # If specified, set negative values in histogram to zero after merge operation, otherwise
+        # If specified, set negative values in histogram to zero after merge operation,
         # otherwise raise an error.
         negative_value_error_msg = (
             "Negative values found in merged histogram. Review data or set"


gnomad/utils/annotations.py
```diff
@@ -1566,34 +1585,80 @@ def merge_freq_arrays(
...
-    This function merges a list of histogram annotations by summing the arrays
-    in an element-wise fashion. It keeps one 'bin_edge' annotation but merges the
-    'bin_freq', 'n_smaller', and 'n_larger' annotations by summing them.
+    This function merges a list of histogram annotations by summing or subtracting the
+    arrays in an element-wise fashion. It keeps one 'bin_edge' annotation but merges the
+    'bin_freq', 'n_smaller', and 'n_larger' annotations by summing or subtracting them.
 
     .. note::
 
         Bin edges are assumed to be the same for all histograms.
```
ch-kr:
```suggestion
        Bin edges are assumed to be the same for all histograms, and lengths of bin edges and frequencies are assumed to be the same for all histograms.
```

mike-w-wilson:
Back to you, @ch-kr -- I refactored again. I agree on putting the negative value handling into a private function but I found it a bit confusing that we were `_handle_negative_values_for_diff` even on sum operations so I pulled it back out. I know the logic made sense but I find it easier to follow this way.

gnomad/utils/annotations.py
```diff
@@ -1566,34 +1585,81 @@ def merge_freq_arrays(
...
+            hl.or_else(i.bin_freq, hl.literal([hl.missing(hl.tint)])),
+            hl.or_else(j.bin_freq, hl.literal([hl.missing(hl.tint)])),
+            fill_missing=True,
+        ).map(lambda x: _sum_or_diff_values(x[0], x[1], operation))
+
+        merged_n_smaller = _sum_or_diff_values(i.n_smaller, j.n_smaller, operation)
+        merged_n_larger = _sum_or_diff_values(i.n_larger, j.n_larger, operation)
+
+        # Handle negative values if performing diff operation or if user wants to
+        # handle negatives, otherwise raise an error.
```
mike-w-wilson:
```suggestion
        # If specified, set negative values in histogram to zero after merge operation, otherwise 
        # otherwise raise an error.
```

ch-kr (approved):

gnomad/utils/annotations.py
```diff
@@ -1566,34 +1585,81 @@ def merge_freq_arrays(
...
+        merged_bin_freq = hl.zip(
+            hl.or_else(i.bin_freq, hl.literal([hl.missing(hl.tint)])),
+            hl.or_else(j.bin_freq, hl.literal([hl.missing(hl.tint)])),
+            fill_missing=True,
+        ).map(lambda x: _sum_or_diff_values(x[0], x[1], operation))
+
+        merged_n_smaller = _sum_or_diff_values(i.n_smaller, j.n_smaller, operation)
+        merged_n_larger = _sum_or_diff_values(i.n_larger, j.n_larger, operation)
+
+        # If specified, set negative values in histogram to zero after merge operation, otherwise
```
mike-w-wilson, line 1636:
```suggestion
        # If specified, set negative values in histogram to zero after merge operation, 
```

PR #794, opened 2025-07-25
Add LEN to compute_stats_per_ref_site after hail updates

mike-w-wilson:
hail updated to use `LEN` in their code 9 months ago -- https://github.com/hail-is/hail/pull/14681/files and we only grab `END` in this function right now https://github.com/broadinstitute/gnomad_methods/blob/daf45c05d3360228815303877a9ad5c981474b07/gnomad/utils/sparse_mt.py#L1220 this adds `LEN` to the set to let it work with hail. Keeping END so its backwards compatible.


ch-kr (approved):
LGTM!

PR #799, opened 2025-08-06
add cdf params to globals

klaricch:
edit default_compute_info to store cdf params in globals


gnomad/utils/sparse_mt.py
```diff
@@ -751,6 +751,8 @@ def default_compute_info(
     if n_partitions is not None:
         info_ht = info_ht.naive_coalesce(n_partitions)
 
+    info_ht = info_ht.annotate_globals(retain_cdfs=retain_cdfs, cdf_k=cdf_k)
```
ch-kr: should this only store cdf_k if retain_cdfs is True?

ch-kr (approved):

PR #800, opened 2025-08-21
Update annotate_downsampling to work without gen_anc_expr

ch-kr:
The current code within `annotate_downsamplings` fails if `gen_anc_expr` isn't specified, despite this being an optional argument.
```
TypeError: annotate_globals: keyword argument 'ds_gen_anc_counts': expected expression of type any, found NoneType: None
```

This PR fixes this issue and also adds tests for this function

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -1918,7 +1918,7 @@ def annotate_downsamplings(
         t = t.annotate(downsampling=ht[t.s])
 
     t = t.annotate_globals(downsamplings=downsamplings)
-    
+
     if gen_anc_counts:
         t = t.annotate_globals(ds_gen_anc_counts=gen_anc_counts)
 
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -1917,12 +1917,13 @@ def annotate_downsamplings(
     else:
         t = t.annotate(downsampling=ht[t.s])
 
-    t = t.annotate_globals(
-        downsamplings=downsamplings,
-        ds_gen_anc_counts=gen_anc_counts,
-    )
+    if gen_anc_counts:
+        return t.annotate_globals(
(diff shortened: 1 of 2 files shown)


gnomad/utils/annotations.py
```diff
@@ -1917,12 +1917,13 @@ def annotate_downsamplings(
...
-        ds_gen_anc_counts=gen_anc_counts,
-    )
+    if gen_anc_counts:
+        return t.annotate_globals(
+            downsamplings=downsamplings,
+            ds_gen_anc_counts=gen_anc_counts,
+        )
 
-    return t
+    return t.annotate_globals(downsamplings=downsamplings)
```
mike-w-wilson:
```suggestion
    return t
```

gnomad/utils/annotations.py
```diff
@@ -1917,12 +1917,13 @@ def annotate_downsamplings(
...
 
-    t = t.annotate_globals(
-        downsamplings=downsamplings,
-        ds_gen_anc_counts=gen_anc_counts,
-    )
+    if gen_anc_counts:
+        return t.annotate_globals(
+            downsamplings=downsamplings,
+            ds_gen_anc_counts=gen_anc_counts,
+        )
```
mike-w-wilson: This is strictly personal preference and obviously makes no difference in what happens but if something exists in both the if and else which is essentially what this is, I typically move it out just for maintainability/readability. This one is super minor but also results in less diff. Feel free to ignore if you like the other way

mike-w-wilson:
```suggestion
    t = t.annotate_globals(downsamplings=downsamplings)
    
    if gen_anc_counts:
        t = t.annotate_globals(ds_gen_anc_counts=gen_anc_counts)
```

mike-w-wilson (approved):
A minor comment but otherwise looks good -- thanks for fixing and adding the tests!

PR #806, opened 2025-09-24
Add `include_version` parameter to `import_gencode` and `remove_y_par` to `add_gencode_transcript_annotations`

jkgoodrich:
### Summary
This PR updates the GENCODE import and transcript annotation functionality to better handle transcript ID versions and PAR region duplicates.

#### 1. Update `import_gencode` to keep full transcript ID including version if requested
- When `include_version=True`, the function now properly preserves the full `gene_id` and `transcript_id` including version numbers in separate fields (`gene_id_version` and `transcript_id_version`)
- This allows downstream functions to access both the versioned and non-versioned IDs as needed

#### 2. Fix `add_gencode_transcript_annotations` to include start and end position
- Updated the default `annotations` parameter in `add_gencode_transcript_annotations` to include `start_position` and `end_position` by default
- The function now properly calculates and includes transcript start and end positions from the GENCODE interval data
- Updated documentation to reflect that these annotations are included by default

#### 3. Add `remove_y_par` parameter to handle PAR region transcript ID duplicates
- Added `remove_y_par` parameter (default: `True`) to `add_gencode_transcript_annotations`
- This parameter filters out Y chromosome PAR (pseudoautosomal region) features to prevent duplicate transcript IDs
- PAR regions can have identical transcript IDs on both chrX and chrY, which can cause issues in constraint calculations
- The filtering uses the `transcript_id_version` field to identify Y_PAR transcripts (those ending with "Y_PAR")
- Updated function documentation to explain the PAR region handling and requirements

### Technical Details
- The `remove_y_par` functionality requires that the input GENCODE table includes the `transcript_id_version` field, which is available when `import_gencode` is called with `include_version=True`
- The PAR region filtering prevents duplicate transcript annotations that could skew constraint calculations
- Start and end position annotations are computed from the GENCODE interval data and provide essential transcript boundary information

### Impact
- Improves data quality by preventing PAR region duplicates in constraint calculations
- Provides more complete transcript annotation data including positional information
- Maintains backward compatibility while adding new functionality
- Enables better handling of GENCODE data with version information

Screenshot 2025-09-24 at 1 34 56 PM

### DIFF
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -715,19 +715,31 @@ def import_sites_vcf(**kwargs) -> hl.Table:
     return hl.import_vcf(**kwargs).rows()
 
 
-def import_gencode(gtf_path: str, **kwargs) -> hl.Table:
+def import_gencode(
+    gtf_path: str,
+    include_version: bool = False,
+    **kwargs,
+) -> hl.Table:
     """
     Import GENCODE annotations GTF file as a Hail Table.
 
     :param gtf_path: Path to GENCODE GTF file.
+    :param include_version: Whether to include additional fields with the full gene_id
+        and transcript_id including the version number.
     :return: Table with GENCODE annotation information.
     """
     ht = hl.experimental.import_gtf(gtf_path, **kwargs)
 
     # Only get gene and transcript stable IDs (without version numbers if they
     # exist), early versions of GENCODE have no version numbers but later ones do.
-    ht = ht.annotate(
-        gene_id=ht.gene_id.split("\\.")[0],
-        transcript_id=ht.transcript_id.split("\\.")[0],
-    )
+    ann_expr = {
+        "gene_id": ht.gene_id.split("\\.")[0],
+        "transcript_id": ht.transcript_id.split("\\.")[0],
+    }
+    if include_version:
+        ann_expr["gene_id_version"] = ht.gene_id
+        ann_expr["transcript_id_version"] = ht.transcript_id
+
+    ht = ht.annotate(**ann_expr)
+
     return ht
--- a/gnomad/resources/resource_utils.py
+++ b/gnomad/resources/resource_utils.py
@@ -724,8 +724,8 @@ def import_gencode(
     Import GENCODE annotations GTF file as a Hail Table.
 
     :param gtf_path: Path to GENCODE GTF file.
-    :param include_version: Whether to include additional fields with the full gene_id
-        and transcript_id including the version number.
+    :param include_version: Whether to include two additional fields that contain the full gene or transcript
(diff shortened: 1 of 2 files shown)


gnomad/resources/resource_utils.py
```diff
@@ -715,19 +715,31 @@ def import_sites_vcf(**kwargs) -> hl.Table:
...
+    gtf_path: str,
+    include_version: bool = False,
+    **kwargs,
+) -> hl.Table:
     """
     Import GENCODE annotations GTF file as a Hail Table.
 
     :param gtf_path: Path to GENCODE GTF file.
+    :param include_version: Whether to include additional fields with the full gene_id
+        and transcript_id including the version number.
```
ch-kr:
```suggestion
    :param include_version: Whether to include two additional fields that contain the full gene or transcript
        ID and the version number (`gene_id_version`, `transcript_id_version`).
```
optional minor rewording to make this clearer

gnomad/utils/constraint.py
```diff
@@ -1727,22 +1736,51 @@ def add_gencode_transcript_annotations(
 
     :param ht: Input Table.
     :param gencode_ht: Table with GENCODE annotations.
-    :param annotations: List of GENCODE annotations to add. Default is ["level", "transcript_type"].
-        Added annotations also become keys for the group by when computing "cds_length" and "num_coding_exons".
+    :param annotations: List of GENCODE annotations to add. Default is
+        ["level", "transcript_type"]. Added annotations also become keys for the group
+        by when computing "cds_length" and "num_coding_exons".
```
ch-kr:
```suggestion
    :param annotations: List of GENCODE annotations to add. Default is
        ["level", "transcript_type", "start_position", "end_position"].
```

gnomad/utils/constraint.py
```diff
@@ -1727,22 +1736,51 @@ def add_gencode_transcript_annotations(
...
+    :param remove_y_par: Whether to remove features for the Y chromosome PAR regions.
+        Default is True because the Y chromosome PAR regions are typically not included
+        in the constraint calculations and both chrX and chrY will have the same
+        'transcript_id' field for these regions. This parameter can only be True if
+        `gencode_ht` includes a 'transcript_id_version' field because Y_PAR is included
+        in the version of the transcript, which has been stripped from the
+        'transcript_id' field.
     :return: Table with transcript annotations from GENCODE added.
     """
+    if remove_y_par and "transcript_id_version" not in gencode_ht.row:
```
ch-kr, line 1753: should this also return `gene_id_version` and `transcript_id_version` if they also exist in `gencode_ht`?

jkgoodrich: yeah, I wasn't sure if that's something that should be passed into `annotations` if the user wants it. I suppose that since I want it, I could add it in the default?

PR #807, opened 2025-09-25
temporary methylation readme

jkgoodrich:
Added here for easy review, It will be added to the GCP public methylation resource directory gs://gcp-public-data--gnomad/resources/grch38/methylation_sites

### DIFF
--- a/gnomad/resources/grch38/README.md
+++ b/gnomad/resources/grch38/README.md
@@ -1,0 +1,49 @@
+# gnomAD Methylation Sites Data
+
+This directory contains methylation site annotations for the GRCh38 reference genome, used in gnomAD constraint calculations and variant annotation pipelines.
+
+## Files Description
+
+### BED Files
+- **methylation.bed**: Methylation site annotations for autosomes (chr1-22)
+- **methylation_chrX.bed**: Methylation site annotations for chromosome X
+
+### Hail Table Files
+- **methylation.ht/**: Hail Table containing methylation annotations for autosomes only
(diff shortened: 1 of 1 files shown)


gnomad/resources/grch38/README.md
```diff
@@ -0,0 +1,49 @@
+# gnomAD Methylation Sites Data
+
+This directory contains methylation site annotations for the GRCh38 reference genome, used in gnomAD constraint calculations and variant annotation pipelines.
```
ch-kr:
```suggestion
This directory contains methylation site annotations for the GRCh38 reference genome used in gnomAD constraint calculations and variant annotation pipelines.
```

gnomad/resources/grch38/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Files Description
+
+### BED Files
+- **methylation.bed**: Methylation site annotations for autosomes (chr1-22)
+- **methylation_chrX.bed**: Methylation site annotations for chromosome X
+
+### Hail Table Files
+- **methylation.ht/**: Hail Table containing methylation annotations for autosomes only
```
ch-kr, line 12: keeping the slash hlps reinforce that HTs are actually directories rather than single files, but they do seem a little distracting (maybe just me?). what do you think about keeping or removing them?

ch-kr (commented):
thank you for adding! just a couple minor comments

gnomad/resources/grch38/README.md
ch-kr: this is minor, and any user looking in the directory with `gsutil ls` will see the slash, so going to resolve this

PR #808, opened 2025-10-02
Add vep115 context resource and init script

jkgoodrich:


init_scripts/vep115-init.sh
```diff
@@ -0,0 +1,114 @@
...
+    gnupg2 \
+    software-properties-common \
+    tabix
+curl -fsSL https://download.docker.com/linux/debian/gpg | sudo apt-key add -
+sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/debian $(lsb_release -cs) stable"
+apt-get update
+apt-get install -y --allow-unauthenticated docker-ce
+
+
+mkdir -p /vep_data/loftee_data
```
ch-kr, line 25: the VEP 105 init script also has this:
```
mkdir -p /vep_data/homo_sapiens
```
is this no longer needed?

init_scripts/vep115-init.sh
```diff
@@ -0,0 +1,114 @@
...
+gsutil -u ${PROJECT} cat gs://${VEP_BUCKET}/loftee-beta/${ASSEMBLY}.tar | tar -xf - -C /vep_data/loftee_data &
+
+docker pull ${VEP_DOCKER_IMAGE}
```
ch-kr, line 28: there are a couple `wait` commands in the 105 init script (one here, and another after 
```
gsutil -u $PROJECT cat gs://gcp-public-data--gnomad/resources/vep/v115/homo_sapiens_merged_vep_115_GRCh38.tar.gz | tar -xzf - -C /vep_data
```
do we still need these?

init_scripts/vep115-init.sh
```diff
@@ -0,0 +1,114 @@
...
+    "--pubmed",
+    "--uniprot",
+    "--mane",
+    "--tsl",
+    "--appris",
+    "--variant_class",
+    "--gene_phenotype",
+    "--mirna",
+    "--allele_number",
+    "--no_stats",
```
ch-kr:
```suggestion
```
there's a [known issue](https://github.com/Ensembl/ensembl-vep?tab=readme-ov-file#known-bugs) with specifying `--hgvs` and `--no_stats`

init_scripts/vep115-init.sh
```diff
@@ -0,0 +1,114 @@
...
...nce_terms:Array[String],context:String,impact:String,regulatory_feature_id:String,variant_allele:String}],seq_region_name:String,start:Int32,strand:Int32,transcript_consequences:Array[Struct{allele_num:Int32,amino_acids:String,appris:String,bam_edit:String,biotype:String,canonical:Int32,ccds:String,cdna_start:Int32,cdna_end:Int32,cds_end:Int32,cds_start:Int32,codons:String,consequence_terms:Array[String],context:String,distance:Int32,domains:Array[Struct{db:String,name:String}],exon:String,flags:String,gene_id:String,gene_pheno:Int32,gene_symbol:String,gene_symbol_source:String,given_ref:String,hgnc_id:String,hgvsc:String,hgvsp:String,hgvs_offset:Int32,impact:String,intron:String,lof:String,lof_filter:String,lof_flags:String,lof_info:String,mane:Array[String],mane_select:String,mane_plus_clinical:String,mirna:Array[String],polyphen_prediction:String,polyphen_score:Float64,protein_end:Int32,protein_id:String,protein_start:Int32,sift_prediction:String,sift_score:Float64,source:String,strand:Int32,swissprot:Array[String],transcript_id:String,trembl:Array[String],tsl:Int32,uniparc:Array[String],uniprot_isoform:Array[String],used_ref:String,variant_allele:String}],variant_class:String}"
```
ch-kr: I have a couple questions about this schema:
- this schema contains `minimsed`, but the docs suggest that field is output when `--minimal` is used. do we always expect the `minimised` field (I see it in the 105 schema as well)?
- similar question about `bam_edit`, `given_ref`, and `used_ref`: these seem to only exist if `--bam` and `--use_transcript_ref` are specified. do we expect that the output should contain these fields even without using those flags?

init_scripts/vep115-init.sh
jkgoodrich: it was never needed. I think for 105 I just copied from a script Nick gave me.

init_scripts/vep115-init.sh
jkgoodrich: No, except that I forgot to remove an &

I removed the &'s which makes things a little bit faster since things are done in parallel. The wait just makes sure it doesn't move onto another step until the things running in the background stop running. I removed them because we were getting silent failures that Wenhan saw

init_scripts/vep115-init.sh
jkgoodrich: huh, odd. Thank you!

init_scripts/vep115-init.sh
jkgoodrich: Thank you. I made a mistake and searched for minimal to remove, but forgot it was minimized in the schema

jkgoodrich: I will remove bam_edit, given_ref, and used_ref. I think I was looking at a few other options and must have kept them in. Will remove

PR #833, opened 2026-04-07
Remove onnx, onnxruntime, skl2onnx from requirements and defer imports

jkgoodrich:
Move onnx/onnxruntime/skl2onnx imports into the functions that use them so the module loads without these packages installed, avoiding the protobuf version conflict with Hail on Dataproc (see #831).


mike-w-wilson (approved):
Looks good -- I just added a note to show up in the docs and also catching the import error to remind users they need to install the packages if they havent already.

PR #834, opened 2026-04-22
Add constants used in generating variant QC annotation HT

ch-kr:
Adds `INFO_FEATURES`, `NON_INFO_FEATURES`, and `TRUTH_DATA` from gnomAD v4 pipeline code: https://github.com/broadinstitute/gnomad_qc/blob/c44f1f76fc5d7a9cd63c46b1826b0d0a376fbe4a/gnomad_qc/v4/annotations/generate_variant_qc_annotations.py#L64

### DIFF
--- a/gnomad/variant_qc/pipeline.py
+++ b/gnomad/variant_qc/pipeline.py
@@ -39,7 +39,7 @@ INFO_FEATURES = [
     "AS_SOR",
     "AS_FS",
 ]
-"""List of features info to be used for variant QC."""
+"""List of features to be used for variant QC that are in the info field."""
 
 NON_INFO_FEATURES = [
     "variant_type",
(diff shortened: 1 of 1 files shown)


gnomad/variant_qc/pipeline.py
```diff
@@ -30,6 +30,29 @@
...
+INFO_FEATURES = [
+    "AS_MQRankSum",
+    "AS_pab_max",
+    "AS_MQ",
+    "AS_QD",
+    "AS_ReadPosRankSum",
+    "AS_SOR",
+    "AS_FS",
+]
+"""List of features info to be used for variant QC."""
```
mike-w-wilson:
```suggestion
"""List of features to be used for variant QC that are in the info field."""
```

mike-w-wilson (approved):
one small thing then LGTM

PR #837, opened 2026-05-19
Fix reduced parent hist lowering by forcing groups to be leafs

mike-w-wilson:
Hail couldn't handle this function when applied to hists. This will be passed onto the hail team. This is a work around that should have been a feature to start. Essentially it forces groups to be leafs, like 'raw', if you do not want to use the other groups sum as their value.

### DIFF
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -2489,9 +2489,9 @@ def generate_freq_group_membership_array(
         Default is None, which resolves to `{"downsampling"}`.
     :param force_leaf_groups: Optional list of `freq_meta` dicts to keep
         as leaves under `reduce_to_minimal_groups` (see
-        `find_minimal_strata_groups`). Pass the groups a downstream caller
-        pins non-reducible aggregations to so they stay directly computed
-        rather than reconstructed per row. Default is None.
+        `find_minimal_strata_groups`). Use this for groups that downstream
+    code accesses directly, so they remain explicitly computed instead
+    of being reconstructed per row from child groups
     :param group_label: Value to use for the `"group"` key on every
         constructed `freq_meta` entry. Default is `"adj"`. Set to `"raw"`
         for callers that don't apply any genotype-level filtering (e.g.,
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -3012,8 +3012,20 @@ def agg_by_strata(
                 s_indices_per_target.append(ht.indices_by_group[payload])
                 adj_per_target.append(ht.adj_groups[payload])
             else:  # parent
+                # Reconstruct the parent's sample set as a single flat
+                # index array (OR of its disjoint leaf-children's
+                # per-sample membership) instead of flattening an
+                # array-of-index-arrays. The flat form matches the leaf
+                # path's shape, so aggregators like `hl.agg.hist` lower
+                # correctly; the array-of-arrays flatten does not.
                 s_indices_per_target.append(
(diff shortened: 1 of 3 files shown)


gnomad/utils/annotations.py
```diff
@@ -3012,8 +3039,20 @@ def _per_target_indices_and_adj(target_resolutions):
                 s_indices_per_target.append(ht.indices_by_group[payload])
                 adj_per_target.append(ht.adj_groups[payload])
             else:  # parent
+                # Reconstruct the parent's sample set as a single flat
+                # index array (OR of its disjoint leaf-children's
+                # per-sample membership) instead of flattening an
+                # array-of-index-arrays. The flat form matches the leaf
+                # path's shape, so aggregators like `hl.agg.hist` lower
+                # correctly; the array-of-arrays flatten does not.
```
klaricch:
```suggestion
# Reconstruct the parent's sample set as a single flat
# index array (OR of its disjoint leaf-children's
# per-sample membership) instead of flattening an
# array-of-index-arrays. The flat form matches the IR shape of
# the leaf path, which lets aggregators like `hl.agg.hist` lower
# correctly. The previous approach — hl.flatten(array of per-leaf
# index arrays) — has the same values but a different IR shape
# that Hail cannot lower for these aggregators.
```

gnomad/utils/annotations.py
```diff
@@ -2466,6 +2487,11 @@ def generate_freq_group_membership_array(
     :param non_summable_strata: Strata names that should never be summed
         across their values when `reduce_to_minimal_groups` is True.
         Default is None, which resolves to `{"downsampling"}`.
+    :param force_leaf_groups: Optional list of `freq_meta` dicts to keep
+        as leaves under `reduce_to_minimal_groups` (see
+        `find_minimal_strata_groups`). Pass the groups a downstream caller
+        pins non-reducible aggregations to so they stay directly computed
+        rather than reconstructed per row. Default is None.
```
klaricch:
```suggestion
    :param force_leaf_groups: Optional list of `freq_meta` dicts to keep
        as leaves under `reduce_to_minimal_groups` (see
        `find_minimal_strata_groups`). Use this for groups that downstream
    code accesses directly, so they remain explicitly computed instead
    of being reconstructed per row from child groups
```

gnomad/utils/annotations.py
```diff
@@ -2180,6 +2181,16 @@ def find_minimal_strata_groups(
         `freq_meta`. Used to validate candidate decompositions.
     :param non_summable_strata: Strata names that should never be summed
         across their values. Default is `{"downsampling"}`.
+    :param force_leaf_groups: Optional list of `freq_meta` dicts to retain
```
klaricch, line 2184: maybe this text as an alternative?

:param force_leaf_groups: Optional list of `freq_meta` dicts to retain
    as leaf groups even if they would otherwise be reconstructed from
    other groups under `reduce_to_minimal_groups`. Use this for groups
    that downstream code accesses directly via
    `entry_agg_group_membership`, so they remain directly materialized
    instead of being rebuilt per row from their leaf children. This
    avoids repeated reconstruction overhead for frequently accessed
    groups. Entries are matched by exact dict equality against
    `freq_meta` entries. Default is None (no forced leaves).


the IR change fixed the  non-reducible aggregation issue right?

klaricch (changes requested):

gnomad/utils/annotations.py
mike-w-wilson: Yeah I was hitting it with the hists so this resolves it.

klaricch (approved):
other than the one failed check lgtm

PR #838, opened 2026-06-12
Add ht to repartition for join, coverage_over_x_bins drops gen anc group size from downsamplings

mike-w-wilson:
This expands the utility of the repartion_for_join by allowing you to pass a HT and the desired partitions. Adds the hist coverage_over_x_bins list and drops the group size entries from the downsamplings list as the function itself will add these.


gnomad/utils/annotations.py
```diff
@@ -89,6 +89,8 @@
     },
 }
 
+COVERAGE_OVER_X_BINS = (1, 5, 10, 15, 20, 25, 30, 50, 100)
```
ch-kr, line 92: maybe we should update `compute_coverage_stats` in sparse_mt.py (https://github.com/broadinstitute/gnomad_methods/blob/8adda7d8a40cc8f6325e26690c074efa827df6aa/gnomad/utils/sparse_mt.py#L1457) to use this new constant?

gnomad/utils/file_utils.py
```diff
@@ -197,31 +197,59 @@ def read_list_data(input_file_path: str) -> List[str]:
 
 
 def repartition_for_join(
```
ch-kr, line 196: I just noticed this function doesn't have any associated tests, should we add some now since we're updating this function?

gnomad/utils/file_utils.py
```diff
@@ -197,31 +197,59 @@ def read_list_data(input_file_path: str) -> List[str]:
...
+
+    :param ht: Table, or path to a Table, to use for interval partition
+        calculation. A path is read with `hl.read_table`; an in-memory Table
+        (e.g. a partition-filtered VDS `reference_data.rows()`) is used directly.
+    :param new_partition_percent: Percent of the input Table's partitions to use
+        when `n_partitions` is not given. Value should be greater than 1 so the
+        input has more partitions for the join. Defaults to 1.1. Ignored when
+        `n_partitions` is provided.
+    :param n_partitions: Explicit number of partitions/intervals to calculate.
+        Overrides `new_partition_percent` when set (e.g. to subdivide into a
```
ch-kr:
```suggestion
        Overrides `new_partition_percent` when set (e.g., to subdivide into a
```
nit

gnomad/utils/file_utils.py
```diff
@@ -197,31 +197,59 @@ def read_list_data(input_file_path: str) -> List[str]:
...
+    locus_intervals: bool = False,
+) -> List[hl.utils.Interval]:
     """
-    Calculate new partition intervals using input Table.
+    Calculate new partition intervals from a Table for co-partitioned joins.
 
-    Reading in all Tables using the same partition intervals (via
-    `_intervals`) before they are joined makes the joins much more efficient.
-    For more information, see:
+    Reading all Tables (and VDSes) with the same partition intervals (via
```
ch-kr: I know MTs are not as common now, but should this docstring also cover that data structure? maybe "Reading all Tables, MatrixTables, and VDSes with the same partition intervals (via `_intervals` for Tables and MatrixTables, `hl.vds.read_vds(intervals=...)` for VDS)"?

gnomad/utils/file_utils.py
ch-kr: should this PR also get a changelog label (maybe new feature?)

ch-kr (changes requested):
just a few minor comments + tests

tests/utils/test_file_utils.py
```diff
@@ -0,0 +1,130 @@
...
+        # The count should follow n_partitions (3), not the percent-derived value
+        # (int(10 * 5.0) == 50).
+        assert len(intervals) == 3
+
+    def test_new_partition_percent_default(self, int_keyed_ht):
+        """Test the interval count derived from the default `new_partition_percent`."""
+        intervals = repartition_for_join(int_keyed_ht)
+
+        # The default requests int(ht.n_partitions() * 1.1) == int(10 * 1.1) == 11
+        # partitions; the partitioner may return slightly fewer (here, 10).
```
ch-kr: since `_calculate_new_partitions` isn't guaranteed to return the number requested (`int(ht.n_partitions() * new_partition_percent)`), should this assertion instead be that the `intervals` is at least `n_partitions` and no larger than `int(ht.n_partitions() * new_partition_percent)`?

tests/utils/test_file_utils.py
```diff
@@ -0,0 +1,130 @@
...
+        assert len(intervals) == 10
+
+    def test_new_partition_percent_custom(self, int_keyed_ht):
+        """Test the interval count derived from a custom `new_partition_percent`."""
+        intervals = repartition_for_join(int_keyed_ht, new_partition_percent=2.0)
+
+        # int(ht.n_partitions() * 2.0) == int(10 * 2.0) == 20.
+        assert len(intervals) == 20
```
ch-kr: should this assertion also be a range rather than a single value?

tests/utils/test_file_utils.py
```diff
@@ -0,0 +1,130 @@
...
+class TestRepartitionForJoin:
+    """Test the repartition_for_join function."""
+
+    def test_returns_list_of_intervals(self, int_keyed_ht):
+        """Test that the function returns a non-empty list of Intervals."""
+        intervals = repartition_for_join(int_keyed_ht, n_partitions=5)
+
+        assert isinstance(intervals, list)
+        assert len(intervals) > 0
+        assert all(isinstance(i, Interval) for i in intervals)
```
ch-kr, line 101: should this also test for correctness in addition to type?

tests/utils/test_file_utils.py
```diff
@@ -0,0 +1,130 @@
...
+        """Test that a `new_partition_percent` below 1 logs a warning."""
+        with caplog.at_level(logging.WARNING, logger="gnomad.utils.file_utils"):
+            repartition_for_join(int_keyed_ht, new_partition_percent=0.5)
+
+        assert any(
+            "new_partition_percent value is less than 1" in record.message
+            for record in caplog.records
+        )
+
+    def test_reads_table_from_path(self, int_keyed_ht, tmp_path):
```
ch-kr: maybe this should do a comparison of calling `repartition_for_join` using the path string and `int_keyed_ht` directly?

tests/utils/test_file_utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@pytest.fixture(scope="module")
+def int_keyed_ht():
+    """Return a range Table keyed by `idx` with 10 partitions."""
+    return hl.utils.range_table(100, n_partitions=10)
+
+
+@pytest.fixture(scope="module")
+def locus_keyed_ht():
```
ch-kr, line 32: should these tests also test against a locus and alleles keyed table?

tests/utils/test_file_utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_locus_intervals_requires_locus_key(self, int_keyed_ht):
+        """Test that `locus_intervals=True` needs a locus-keyed Table."""
+        # The docstring requires the input to be keyed by `locus`; an `idx`-keyed
+        # Table has no `locus` field in its interval bounds.
+        with pytest.raises(KeyError):
+            repartition_for_join(int_keyed_ht, n_partitions=5, locus_intervals=True)
+
+    def test_intervals_usable_for_copartitioned_read(self, int_keyed_ht, tmp_path):
+        """Test that the returned intervals work as co-partition read intervals.
```
ch-kr, line 211: should this also test that the intervals returned can be used by a VDS or MT? or is creating a fake VDS too much of a pain

ch-kr (commented):
a few comments on the tests

tests/utils/test_file_utils.py
mike-w-wilson: Add one below. If you think one is good, we can drop this one.

tests/utils/test_file_utils.py
mike-w-wilson: Added below

tests/utils/test_file_utils.py
mike-w-wilson: Added a synthetic VDS and test MT and VDS below

gnomad/utils/annotations.py
```diff
@@ -1975,6 +1980,14 @@ def annotate_downsamplings(
         gen_anc_counts = ht.aggregate(
             hl.agg.filter(hl.is_defined(ht.gen_anc), hl.agg.counter(ht.gen_anc))
         )
+        # Optionally restrict the per-group downsamplings to the requested genetic
+        # ancestry groups (default: every present group). Groups left out still get
+        # a per-group index but no per-group downsampling; the global downsamplings
+        # continue to cover all samples.
+        if gen_ancs_to_downsample is not None:
+            gen_anc_counts = {
```
ch-kr, line 2044: resetting gen_anc_counts here means that if a requested downsampling is smaller than the sum of the specific genetic ancestry group sizes, then it'll get excluded (e.g., if `gen_ancs_to_downsample` is just `['mid']`, then a downsampling size of 1000 will be removed even if that's a valid downsampling size for the full dataset). you could rename `gen_anc_counts` above on 1980 to `total_gen_anc_counts` and reuse it on 1991 to get around this

tests/utils/test_annotations.py
```diff
@@ -1603,6 +1603,57 @@ def test_annotate_downsamplings_table_with_gen_anc(self, sample_table):
...
+            rows, hl.tstruct(s=hl.tstr, gen_anc=hl.tstr)
+        ).key_by("s")
+
+    def test_gen_anc_group_sizes_added_dynamically(self, gen_anc_sized_table):
+        """Test that per-group sizes are appended to the downsamplings list.
+
+        The input list holds only a "global" size (6); each genetic ancestry
+        group's sample count (AFR=5, EUR=4, SAS=3) is added dynamically. This is
+        why the hardcoded per-group sizes were removed from the DOWNSAMPLINGS
+        constant: passing `gen_anc_expr` regenerates them from the data.
```
ch-kr: should this say this is why per-group sizes don't need to be included in the `DOWNSAMPLINGS` constant instead?

tests/utils/test_annotations.py
```diff
@@ -1603,6 +1603,57 @@ def test_annotate_downsamplings_table_with_gen_anc(self, sample_table):
...
+
+        # None of the group sizes (3, 4, 5) were in the input list yet all appear.
+        assert hl.eval(result.downsamplings) == [3, 4, 5, 6]
+        assert hl.eval(result.ds_gen_anc_counts) == {"AFR": 5, "EUR": 4, "SAS": 3}
+
+    def test_gen_ancs_to_downsample_restricts_added_groups(self, gen_anc_sized_table):
+        """Test that `gen_ancs_to_downsample` limits which group sizes are added."""
+        result = annotate_downsamplings(
+            gen_anc_sized_table,
+            [6],
```
ch-kr, line 1639: we should also add a downsampling count >= 9 here to make sure that valid downsamplings larger than sum(gen_ancs_to_downsample) are returned

gnomad/utils/annotations.py
```diff
@@ -2575,9 +2590,15 @@ def generate_freq_group_membership_array(
             downsampling_idx = "global_idx"
             if ds is not None:
                 if gen_anc is not None and gen_anc != "global":
-                    # Don't include genetic ancestry group downsamplings where the downsampling is
-                    # larger than the number of samples in the genetic ancestry group.
-                    if ds > ds_gen_anc_counts[gen_anc]:
+                    # Skip a genetic ancestry group downsampling when the group was
+                    # not downsampled at all (absent from ds_gen_anc_counts; see
+                    # annotate_downsamplings' gen_ancs_to_downsample) or when the
```
ch-kr:
```suggestion
                    # `annotate_downsamplings`' gen_ancs_to_downsample) or when the
```

gnomad/utils/annotations.py
```diff
@@ -2575,9 +2590,15 @@ def generate_freq_group_membership_array(
             downsampling_idx = "global_idx"
             if ds is not None:
                 if gen_anc is not None and gen_anc != "global":
-                    # Don't include genetic ancestry group downsamplings where the downsampling is
-                    # larger than the number of samples in the genetic ancestry group.
-                    if ds > ds_gen_anc_counts[gen_anc]:
+                    # Skip a genetic ancestry group downsampling when the group was
```
ch-kr, line 2656: should we add tests for this function now too? 😬

gnomad/utils/annotations.py
ch-kr, line 1779: should we add a note that this function will always return downsamplings for all gen anc groups / doesn't support `gen_ancs_to_downsample`?

ch-kr (changes requested):
one comment on `gen_anc_counts` and a few minor comments on tests

gnomad/utils/annotations.py
mike-w-wilson: Added support for this since its the mega function instead of adding a note.

gnomad/utils/annotations.py
mike-w-wilson: Thank you for catching and your suggestion, implemented above.

gnomad/utils/annotations.py
mike-w-wilson: 😬

mike-w-wilson: Added.

gnomad/utils/annotations.py
```diff
@@ -1702,6 +1702,7 @@ def annotate_freq(
     annotate_mt: bool = True,
     reduce_to_minimal_groups: bool = False,
     non_summable_strata: Optional[Set[str]] = None,
+    gen_ancs_to_downsample: Optional[List[str]] = None,
```
ch-kr, line 1705: does this function also need tests? 🙈

ch-kr (commented):
one more comment about tests

gnomad/utils/annotations.py
mike-w-wilson: Added!

gnomad/utils/annotations.py
```diff
@@ -1839,9 +1839,11 @@ def annotate_freq(
         Default is None, which resolves to `{"downsampling"}`.
     :param gen_ancs_to_downsample: Optional list of genetic ancestry groups to
         generate per-group downsamplings for. When None (default), every genetic
-        ancestry group present is downsampled. Only used when `gen_anc_expr` is
-        supplied and `annotate_freq` generates the downsamplings (i.e. when
-        `downsamplings` is set and `downsampling_expr` is not).
+        ancestry group present is downsampled; an empty list generates no
```
ch-kr, line 1842: this note should also get added to the docstring for `annotate_downsamplings`

tests/utils/test_annotations.py
```diff
@@ -2871,6 +2871,381 @@ def test_unaccounted_index_raises(self):
...
+            entries,
+            hl.tstruct(
+                locus=hl.tlocus("GRCh38"),
+                alleles=hl.tarray(hl.tstr),
+                s=hl.tstr,
+                GT=hl.tcall,
+                adj=hl.tbool,
+            ),
+        ).to_matrix_table(row_key=["locus", "alleles"], col_key=["s"])
+        mt = mt.annotate_cols(gen_anc=sample_table[mt.s].gen_anc)
```
ch-kr: this seems like it's the same MT as the one defined in `test_gen_ancs_to_downsample_restricts_downsampling_groups`. maybe this should also be a fixture?

in general, is redundancy in our testing code an issue?

ch-kr (commented):
so close! one docstring suggestion and a question about test code redundancy

gnomad/utils/annotations.py
```diff
@@ -1941,16 +1976,33 @@ def annotate_downsamplings(
...
+        deduplicated, sorted, or filtered).
     :param gen_anc_expr: Optional expression for genetic ancestry group. When provided, genetic ancestry group
         downsamplings will be computed for each genetic ancestry group.
+    :param gen_ancs_to_downsample: Optional list of genetic ancestry groups to
+        generate per-group downsamplings for. When None (default), every genetic
+        ancestry group present is downsampled; an empty list generates no
+        per-group downsamplings (only the global downsamplings are used). Only
+        used when `gen_anc_expr` is supplied and `annotate_freq` generates the
+        downsamplings (i.e. when `downsamplings` is set and `downsampling_expr`
+        is not).
```
ch-kr:
```suggestion
        per-group downsamplings (only the global downsamplings are used).
        Requires `gen_anc_expr`.
```

ch-kr (approved):
one docstring update then LGTM!

PR #841, opened 2026-06-24
Update hadoop to hailtop after hail's hadoop deprecation

mike-w-wilson:
Hail deprecated hadoop functions in hail 0.2.137. This updates to use hailtop which does not need to initialize hail and is in general a drop in replacement as noted by Chris here: https://github.com/hail-is/hail/pull/15182. The gz exception is covered here as well as are hidden files.

### DIFF
--- a/gnomad/assessment/parse_validity_logs.py
+++ b/gnomad/assessment/parse_validity_logs.py
@@ -2,7 +2,7 @@
 
 import re
 
-import hail as hl
+import hailtop.fs as hfs
 
 
 def parse_log_file(log_file):
@@ -38,7 +38,7 @@ def parse_log_file(log_file):
         "load_gnomad_data": "load gnomAD data",
     }
 
-    with hl.hadoop_open(log_file, "r") as f:
+    with hfs.open(log_file, "r") as f:
         current_message = ""
         current_table = []
         current_metadata = None  # Stores log metadata before a table appears.
@@ -288,5 +288,5 @@ def generate_html_report(parsed_logs, output_file):
 
     html_template += "</table></body></html>"
 
-    with hl.hadoop_open(output_file, "w") as f:
+    with hfs.open(output_file, "w") as f:
         f.write(html_template)
--- a/gnomad/utils/file_utils.py
+++ b/gnomad/utils/file_utils.py
@@ -9,6 +9,7 @@ import uuid
 from typing import List, Optional, Tuple, Union
 
 import hail as hl
+import hailtop.fs as hfs
 
 from gnomad.resources.resource_utils import DataException
 
@@ -36,14 +37,8 @@ def file_exists(fname: str) -> bool:
     else:
         paths = [fname]
 
-    if fname.startswith("gs://"):
-        exists_func = hl.hadoop_exists
-    else:
-        exists_func = os.path.isfile
-
-    exists = all([exists_func(p) for p in paths])
-
-    return exists
+    # hfs.exists handles local, gs://, and other cloud schemes uniformly.
+    return all(hfs.exists(p) for p in paths)
 
 
 def check_file_exists_raise_error(
@@ -177,7 +172,7 @@ def read_list_data(input_file_path: str) -> List[str]:
     :return: List of lines
     """
     if input_file_path.startswith("gs://"):
-        hl.hadoop_copy(input_file_path, "file:///" + input_file_path.split("/")[-1])
+        hfs.copy(input_file_path, "file:///" + input_file_path.split("/")[-1])
         f = (
             gzip.open("/" + os.path.basename(input_file_path), encoding="utf-8")
             if input_file_path.endswith("gz")
--- a/gnomad/utils/plotting.py
+++ b/gnomad/utils/plotting.py
@@ -353,9 +353,12 @@ def _ls(path: str) -> List[Dict]:
     """
     List a path via `hailtop.fs`, mirroring the deprecated `hl.hadoop_ls`.
 
-    Unlike `hl.hadoop_ls`, `hfs.ls` lists hidden files (e.g. the local
-    filesystem's `.crc` checksum files). Hidden entries (basename starting with
-    `.`) are dropped so downstream `part-*` parsing matches the prior behavior.
+    Hidden entries (basename starting with `.`) are dropped as a defensive
+    measure so downstream `part-*` parsing matches the prior behavior. In
+    practice this chiefly affects the local filesystem's `.crc` checksum files,
+    which the Spark backend's JVM Hadoop FS hides from `hl.hadoop_ls` but
+    `hfs.ls` does not. For GCS paths (and on the Batch backend, where
+    `hl.hadoop_ls` and `hfs.ls` share the same filesystem) this is a no-op.
 
     :param path: Path to list.
     :return: List of legacy `hl.hadoop_ls`-style dicts (with `path`,
--- a/gnomad/utils/plotting.py
+++ b/gnomad/utils/plotting.py
@@ -6,6 +6,7 @@ from typing import Callable, Dict, List, Optional, Union
 
 import bokeh
 import hail as hl
+import hailtop.fs as hfs
 import numpy as np
 import pandas as pd
 from bokeh.layouts import gridplot
@@ -358,7 +359,7 @@ def plot_hail_file_metadata(
     panel_size = 600
     subpanel_size = 150
 
-    files = hl.hadoop_ls(t_path)
+    files = [x.to_legacy_dict() for x in hfs.ls(t_path)]
     rows_file = [x["path"] for x in files if x["path"].endswith("rows")]
     entries_file = [x["path"] for x in files if x["path"].endswith("entries")]
     # cols_file = [x['path'] for x in files if x['path'].endswith('cols')]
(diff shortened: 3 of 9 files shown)


gnomad/utils/file_utils.py
```diff
@@ -36,14 +37,8 @@ def file_exists(fname: str) -> bool:
...
 
-    if fname.startswith("gs://"):
-        exists_func = hl.hadoop_exists
-    else:
-        exists_func = os.path.isfile
-
-    exists = all([exists_func(p) for p in paths])
-
-    return exists
+    # hfs.exists handles local, gs://, and other cloud schemes uniformly.
```
ch-kr, line 40: 🤩

gnomad/assessment/parse_validity_logs.py
```diff
@@ -2,7 +2,7 @@
 
 import re
 
-import hail as hl
+import hailtop.fs as hfs
 
 
 def parse_log_file(log_file):
```
ch-kr, line 8: 😬 should this function have tests?

gnomad/utils/file_utils.py
```diff
@@ -36,14 +37,8 @@ def file_exists(fname: str) -> bool:
...
+    return all(hfs.exists(p) for p in paths)
 
 
 def check_file_exists_raise_error(
```
ch-kr, line 44: should this also have tests?

gnomad/variant_qc/random_forest.py
ch-kr, line 521: I know our policy is to add tests whenever we update code, but I'm not sure this function needs tests. what do you think?

gnomad/utils/vep.py
ch-kr, line 1: 😬 should these functions also get tests?

gnomad/utils/plotting.py
```diff
@@ -373,19 +398,21 @@ def plot_hail_file_metadata(
         logger.warning("No metadata file found. Exiting...")
         return None
 
-    with hl.hadoop_open(metadata_file[0], "rb") as f:
+    # `hfs.open` does not auto-decompress gzipped files (unlike the deprecated
```
ch-kr, line 403: it might be a bit extra, but should we point to https://github.com/hail-is/hail/pull/15182 here?

gnomad/utils/plotting.py
```diff
@@ -346,6 +349,28 @@ def linear_and_log_tabs(plot_func: Callable, **kwargs) -> Tabs:  # noqa: D103
...
+    `hl.hadoop_ls` and `hfs.ls` share the same filesystem) this is a no-op.
+
+    :param path: Path to list.
+    :return: List of legacy `hl.hadoop_ls`-style dicts (with `path`,
+        `size_bytes`, `modification_time`, etc.) for non-hidden entries.
+    """
+    return [
+        entry.to_legacy_dict()
+        for entry in hfs.ls(path)
+        if not os.path.basename(entry.path).startswith(".")
```
ch-kr: I asked claude to verify, and it wrote this test
```
cat > /tmp/crc_test2.py <<'PY'
import os, hail as hl
import hailtop.fs as hfs
hl.init(log="/dev/null", quiet=True)

d = "/tmp/crc_probe2"
os.system(f"rm -rf {d}; mkdir -p {d}")
# Plain files written directly to disk (no Hadoop involvement in creating them)
for name in ["part-0", ".part-0.crc", ".hidden", "_SUCCESS", ".foo.crc"]:
    with open(os.path.join(d, name), "w") as fh:
        fh.write("x")

print("--- raw os.listdir ---", sorted(os.listdir(d)))
print("--- hl.hadoop_ls   ---", sorted(os.path.basename(x['path']) for x in hl.hadoop_ls(d)))
print("--- hfs.ls         ---", sorted(os.path.basename(x.path) for x in hfs.ls(d)))
PY
cd /tmp && HAIL_QUERY_BACKEND=spark PYSPARK_SUBMIT_ARGS="--conf spark.driver.bindAddress=127.0.0.1 --conf spark.driver.host=127.0.0.1 pyspark-shell" /Users/kchao/miniconda3/envs/hail-0.2.137/bin/python /tmp/crc_test2.py 2>&1 | grep -E "^---"
```
which printed
```
--- raw os.listdir --- ['.foo.crc', '.hidden', '.part-0.crc', '_SUCCESS', 'part-0']
--- hl.hadoop_ls   --- ['.hidden', '_SUCCESS', 'part-0']
--- hfs.ls         --- ['.foo.crc', '.hidden', '.part-0.crc', '_SUCCESS', 'part-0']
```
it seems like the old `hadoop_ls` would still list hidden files but not `.crc` files specifically, so maybe this function only needs to filter those files to mirror the old behavior

ch-kr (changes requested):
thank you for updating these! one comment on `hfs.ls` and a few test-related questions

gnomad/assessment/parse_validity_logs.py
mike-w-wilson: Yes

gnomad/utils/plotting.py
mike-w-wilson: Nice, I added a test for this

gnomad/variant_qc/random_forest.py
mike-w-wilson: Agree

gnomad/utils/vep.py
mike-w-wilson: I added some here though they are a bit ugly. I pushed claude to mock things without actually executing vep since that would make test runs way too long.

mike-w-wilson:
Thanks @ch-kr , back to you!

tests/assessment/test_parse_validity_logs.py
```diff
@@ -0,0 +1,182 @@
...
+        assert len(parsed) == 2
+        # First entry keeps its message and captures the table.
+        assert parsed[0][3] == "found issues"
+        assert "| locus |" in parsed[0][4]
+        # The trailing entry has no table.
+        assert parsed[1][3] == "all done"
+        assert parsed[1][4] == ""
+
+
+class TestGenerateHtmlReport:
```
ch-kr, line 139: thank you for adding tests for this also!

gnomad/utils/file_utils.py
```diff
@@ -177,7 +172,7 @@ def read_list_data(input_file_path: str) -> List[str]:
     :return: List of lines
     """
     if input_file_path.startswith("gs://"):
-        hl.hadoop_copy(input_file_path, "file:///" + input_file_path.split("/")[-1])
+        hfs.copy(input_file_path, "file:///" + input_file_path.split("/")[-1])
         f = (
             gzip.open("/" + os.path.basename(input_file_path), encoding="utf-8")
```
ch-kr, line 177: claude found a bug in this line that's slightly out of scope for this PR: `gzip.open` defaults to `mode='rb'` if unspecified, which isn't compatible with `encoding` (raises `ValueError: Argument 'encoding' not supported in binary mode`). it suggests adding `'rt'` to this gzip call (and the one below) to fix this error and adding a test to make sure the `gzip.open`s also function as expected

tests/utils/test_vep.py
```diff
@@ -256,3 +289,151 @@ def test_empty_filter_handling(self):
...
+            ):
+                with patch("gnomad.utils.vep.get_vep_context", return_value=context):
+                    with pytest.raises(AssertionError, match="configuration does"):
+                        vep_or_lookup_vep(
+                            hl.utils.range_table(1),
+                            reference="GRCh38",
+                            vep_config_path="gs://bucket/vep.json",
+                        )
+
+    def test_uses_context_ht_when_all_variants_present(self):
```
ch-kr, line 403: claude seems to think writing a patch test for when some variants have been run through hl.vep and others have been annotated using the VEP context is feasible here
```
    def test_unions_context_lookup_with_revep_of_missing_variants(self):
        """Test the hybrid path: missing variants are re-VEPed and unioned back.

        The context Table covers only a subset of the input keys, so the
        variants absent from it must be VEPed via ``hl.vep`` and unioned with
        the looked-up variants. ``hl.vep`` is mocked to an identity so no VEP
        process is launched; this exercises the non-empty re-VEP/union branch
        that ``test_uses_context_ht_when_all_variants_present`` deliberately
        does not.
        """
        # Context HT holds keys {0, 1}; input has keys {0, 1, 2}, so key 2 is
        # absent from the context and must be re-VEPed.
        context_ht = hl.utils.range_table(2).annotate(vep=hl.struct(found=True))
        context_ht = context_ht.annotate_globals(
            vep_help="CONTEXT_HELP", vep_config="CONTEXT_CONFIG"
        )
        context = _fake_vep_context("105", {"105": context_ht})
        ht = hl.utils.range_table(3)
        with patch("gnomad.utils.vep.get_vep_help", return_value="CONTEXT_HELP"):
            with patch(
                "gnomad.utils.vep.hfs.open", _mock_open_returning("CONTEXT_CONFIG")
            ):
                with patch("gnomad.utils.vep.get_vep_context", return_value=context):
                    with patch(
                        "gnomad.utils.vep.hl.vep", side_effect=lambda t, p: t
                    ) as mock_vep:
                        result = vep_or_lookup_vep(
                            ht,
                            reference="GRCh38",
                            vep_config_path="gs://bucket/vep.json",
                        )

        # Looked-up and re-VEPed variants are unioned back into one Table.
        assert result.count() == 3
        # Only the 2 context-matched variants carry the context annotation; the
        # re-VEPed key 2 keeps a missing vep (identity-mocked hl.vep adds none).
        assert result.aggregate(hl.agg.count_where(result.vep.found)) == 2
        # hl.vep ran exactly once, on the single missing variant, with the config.
        mock_vep.assert_called_once()
        revep_subset, passed_path = mock_vep.call_args[0]
        assert revep_subset.count() == 1
        assert passed_path == "gs://bucket/vep.json"
        # Globals survive the union and reflect the resolved version/help/config.
        assert hl.eval(result.vep_version) == "v105"
        assert hl.eval(result.vep_help) == "CONTEXT_HELP"
        assert hl.eval(result.vep_config) == "CONTEXT_CONFIG"
```

ch-kr (commented):
just a couple more things

gnomad/utils/file_utils.py
mike-w-wilson: Yeah -- it found that in my pass too. I felt similarly its out of scope for this but I was considering having it give a full pass on the repo to check for bugs.

mike-w-wilson:
Back to you @ch-kr

tests/utils/test_vep.py
mike-w-wilson: Nice , added!

gnomad/utils/file_utils.py
ch-kr: adding a note here that we decided to fix this later

ch-kr (approved):
LGTM!

PR #843, opened 2026-08-05
Add a gnomAD & Hail institutional-knowledge reference

jkgoodrich:
## Summary

- Adds `knowledge/`, a shared reference of gnomAD + Hail institutional knowledge maintained by the methods group: bug classes that produce plausible-looking wrong output rather than crashing, design tradeoffs that aren't visible in the code, and facts about the released data that are otherwise buried in the source or in meeting minutes.
- Publishes it to the documentation site alongside the API reference, so it's reachable by people outside the team.
- Adds the repo-level assistant configuration that points at it, plus the team's shared Claude Code skills.

## Changes

- **`knowledge/` (14 files, ~2,900 lines).** A routing README plus topic sub-documents — missingness, Hail idioms, lazy evaluation/materialization, partitioning/shuffles/OOM, hidden APIs, gnomAD-specific, VDS, biology, analysis patterns, Dataproc/operational, release conventions, testing, and working with an AI assistant. The README is an index: a reader loads only the sub-document they need rather than the whole reference.
- **Entry conventions.** Every entry is tagged `[Hail]` and/or `[gnomAD data]`, marked `[internal only]` where it describes artifacts outside the public release, and stamped with the Hail version where behavior is version-dependent. Contribution rules are in the README.
- **Docs wiring.** `docs/build.sh` stages a copy of `knowledge/` into the Sphinx source dir the way `api_reference` is generated; `docs/.gitignore` ignores the copy; `docs/index.rst` gains a toctree entry; `docs/conf.py` sets `myst_heading_anchors = 3` so cross-document `file.md#heading` links resolve. The canonical copy stays at the repo root — readable on GitHub, not coupled to the build.
- **`CLAUDE.md`** (new, deliberately minimal): a pointer to `knowledge/README.md` and its contribution rules, a verbosity rule, and the `CLAUDE.md` vs `CLAUDE.local.md` split. Broader repo context is left to the separate PR that adds it.
- **`.gitignore`**: ignores `CLAUDE.local.md`, which was untracked but not ignored.
- **`.claude/settings.json`** (new): allowlists read-only inspection commands (`git status/log/diff/show`, `ls`, `grep`, `find`, `wc`, `head`, `tail`, `pytest`, `./docs/build.sh`).
- **`.claude/skills/`**: the team's shared commit / review / PR-draft / history-cleanup workflows. `review` is bw2's MIT-licensed skill, vendored with its LICENSE.

## Hail / Compute notes

No pipeline code, no Hail operations, no resource paths, and no schema changes — the only Python touched is a single config line in `docs/conf.py`. Nothing here reads or writes GCS, and there is no compute cost.

Claims *about* Hail and gnomAD data in the reference were verified against Hail 0.2.134 source, the released v4.1 exomes HT globals, and small local jobs. Entries that come from the project's meeting record rather than from re-derivation are labeled `reported · project meeting` so a reader can weigh them differently.

## Testing

- Docs build verified locally against the pinned dependencies CI uses (`sphinx 6.2.1`, `myst-parser 2.0.0`) with `sphinx -W`, exactly as `./docs/build.sh` runs it: build succeeds, all 13 knowledge pages render.
- Every cross-page anchor link in the **generated HTML** was checked against the ids that actually exist on the target pages — none broken.
- No unit tests added: this PR contains no library code. `pytest` and lint jobs are unaffected.
- **The `docs` CI job now covers this content.** A malformed entry fails the build rather than merely rendering oddly on GitHub — that's intentional, and the contribution rules tell contributors to run `./docs/build.sh` before opening a PR.

## Code review

- [x] `/review` run — models: Opus, Sonnet (detection and validation; `agy` not installed, so not in the roster)

Three findings, all agreed by both validators. Two are fixed in this branch:

- **Fixed:** `CLAUDE.md` documented an import block in which 12 of 20 names did not exist anywhere in the repo. The file is now cut to the minimum described above, so the list is gone entirely.
- **Fixed:** the draft-PR skill inferred the reviewer roster from detection prompts alone, assuming detection and validation share a roster; the review skill documents them as independent. It now reads both and reports what actually validated.
- **Not fixed (pre-existing on `main`):** `docs/conf.py` calls `requests.head(...)` at config load with no `try/except` and no `timeout=`, so an offline or network-restricted environment aborts or hangs the docs build instead of degrading to the no-intersphinx path. This branch didn't touch those lines, but the new contribution rules make people run `./docs/build.sh` more often, so it's worth a follow-up PR.

## Reviewer notes

- **This PR puts documentation under CI.** The `docs` job builds with `sphinx -W`, so from here on a malformed entry in `knowledge/` fails the build. Worth agreeing that's the tradeoff you want.
- **`.claude/settings.json` changes permission behavior for everyone** on the repo, not just the author. It's read-only commands and a local build, but it's the one change here with a blast radius beyond documentation — easy to trim or drop.
- **Please sanity-check the `[internal only]` tagging.** Parts of the release-conventions and biology entries came from internal meeting records; only facts about *released* data and public-facing policy were kept, with no names, cost figures, or unpublished work. A second pair of eyes on that boundary is the most valuable review this PR can get.
- **Domain review welcome on the data claims** — annotation-version pinning and its GTEx/pext coupling, LOEUF cutoffs not porting across releases, which subsets exist and why, and the CMRG false-duplication genes (CBS, KCNE1, CRYAA).
- `CLAUDE.md` is intentionally minimal here to avoid colliding with the separate PR that adds the fuller version.

## Merge strategy

**Rebase, not squash.** The five commits are deliberately structured — reference content, docs publishing, contributor config, permissions, and tooling each stand alone and are individually revertible. The `.claude/settings.json` commit in particular is the one a reviewer might want to drop on its own, which squashing would make awkward.

### DIFF
--- a/.claude/skills/commit/SKILL.md
+++ b/.claude/skills/commit/SKILL.md
@@ -1,0 +1,86 @@
+---
+name: commit
+description: Use this when the user wants to commit staged changes. Drafts a commit message matching the repo's style, confirms with the user, then commits.
+allowed-tools: bash, read_file
+---
+
+## Objective
+
+Draft a concise commit message matching the style of recent commits in this repository, run the repo's pre-commit hooks on the staged changes, show the user what will be committed, confirm, then run the commit.
+
+## Workflow
+
+### 1. Check staged changes
+
+Run `git diff --staged --stat` to see what is staged. If nothing is staged:
+- Run `git status` to show what is available
+- Tell the user what is unstaged and ask what to add before proceeding
+
+### 2. Run pre-commit hooks
+
+Always run the repo's pre-commit hooks on the staged files before reviewing, so the commit matches what CI enforces (black, isort, autopep8, pydocstyle, whitespace fixers, etc.). This runs even if the user has not installed the git hook.
+
+Skip this step only if the repo has no `.pre-commit-config.yaml`.
+
+Capture the staged file list first (only run hooks on staged files — never all files, to avoid reformatting unrelated code):
+```bash
+git diff --staged --name-only
+```
+
+Run the hooks scoped to those files:
+```bash
+pre-commit run --files <staged files>
+```
+- If `pre-commit` is not on PATH, try `python -m pre_commit run --files <staged files>`.
+- If pre-commit is unavailable entirely, fall back to running the tools listed in `.pre-commit-config.yaml` directly on the staged files, using the versions pinned in `requirements-dev.txt` / the config's `rev:` (a throwaway venv is fine). For this repo that means: `black`, `isort --profile black --filter-files`, `autopep8 --exit-code --in-place`, `pydocstyle`. autopep8 needs `lib2to3`, so run it under Python ≤ 3.12, not 3.13.
+- If neither pre-commit nor the individual tools can be run, tell the user the hooks could not run and why, and ask how to proceed — do NOT silently commit unformatted code.
+
+Auto-fixing hooks (black, isort, autopep8, end-of-file-fixer, trailing-whitespace) modify the working tree, which unstages those edits. Re-stage the same paths and re-run until the hooks pass cleanly:
+```bash
+git add -- <staged files>
+```
+If a non-autofixable hook fails (e.g. pydocstyle, check-yaml), stop and surface the failure to the user rather than committing.
+
+### 3. Review the diff
+
+Run `git diff --staged` and read the full diff. Look for anything that shouldn't be there: debug prints, commented-out code, unintended files, or anything that looks like a secret. If something looks wrong, stop and tell the user before proceeding.
+
+### 4. Read recent commit style
+
+Run `git log --oneline -10` to understand the repo's commit message convention (capitalization, prefix style, length, etc.).
+
+### 5. Draft a commit message
+
+Match the observed style exactly. General rules unless the repo overrides them:
+- Short imperative phrase, ≤ 72 characters
+- No conventional commit prefixes (`feat:`, `fix:`, etc.) unless the repo uses them
+- No trailing period
+- Add a blank line and a body paragraph only when the change genuinely needs more context than the subject line can carry
+
+### 6. Confirm with the user
+
+Show:
+- The list of staged files (`git diff --staged --name-only`)
+- The draft commit message
+
+Ask the user to confirm, request edits, or abort. Do not commit until they confirm.
+
+### 7. Commit
+
+Run:
+```bash
+git commit -m "$(cat <<'EOF'
+<message here>
+
+Assisted-by: ClaudeCode:<model-id>
+EOF
+)"
+```
+
+`<model-id>` is the model **you are actually running as** — e.g.
+`claude-opus-5`, `claude-sonnet-5`. Do not copy a model name out of this
+file or out of an earlier commit; the trailer records which model
+assisted, so a stale name makes it useless. If the repo's existing
+trailers use a different format, match the repo.
+
+Do NOT push unless the user explicitly asks.
--- a/.claude/skills/draft-pull-request-description/SKILL.md
+++ b/.claude/skills/draft-pull-request-description/SKILL.md
@@ -1,0 +1,107 @@
+---
+name: draft-pull-request-description
+description: Use this whenever the user wants to open a pull request (PR), draft a PR, or summarize their branch for merging.
+allowed-tools: bash, read_file
+---
+
+## Objective
+
+You are a Staff Bioinformatics Engineer preparing a comprehensive, professional, and concise Pull Request description for a gnomAD pipeline codebase. Your goal is to produce a description thorough enough that the reviewer can understand:
+
+- The intent and motivation for the change
+- The specific pipeline steps or resources affected (sample QC, variant QC, frequency annotation, VDS/HT/MT operations, Dataproc cluster config, etc.)
+- The design/architecture chosen and why (especially if Hail-specific tradeoffs were involved)
+- The exact changes and any new output paths or resource names
+- What tests were added or changed (check `gnomad_qc/tests/` and `.github/workflows/ci.yml`)
+- Anything requiring specific reviewer attention (e.g. breaking changes to resource paths, schema changes to HTs, large compute cost implications)
+
+## Workflow
+
+### 1. Determine the base branch
+
+Default to `main`. Ask the user only if they indicate this is a stacked branch.
+
+### 2. Analyze the diff
+
+Run `git diff main...HEAD` to understand the full set of changes. Then read the changed files in full — especially any new or modified Python scripts, resource definitions, or config changes.
+
+### 3. Identify gnomAD-specific context
+
+Check whether the diff touches any of the following and note it in the description:
+
+- **Hail operations**: new/modified VDS, MT, HT read/write paths, aggregations, or annotations
+- **Dataproc / cluster config**: cluster size, autoscaling policy, init scripts, `hailctl dataproc submit` invocations
+- **Resource paths**: new or renamed GCS paths (especially under `gs://gnomad*/`)
+- **Pipeline step flags**: new CLI arguments, `--overwrite`, step-gating logic
+- **Privacy or release filters**: anything touching allele frequency output, sample suppression, or release HTs
+- **Test coverage**: added/changed tests under `gnomad_qc/tests/`
+
+### 4. Check for a completed code review
+
+Look for a completion marker written by the `/review` skill:
+
+```bash
+ls .code_review/*CODE_REVIEW_COMPLETED 2>/dev/null | sort | tail -1
+```
+
+If a marker exists, take its timestamp prefix (format `YYYYMMDD_HHMMSS`) and identify which models were in the roster by listing prompt files that share that prefix:
+
+```bash
+ls .code_review/YYYYMMDD_HHMMSS_PROMPT__*__*.md 2>/dev/null        # detection
+ls .code_review/YYYYMMDD_HHMMSS_PROMPT__*_validate.md 2>/dev/null  # validation
+```
+
+The suffix before `.md` is the model name (`opus`, `sonnet`, `agy`).
+
+**Check both lists — the two rosters can differ.** A reviewer that completes
+detection can still drop out of validation (an `agy` run that times out at the
+validation call, a model whose CLI invocation fails partway), and the review
+skill treats the two rosters as decided independently. Report the models that
+actually *validated*, since that is what the confidence labels rest on. If the
+two lists differ, say so rather than printing one merged roster — e.g.
+"reviewed by opus, sonnet, agy; validated by opus, sonnet".
+
+### 5. Draft the PR description
+
+Use the following template — fill every section; do not leave placeholders:
+
+```markdown
+## Summary
+
+<!-- 1-3 bullet points on what this PR does and why -->
+
+## Changes
+
+<!-- Bullet list of specific changes: functions added/modified, new CLI flags, resource paths changed, etc. -->
+
+## Hail / Compute notes
+
+<!-- Any cluster config, VDS/HT schema changes, estimated cost or runtime if significant. Write "None" if not applicable. -->
+
+## Testing
+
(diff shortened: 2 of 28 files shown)


knowledge/README.md
```diff
@@ -0,0 +1,234 @@
...
+| [VDS](vds.md) | you subset or densify a VDS, or you're choosing between Table / MatrixTable / VDS |
+| [Biology](biology.md) | reference builds, multiallelics, sex chromosomes, trios, AC/AF null semantics, or what QUAL and LOFTEE do and don't tell you |
+| [Release conventions and data policy](dataset-conventions.md) | you wonder why the data is shaped the way it is — annotation versions, subsets, constraint cutoffs across releases, labels, requester-pays |
+| [Codebase and analysis patterns](analysis-patterns.md) | you pick a starting table, a VEP transcript, a bucket scheme, or a per-stratum aggregation |
+| [Dataproc / operational](dataproc-operational.md) | you run, size, package, or post-mortem a job on Dataproc or QoB, or you're wondering why storage costs what it does |
+| [Testing](testing.md) | after any refactor of the join graph or aggregation semantics, and before you pay for a cluster run |
+| [Working with an AI assistant](working-with-ai.md) | you're setting up `CLAUDE.md` / `CLAUDE.local.md`, or wondering what's burning tokens and cluster time |
```
ch-kr, line 39: we should break this down into two (or more, if there are other categories) tables (one for Hail, one for gnomAD) to further stratify

knowledge/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+
+Entries are tagged **`[Hail]`** (universal Hail behavior) and/or
+**`[gnomAD data]`** (dataset- or schema-specific). Entries tagged
+**`[internal only]`** describe data or artifacts available only to the
+internal production team — unreleasable samples, the pre-release VDS,
+internal QC tables. Everything untagged that way applies to anyone
+working with released gnomAD data.
```
ch-kr, line 47: maybe we should further stratify internal only to split production (QC) work and research work, since some research projects require VDS access but don't need to reference QC outputs

knowledge/README.md
ch-kr: we could split the table based on the tags below

knowledge/README.md
```diff
@@ -0,0 +1,234 @@
...
+conventions, and design-choice tradeoffs — the kind of institutional
+knowledge that usually only gets learned by running into the problem
+(or being told by someone who already did).
+
+**When to consult it:** any time you're touching gnomAD annotation
+tables, writing Hail aggregations, filtering by QC status, making
+assumptions about coverage between HTs, or picking between two
+reasonable-sounding analysis approaches. See the [When to consult this
+](#when-to-consult-this) section at the bottom for the concrete trigger
+list.
```
ch-kr, line 64: move to top

knowledge/README.md
```diff
@@ -0,0 +1,234 @@
...
+- **Symptom** for silent-fail items ("aggregation totals < input
+  rowcount", etc.), OR **Tradeoff** for design items (what does each
+  option cost / gain), OR **Where to find more** for hidden-fact items
+  (file paths, upstream repo, etc.).
+- **Recommendation / rule of thumb**.
+- **Scope tag** — one of **`[Hail]`** (universal Hail behavior),
+  **`[gnomAD data]`** (dataset/schema-specific), or both. Add
+  **`[internal only]`** for anything that applies solely to the internal
+  production team (unreleasable samples, internal bucket paths).
+- **Hail version** — whenever the behavior is version-dependent, state
```
ch-kr, line 149: also add a section to describe known buggy version of Hail versions that aren't described in the changelog but are findable in zulip, the Hail forum, Hail PRs (e.g., requester-pays is broken in  Hail 0.2.138), or comments in the gnomad_qc repo

ch-kr: also add a note which Hail versions are not backwards-compatible and which versions changed table/data formats under the hood

knowledge/README.md
```diff
@@ -0,0 +1,234 @@
...
+  quick-reference, not a textbook
+
+**When to re-look at this doc:** on every Hail version bump, re-verify
+the entries carrying a `Verified on Hail X` line — those are the ones
+whose behavior can silently change under you. Also re-read it when a
+gnomAD release changes schema (v4.1 → v5).
+
+---
+
+## When to consult this
```
ch-kr, line 171: do we need this? are we able to condense this section?

knowledge/README.md
```diff
@@ -0,0 +1,234 @@
...
+  operation-order item under [gnomAD-specific](gnomad-specific.md#gnomad-specific).
+- Anything you **run or debug on Dataproc** — log capture and
+  post-mortems in [Dataproc / operational](dataproc-operational.md#dataproc--operational),
+  especially when a job dies with no Python traceback.
+
+And after any bug that looks like "the total is smaller than I
+expected but there's no error message" — that's almost always a
+`hl.case`/`hl.if_else`/`group_by` null drop.
+
+```{toctree}
```
ch-kr, line 217: what is this section for? this seems redundant with the table above, which has a column "when to consult"

knowledge/README.md
```diff
@@ -0,0 +1,234 @@
...
+- Anything involving **sex chromosomes, PAR, or multiallelics** — see
+  [Biology](biology.md#biology).
+- Anything that **densifies the v4 VDS and classifies genotypes**
+  (het / hom-var counts, freq, phasing) — the high-AB correction +
+  operation-order item under [gnomAD-specific](gnomad-specific.md#gnomad-specific).
+- Anything you **run or debug on Dataproc** — log capture and
+  post-mortems in [Dataproc / operational](dataproc-operational.md#dataproc--operational),
+  especially when a job dies with no Python traceback.
+
+And after any bug that looks like "the total is smaller than I
```
ch-kr, line 213: why is this here and why is only this topic explicitly called out

knowledge/README.md
```diff
@@ -0,0 +1,234 @@
...
+| [Release conventions and data policy](dataset-conventions.md) | you wonder why the data is shaped the way it is — annotation versions, subsets, constraint cutoffs across releases, labels, requester-pays |
+| [Codebase and analysis patterns](analysis-patterns.md) | you pick a starting table, a VEP transcript, a bucket scheme, or a per-stratum aggregation |
+| [Dataproc / operational](dataproc-operational.md) | you run, size, package, or post-mortem a job on Dataproc or QoB, or you're wondering why storage costs what it does |
+| [Testing](testing.md) | after any refactor of the join graph or aggregation semantics, and before you pay for a cluster run |
+| [Working with an AI assistant](working-with-ai.md) | you're setting up `CLAUDE.md` / `CLAUDE.local.md`, or wondering what's burning tokens and cluster time |
+
+
+Entries are tagged **`[Hail]`** (universal Hail behavior) and/or
+**`[gnomAD data]`** (dataset- or schema-specific). Entries tagged
+**`[internal only]`** describe data or artifacts available only to the
```
ch-kr, line 44: we should add a note to the internal section to make sure that code changes to one repo (gnomad_methods or gnomad_qc) do not break code in the other repo

CLAUDE.md
```diff
@@ -0,0 +1,29 @@
+# gnomad_methods
+
+Shared Hail utility library for gnomAD pipelines. **This is a library, not a pipeline** — it exposes APIs that other repos (`gnomad_qc`, `gnomad-constraint`, and others) import, so public-API changes ripple outward. Prefer additive changes over breaking renames.
```
ch-kr, line 3: add comment here that code changes in this repo need to be checked against gnomad_qc (no breaking changes)

.claude/skills/review/README.md
```diff
@@ -0,0 +1,49 @@
+# claude-code-review-skill
```
ch-kr, line 1: @jkgoodrich can you confirm this is the best way to use a skill from another GitHub repo?

.claude/skills/draft-pull-request-description/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+## Code review
+
+<!-- Emit exactly ONE of the two lines below, whichever is true, and delete the other.
+     Never emit both: an unchecked "not run" sitting above a checked "run" reads as
+     if no review happened, and a skimmer will believe the first line.
+       run:     - [x] `/review` run — models: Opus, Sonnet
+       not run: - [ ] `/review` not run
+     If a review was run, follow the line with its findings and their disposition
+     (fixed here / deferred / rejected as a false positive). -->
+- [x] `/review` run — models: <!-- e.g. Opus, Sonnet -->
```
ch-kr, line 94: should also add whether any libraries added in the pull request are compatible with existing dependencies in the repo. if not, suggest fixes

.claude/skills/review/README.md
ch-kr: https://github.com/bw2/claude-code-review-skill

.claude/skills/commit/SKILL.md
ch-kr, line 1: there should be a section somewhere in this skill specifying that all code (new or updated) needs to have tests

ch-kr (commented):
I ran `/doctor` on this branch just to see what claude would think. here are the relevant points (it also flagged the conflict between this branch and the newly merged CLAUDE.md work, so I haven't copied those comments over):

> .claude/settings.json pre-approves several commands that aren't read-only (details inline) — and since it's checked in, it applies to every contributor's sessions.

For this point, it flagged:
```
  .claude/settings.json line 8 — "Bash(git branch:*)"

  ▎ This one isn't read-only: git branch <name> creates a branch, -D deletes one, and -f main <sha> force-moves one. :* rules are prefix string
  ▎ matches with no flag analysis, so all of those are pre-approved. Suggest replacing with the exact read-only forms:
  ▎ "Bash(git branch)",
  ▎ "Bash(git branch -a)",
  ▎ "Bash(git branch --show-current)",
  ▎ "Bash(git branch --contains:*)"

  .claude/settings.json line 13 — "Bash(find:*)"

  ▎ find -exec <cmd> is arbitrary command execution and -delete removes files, so this wildcard is effectively an execute/delete grant, not a
  ▎ read-only one. Suggest dropping it — Claude Code's built-in Glob/Grep tools cover most of what find gets used for without a permission rule.

  .claude/settings.json line 11 — "Bash(rg:*)"

  ▎ ripgrep's --pre <cmd> runs an arbitrary program on every file searched, so rg:* is also an execution grant. Either drop it (the built-in Grep tool
  ▎ uses ripgrep internally and needs no rule) or keep it with that tradeoff stated in the PR description.

  .claude/settings.json lines 5–7 — git log / git diff / git show wildcards

  ▎ Lower severity, but worth knowing: all three accept --output=<file>, which writes an arbitrary file, so the :* wildcards admit a write path. If
  ▎ you want to be strict, exact rules for the common invocations are the safe pattern; if the team is fine with this residual risk, no change needed
  ▎ — just deciding it consciously.
  ▎
  ▎ Related: lines 17–19 (pytest, python -m pytest, ./docs/build.sh) execute repo code by design. That's a defensible choice for a dev repo, but the
  ▎ commit message says "routine read-only commands" — suggest the PR description name the execution grants explicitly so they're approved as such.
```

I don't have strong feelings about these points, e.g., I'm ok with claude being able to create/delete/move branches


> The allowed-tools frontmatter values in three skills don't match real Claude Code tool names and are silently ignored (details inline).

```
  .claude/skills/commit/SKILL.md line 4 (same comment applies to clean-git-history/SKILL.md:4 and draft-pull-request-description/SKILL.md:4)

  ▎ allowed-tools: bash, read_file — Claude Code tool names are case-sensitive title-case, and read_file isn't a Claude Code tool name (these look
  ▎ ported from a different harness). Unrecognized names are silently ignored, so this restriction currently does nothing. Fix:
  ▎ allowed-tools: Bash, Read
```

knowledge/hail-idioms.md
ch-kr, line 1: (this might not be the right file, but wanted to drop this comment somewhere) this came up in Kaitlin's lab meeting today, but maybe we should add that 0.2.128 is the most stable Hail version for densify, and double check that we ask AI to verify when format changes make Hail incompatible with previous versions

PR #844, opened 2026-09-04
Optimizations to sex-ploidy adjustment and reducing groups during aggs

mike-w-wilson:
Two changes that make compute_stats_per_ref_site (per-reference-site AN and coverage over a densified VDS) cheaper to run without changing its output. Both are needed by the gnomAD v5 AoU coverage/AN pipeline.

**Breaking change:** `annotate_and_index_source_mt_for_sex_ploidy` now takes the MatrixTable and karyotype expression and returns the annotated MT plus its column and row flag structs, instead of taking a locus expression and indexing the flags back through a `cols()`/`rows()` self-join. The only downstream caller is gnomad_qc v4 `fix_freq_an.py` (line 217). The old behavior is available unchanged under the new public name `index_sex_ploidy_flags`, so that call can be fixed by renaming it. `adjust_sex_ploidy` and `adjusted_sex_ploidy_expr` keep their signatures.

1. **Sex-ploidy adjustment without a second densify.** `adjusted_sex_ploidy_expr` looks up each sample's karyotype and each locus's PAR status by reaching back into the source MatrixTable's columns and rows, which Hail evaluates as a second full densify on a densified VDS. `annotate_and_index_source_mt_for_sex_ploidy` now annotates those flags directly onto the MT it is given, `adjust_sex_ploidy` is a thin wrapper over it, and `compute_stats_per_ref_site` uses that path. The flags themselves are defined once in `get_sex_ploidy_col_flags_expr` / `get_sex_ploidy_row_flags_expr`, and the genotype rules are shared with `adjusted_sex_ploidy_expr` via `_sex_ploidy_case_expr`. Temporary flag fields use collision-safe names.

2. **Aggregate once per "cell" instead of once per group.** Allele-number aggregation previously ran once for every `freq_meta` group, and the minimal-groups reduction could not help with downsampling groups. The new `reduce_to_cells` option groups samples by their exact pattern of group memberships. Cells don't overlap and every group is exactly a set of cells, so every group's value, downsamplings included, is recovered by summing its cells. It uses the same reduction globals as before. `force_leaf_groups` still works, groups with no samples are kept as real leaves, and a `freq_meta` entry without a `"group"` key or a `force_leaf_groups` target not in `freq_meta` raises up front.

Tests cover both changes, including that cell reconstruction matches both the full aggregation and `find_minimal_strata_groups`. The `/review` skill was run with Opus and Sonnet 5; both of its findings were accepted and addressed (the breaking change above is documented, and `index_sex_ploidy_flags` was made public).

### DIFF
--- a/gnomad/sample_qc/sex.py
+++ b/gnomad/sample_qc/sex.py
@@ -17,6 +17,40 @@ logger.setLevel(logging.INFO)
 SEXES = {"XY": "XY", "XX": "XX"}
 
 
+def _sex_ploidy_case_expr(
+    gt_expr: hl.expr.CallExpression,
+    col_flags: hl.expr.StructExpression,
+    row_flags: hl.expr.StructExpression,
+) -> hl.expr.CallExpression:
+    """
+    Build the sex-ploidy genotype adjustment from precomputed sample and locus flags.
+
+    :param gt_expr: Genotype expression.
+    :param col_flags: Struct with per-sample booleans ``xy`` and ``xx``.
+    :param row_flags: Struct with per-locus booleans ``in_autosome``,
+        ``in_non_par``, ``x_nonpar``, ``y_par`` and ``y_nonpar``.
+    :return: Genotype adjusted for sex ploidy.
+    """
+    return (
+        hl.case(missing_false=True)
+        # Added to reduce the checks by entry.
+        .when(row_flags.in_autosome, gt_expr)
+        .when(
+            (row_flags.y_par | row_flags.y_nonpar) & col_flags.xx, hl.missing(hl.tcall)
+        )
+        .when(~row_flags.in_non_par, gt_expr)
+        .when(
+            (row_flags.x_nonpar | row_flags.y_nonpar) & col_flags.xy & gt_expr.is_het(),
+            hl.missing(hl.tcall),
+        )
+        .when(
+            (row_flags.x_nonpar | row_flags.y_nonpar) & col_flags.xy,
+            hl.call(gt_expr[0], phased=False),
+        )
+        .default(gt_expr)
+    )
+
+
 def adjusted_sex_ploidy_expr(
     locus_expr: hl.expr.LocusExpression,
     gt_expr: hl.expr.CallExpression,
@@ -26,6 +60,15 @@ def adjusted_sex_ploidy_expr(
 ) -> hl.expr.CallExpression:
     """
     Create an entry expression to convert XY to haploid on non-PAR X/Y and XX to missing on Y.
+
+    .. note::
+
+        The per-sample and per-locus flags are indexed back from the source
+        MatrixTable's ``cols()`` and ``rows()`` (see
+        `annotate_and_index_source_mt_for_sex_ploidy`). Hail evaluates those
+        self-joins as a second pass over the source MatrixTable's upstream
+        pipeline, so when the source is expensive to compute (e.g. a densified
+        MatrixTable) use `adjust_sex_ploidy`, which annotates the flags in place.
 
     :param locus_expr: Locus expression.
     :param gt_expr: Genotype expression.
@@ -40,22 +83,7 @@ def adjusted_sex_ploidy_expr(
         locus_expr, karyotype_expr, xy_karyotype_str, xx_karyotype_str
     )
 
-    return (
-        hl.case(missing_false=True)
-        # Added to reduce the checks by entry.
-        .when(row_idx.in_autosome, gt_expr)
-        .when((row_idx.y_par | row_idx.y_nonpar) & col_idx.xx, hl.missing(hl.tcall))
-        .when(~row_idx.in_non_par, gt_expr)
-        .when(
-            (row_idx.x_nonpar | row_idx.y_nonpar) & col_idx.xy & gt_expr.is_het(),
-            hl.missing(hl.tcall),
-        )
-        .when(
-            (row_idx.x_nonpar | row_idx.y_nonpar) & col_idx.xy,
-            hl.call(gt_expr[0], phased=False),
-        )
-        .default(gt_expr)
-    )
+    return _sex_ploidy_case_expr(gt_expr, col_idx, row_idx)
 
 
 def adjust_sex_ploidy(
@@ -63,19 +91,46 @@ def adjust_sex_ploidy(
     sex_expr: hl.expr.StringExpression,
     xy_str: str = "XY",
     xx_str: str = "XX",
+    gt_field: str = "GT",
 ) -> hl.MatrixTable:
     """
     Convert XY to haploid on non-PAR X/Y, sets XX to missing on Y.
+
+    The per-sample (``xy``/``xx``) and per-locus (PAR/non-PAR) flags are
+    annotated directly onto `mt`'s columns and rows and dropped again after the
+    genotype adjustment, so `mt`'s upstream pipeline is evaluated once. This is
+    the same adjustment as `adjusted_sex_ploidy_expr` without that function's
+    ``cols()``/``rows()`` self-joins.
 
     :param mt: Input MatrixTable
     :param sex_expr: Expression pointing to sex in MT (if not xy_str or xx_str, no change)
     :param xy_str: String for XY (default 'XY')
     :param xx_str: String for XX (default 'XX')
+    :param gt_field: Name of the genotype entry field to adjust. Default is "GT".
     :return: MatrixTable with fixed ploidy for sex chromosomes
     """
-    return mt.annotate_entries(
-        GT=adjusted_sex_ploidy_expr(mt.locus, mt.GT, sex_expr, xy_str, xx_str)
+    mt = mt.annotate_cols(
+        _sex_ploidy_col=hl.struct(
+            xy=sex_expr.upper() == xy_str, xx=sex_expr.upper() == xx_str
+        )
     )
+    mt = mt.annotate_rows(
+        _sex_ploidy_row=hl.struct(
+            in_non_par=~mt.locus.in_autosome_or_par(),
+            in_autosome=mt.locus.in_autosome(),
+            x_nonpar=mt.locus.in_x_nonpar(),
+            y_par=mt.locus.in_y_par(),
+            y_nonpar=mt.locus.in_y_nonpar(),
+        )
+    )
+    mt = mt.annotate_entries(
+        **{
+            gt_field: _sex_ploidy_case_expr(
+                mt[gt_field], mt._sex_ploidy_col, mt._sex_ploidy_row
+            )
+        }
+    )
+    return mt.drop("_sex_ploidy_col", "_sex_ploidy_row")
 
 
 def gaussian_mixture_model_karyotype_assignment(
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -2413,7 +2413,9 @@ def find_strata_cells(
     `expand_strata_array_from_leaves` and `agg_by_strata`, with their metadata
     (`{"group": <label>, "cell": "<k>"}`) carried only in the reduced
     `freq_meta`. Groups listed in `force_leaf_groups` are kept as real leaves
-    (their own `freq_meta` index) ahead of the cells.
+    (their own `freq_meta` index) ahead of the cells. A group with no samples
+    has no cell to be built from, so it is also kept as a real leaf (after its
+    label's cells) rather than left without a decomposition.
 
     Samples that belong to no group of a label (an all-False pattern) do not
     form a cell; no group needs them.
@@ -2479,6 +2481,7 @@ def find_strata_cells(
             leaf_indices.append(cell_leaf[p])
             leaf_meta.append({"group": label, "cell": str(k)})
             leaf_sample_count.append(counts_by_label[label][p])
+        zero_sample = []
         for pos, i in enumerate(idxs):
             if i in forced:
                 continue
@@ -2489,6 +2492,9 @@ def find_strata_cells(
                     f"Cells of group {freq_meta[i]} sum to {n_children} samples,"
                     f" expected {freq_meta_sample_count[i]}."
                 )
+            if not children:
+                zero_sample.append(i)
+                continue
             decomposition[i] = children
         if cells:
             cell_of_sample = hl.literal({p: k for k, p in enumerate(cells)}).get(
@@ -2498,6 +2504,14 @@ def find_strata_cells(
                 hl.coalesce(cell_of_sample == k, False) for k in range(len(cells))
             )
         next_leaf += len(cells)
+        # No cell covers a group with no samples, so it would otherwise be
+        # neither a leaf nor decomposable. Keep it as a real leaf with its own
+        # (all-False) membership; aggregating it directly costs nothing.
+        for i in zero_sample:
+            leaf_indices.append(i)
+            leaf_meta.append(dict(freq_meta[i]))
+            leaf_sample_count.append(0)
+            membership_exprs.append(ht.group_membership[i])
 
     logger.info(
         "Reducing freq_meta from %d groups to %d cells (%s) plus %d forced leaves.",
@@ -2706,10 +2720,11 @@ def generate_freq_group_membership_array(
         across their values when `reduce_to_minimal_groups` is True.
         Default is None, which resolves to `{"downsampling"}`.
     :param force_leaf_groups: Optional list of `freq_meta` dicts to keep
-        as leaves under `reduce_to_minimal_groups` (see
-        `find_minimal_strata_groups`). Use this for groups that downstream
-        code accesses directly, so they remain explicitly computed instead
-        of being reconstructed per row from child groups.
+        as leaves under either reduction mode (see
+        `find_minimal_strata_groups` and `find_strata_cells`). Use this for
+        groups that downstream code accesses directly, so they remain
+        explicitly computed instead of being reconstructed per row from
+        child groups or cells.
     :param group_label: Value to use for the `"group"` key on every
         constructed `freq_meta` entry. Default is `"adj"`. Set to `"raw"`
         for callers that don't apply any genotype-level filtering (e.g.,
@@ -2721,7 +2736,7 @@ def generate_freq_group_membership_array(
         reconstructed by summing its cells, so each sample is aggregated once
         per label. The same reduction-tracking globals are written. Mutually
         exclusive with `reduce_to_minimal_groups`; `non_summable_strata` is
-        unused. Default is False.
+        unused, but `force_leaf_groups` is still applied. Default is False.
     :return: Table with the 'group_membership' array annotation.
     """
     if reduce_to_minimal_groups and reduce_to_cells:
--- a/gnomad/utils/annotations.py
+++ b/gnomad/utils/annotations.py
@@ -2454,10 +2454,18 @@ def find_strata_cells(
         idx_by_label[label].append(i)
 
     # One pass over the sample Table: per label, count samples per membership
-    # pattern, encoded as a "0"/"1" string so it can key a literal dict.
+    # pattern, encoded as a "0"/"1" string so it can key a literal dict. A
+    # membership bit is missing when a sample lacks a stratification value;
+    # the group aggregations (`hl.agg.filter`) and `freq_meta_sample_count`
+    # (`hl.agg.count_where`) both treat that as not a member, so the pattern
+    # must too (`hl.delimit` would otherwise render it as "null" and shift
+    # every later position).
     def _pattern_expr(label: str) -> hl.expr.StringExpression:
         return hl.delimit(
-            [hl.if_else(ht.group_membership[i], "1", "0") for i in idx_by_label[label]],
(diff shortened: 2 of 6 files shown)


gnomad/sample_qc/sex.py
```diff
@@ -17,6 +17,40 @@
...
+
+    :param gt_expr: Genotype expression.
+    :param col_flags: Struct with per-sample booleans ``xy`` and ``xx``.
+    :param row_flags: Struct with per-locus booleans ``in_autosome``,
+        ``in_non_par``, ``x_nonpar``, ``y_par`` and ``y_nonpar``.
+    :return: Genotype adjusted for sex ploidy.
+    """
+    return (
+        hl.case(missing_false=True)
+        # Added to reduce the checks by entry.
```
ch-kr, line 36: I didn't realize this comment was copied and pasted. I'm not sure I understand what it means; maybe we should just remove it?

gnomad/sample_qc/sex.py
```diff
@@ -17,6 +17,40 @@
...
+        .when(row_flags.in_autosome, gt_expr)
```
ch-kr, line 37: I know this was copied and pasted from below, but maybe this flag should be `in_autosome_or_par`, which would remove the need for `.when(~row_flags.in_non_par, gt_expr)` below

ch-kr: I changed my mind since this would also mean updating https://github.com/broadinstitute/gnomad_methods/blob/af5ccc115cf9b1a4e431be28020801d1b7306cb4/gnomad/utils/annotations.py#L697 for a pretty trivial change, but I left these as comments in case this comes up again

gnomad/sample_qc/sex.py
```diff
@@ -27,6 +61,15 @@ def adjusted_sex_ploidy_expr(
     """
     Create an entry expression to convert XY to haploid on non-PAR X/Y and XX to missing on Y.
 
+    .. note::
+
+        The per-sample and per-locus flags are indexed back from the source
+        MatrixTable's ``cols()`` and ``rows()`` (see
+        `annotate_and_index_source_mt_for_sex_ploidy`). Hail evaluates those
+        self-joins as a second pass over the source MatrixTable's upstream
+        pipeline, so when the source is expensive to compute (e.g. a densified
```
ch-kr, line 70: maybe it would be better to update `annotate_and_index_source_mt_for_sex_ploidy` directly rather than creating a second option to adjust sex ploidy? it is a little against our usual conventions for this repo, but that function could take the source MT as input (rather than deriving the source using Hail internal code) and return column/row expressions so that these extra joins are never computed, regardless of dataset size. this would remove redundant looking code in `adjust_sex_ploidy` but would be a breaking change

gnomad/utils/annotations.py
ch-kr, line 2581:
```suggestion
    This is the post-processing companion to `find_minimal_strata_groups` or `find_strata_cells`.
```

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2381,155 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
...
+            raise ValueError(
+                f"`force_leaf_groups` entry {target} is not in `freq_meta`."
+            )
+        forced.extend(matches)
+
+    # Group positions by label, preserving first-seen label order.
+    labels: List[str] = []
+    idx_by_label: Dict[str, List[int]] = {}
+    for i, m in enumerate(freq_meta):
+        label = m.get("group")
```
ch-kr, line 2456: group will always be defined for gnomad, but gemini flagged this:
> The Issue:
If any dictionary in freq_meta is missing the "group" key, m.get("group") will return None.
Later, when you attempt to unpack the dictionary into hl.struct(**{...}), Python will encounter {None: hl.agg.counter(...)}. In Python, unpacking a dictionary as keyword arguments strictly requires the keys to be strings. This will instantly throw a TypeError: keywords must be strings.
The Fix:
Ensure a default string fallback or filter out missing groups.

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2381,155 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
...
+                zero_sample.append(i)
+                continue
+            decomposition[i] = children
+        if cells:
+            cell_of_sample = hl.literal({p: k for k, p in enumerate(cells)}).get(
+                _pattern_expr(label)
+            )
+            membership_exprs.extend(
+                hl.coalesce(cell_of_sample == k, False) for k in range(len(cells))
+            )
```
ch-kr: gemini flagged potential IR bloat here:
> The Issue:
This causes massive Expression Tree (IR) Bloat. cell_of_sample is a Python variable holding a complex Hail expression tree (it contains the hl.literal dict, a .get() call, and the entire _pattern_expr which contains a delimited loop of hl.if_else statements).

> Because you are iterating over a Python generator (for k in range(len(cells))), Hail's Python frontend does not cache cell_of_sample. Instead, it literally copies and pastes the entire massive AST tree into the graph for every single cell. If you have 500 cells, the _pattern_expr logic is inserted into the execution graph 500 times. This will cause Spark/JVM compilation times to grind to a halt before the job even starts.

> The Fix:
Shift the iteration into the JVM by using hl.bind or building the array natively in Hail rather than using a Python list comprehension.

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2381,155 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
...
+        for k, p in enumerate(cells):
+            leaf_indices.append(cell_leaf[p])
+            leaf_meta.append({"group": label, "cell": str(k)})
+            leaf_sample_count.append(counts_by_label[label][p])
+        zero_sample = []
+        for pos, i in enumerate(idxs):
+            if i in forced:
+                continue
+            children = [cell_leaf[p] for p in cells if p[pos] == "1"]
+            n_children = sum(counts_by_label[label][p] for p in cells if p[pos] == "1")
```
ch-kr, line 2509: suggestion from gemini:
> The Issue:
You are iterating over cells and checking p[pos] == "1" twice sequentially.
The Fix:
Combine them into a single pass to save a few CPU cycles during this Python-side evaluation.

```
children = []
n_children = 0
for p in cells:
    if p[pos] == "1":
        children.append(cell_leaf[p])
        n_children += counts_by_label[label][p]
```

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2381,155 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
...
+        if label not in idx_by_label:
+            labels.append(label)
+            idx_by_label[label] = []
+        idx_by_label[label].append(i)
+
+    # One pass over the sample Table: per label, count samples per membership
+    # pattern, encoded as a "0"/"1" string so it can key a literal dict.
+    def _pattern_expr(label: str) -> hl.expr.StringExpression:
+        return hl.delimit(
+            [hl.if_else(ht.group_membership[i], "1", "0") for i in idx_by_label[label]],
```
ch-kr: gemini also flagged some inefficiency here:
> The Error:
```
def _pattern_expr(label: str) -> hl.expr.StringExpression:
    return hl.delimit(
        [hl.if_else(ht.group_membership[i], "1", "0") for i in idx_by_label[label]],
        "",
    )
```

> Similarly, this generates a distinct hl.if_else node for every single index in a label. For labels with hundreds of groups (like downsamplings), this creates an excessively wide AST.

> The Fix:
Pass the list of indices as a single hl.literal array, and use .map() to evaluate the conditions inside the JVM context:
```
def _pattern_expr(label: str) -> hl.expr.StringExpression:
    indices_array = hl.literal(idx_by_label[label])
    return hl.delimit(
        indices_array.map(lambda i: hl.if_else(ht.group_membership[i], "1", "0")), 
        ""
    )
```

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2381,155 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
...
+        if not matches:
+            raise ValueError(
+                f"`force_leaf_groups` entry {target} is not in `freq_meta`."
+            )
+        forced.extend(matches)
+
+    # Group positions by label, preserving first-seen label order.
+    labels: List[str] = []
+    idx_by_label: Dict[str, List[int]] = {}
+    for i, m in enumerate(freq_meta):
```
ch-kr: gemini also flagged that any groups in `force_leaf_groups` are being included in `idx_by_label`, which means that this function will generate more cells than is necessary when forced groups exist:
> Groups listed in force_leaf_groups are skipped during cell decomposition but are still included in idx_by_label and therefore in _pattern_expr(label).
Because forced groups are part of the 0/1 string pattern, their variations artificially split the samples into separate cells for the remaining groups. This inflates the number of generated cells, defeating the algorithm's purpose.
Fix: Filter out forced indices before appending to idx_by_label, which naturally removes them from the _pattern_expr strings and the pos indexing logic entirely.

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2381,155 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
...
+    n_full = len(freq_meta)
+    force_leaf_groups = force_leaf_groups or []
+    forced = []
+    for target in force_leaf_groups:
+        matches = [i for i, m in enumerate(freq_meta) if m == target]
+        if not matches:
+            raise ValueError(
+                f"`force_leaf_groups` entry {target} is not in `freq_meta`."
+            )
+        forced.extend(matches)
```
ch-kr, line 2444: gemini:
> If force_leaf_groups contains duplicate target metadata dictionaries, forced.extend(matches) will append the same indices multiple times. This leads to duplicate leaf_indices and redundant checks.
Fix: Use a set or deduplicate forced before proceeding: forced = list(set(forced_raw)).

tests/sample_qc/test_sex.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def test_adjusts_expected_entries(self, mt):
+        """Spot-check every branch of the adjustment on known entries."""
+        actual = self._entries(adjust_sex_ploidy(mt, mt.sex_karyotype), "GT")
+        by_key = {(c, p, s): gt for c, p, s, gt in actual}
+        # Genotype of sample s_idx at locus l_idx is gts[(l_idx + s_idx) % 4].
+        # Autosome and PAR genotypes are untouched (XY, PAR1 and PAR2).
+        assert by_key[("chr1", 1000, "s2")] == hl.Call([0, 1])
+        assert by_key[("chrX", 20000, "s2")] == hl.Call([1, 1])
+        assert by_key[("chrX", 155800000, "s3")] == hl.Call([0, 0])
```
ch-kr, line 82: maybe one of these genotypes should be het, just to make this test more thorough, since s2 and s3 are both XY

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2381,194 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
...
+        leaf_sample_count, leaf_membership_expr)`: the leaf encoding as for
+        `find_minimal_strata_groups` (cell leaves are indices >=
+        `len(freq_meta)`), the reduced `freq_meta` and sample counts, and the
+        per-sample leaf-only `group_membership` expression on `ht`.
+    """
+    n_full = len(freq_meta)
+    force_leaf_groups = force_leaf_groups or []
+    forced = []
+    for target in force_leaf_groups:
+        matches = [i for i, m in enumerate(freq_meta) if m == target]
```
ch-kr, line 2439: minor, but this check could be combined with the for loop enumerating `freq_meta` below. if you find `i` in `force_leaf_groups`, add it to `forced`, and then check the difference between force_leaf_groups and forced at the end of the loop. this would do a little extra work, since you'd build `labels` and `idx_by_label` but hit an error if any entries in `force_leaf_groups` aren't in `freq_meta`, but it would clean this up slightly

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2381,194 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
...
+    # must too (`hl.delimit` would otherwise render it as "null" and shift
+    # every later position).
+    def _pattern_expr(label: str) -> hl.expr.StringExpression:
+        return hl.delimit(
+            [
+                hl.if_else(hl.coalesce(ht.group_membership[i], False), "1", "0")
+                for i in idx_by_label[label]
+            ],
+            "",
+        )
```
ch-kr, line 2476: from gemini:
> In your _pattern_expr function, you are using a Python list comprehension ([... for i in ...]) instead of Hail's native .map(). You must convert the Python list of indices into a Hail array literal and map over it natively. This shifts the loop execution to the JVM:
```
# FIXED (Executes natively in Hail Engine)
def _pattern_expr(label: str) -> hl.expr.StringExpression:
    indices_array = hl.literal(idx_by_label[label])
    return hl.delimit(
        indices_array.map(
            lambda i: hl.if_else(hl.coalesce(ht.group_membership[i], False), "1", "0")
        ),
        "",
    )```

File: tests/utils/test_annotations.py [hunk 13]
```diff
@@ -2752,6 +2753,99 @@ def test_misaligned_sample_count_raises(self):
             find_minimal_strata_groups([{"group": "adj"}], [1, 2])
 
 
+class TestFindStrataCells:
+    """Test the find_strata_cells function."""
```
ch-kr, line 2757: none of the tests below seem to test running `find_strata_cells` with `force_leaf_groups` set; this should probably be covered.

I also don't know if I missed this, but should there be a test ensuring that the output of `find_strata_cells` and `find_minimal_strata_groups` are equivalent?

ch-kr (changes requested):
thank you for adding this! a few efficiency notes

tests/sample_qc/test_sex.py
mike-w-wilson: Yes good point

gnomad/sample_qc/sex.py
mike-w-wilson: Good call, I took your advice and shortened things up a bit with helpers too. Like you said its a breaking change for the annotate_and_index with the change in params but currently (internally) only fix_freq_an in gnomad_qc/v4 uses this. We can mark this as a breaking change anyways and its an easy caller update for anyone who actually uses it. Worth the optimization to me.

mike-w-wilson:
back to you @ch-kr ! I took your suggestion around updating the annotate_and _index function instead to remove bloat and rewrote the PR description to include the breaking change.

mike-w-wilson:
Ive added https://github.com/broadinstitute/gnomad_qc/pull/787 which fixes the QC check. I originally was not going to change it but figured its simple change is easier than getting around the action.

gnomad/utils/annotations.py
```diff
@@ -694,53 +694,142 @@ def create_frequency_bins_expr(
     return bin_expr
 
 
-def annotate_and_index_source_mt_for_sex_ploidy(
-    locus_expr: hl.expr.LocusExpression,
+def get_sex_ploidy_col_flags_expr(
```
ch-kr: should this function (and the row flags) have tests?

gnomad/utils/annotations.py
```diff
@@ -2381,6 +2468,205 @@ def non_summable_strata_of(entry: Dict[str, str]) -> frozenset:
     return leaf_indices, decomposition
 
 
+def find_strata_cells(
```
ch-kr, line 2473: from claude, though it noted this doesn't need to be fixed as part of this PR since it existed before 844


**Issue.** In `agg_by_strata`, an `entry_agg_group_membership` target that resolves to a parent (not a leaf) builds its sample index list inline in the per-row expression (`gnomad/utils/annotations.py:3458`), by filtering all samples against every child position. Leaf targets read the precomputed `indices_by_group` global instead (`annotations.py:3388`). Hail does not hoist global-only sub-expressions out of a row map, so the parent list is rebuilt at every reference site at a cost of `n_samples × n_children`. Quick check (global-only filter over a 100k array, inline in `annotate` vs precomputed with `annotate_globals`):

| rows   | inline in row expr | precomputed global |
|--------|--------------------|--------------------|
| 2,000  | 0.6 s              | 0.2 s              |
| 40,000 | 2.5 s              | 0.5 s              |

This path predates the PR, but `reduce_to_cells` makes it heavier: under cells `{"group": "adj"}` is never a leaf, so it decomposes into every adj cell. Today the only parent pin is gnomAD `coverage_stats -> {"group": "adj"}` with ~800 samples and ~20 cells, so the current cost is negligible. On the AoU run (~400k samples, hundreds of downsampling cells) any future pin to `adj` would add `400k × n_cells` work per site and give back a large share of the cell savings. Forcing `{"group": "adj"}` as a leaf avoids the scan but adds an extra all-samples AN aggregation per site, so it is the wrong fix for AoU.

**Suggestion.** Compute the per-target parent index lists once as globals and have the row expression read them. Output is unchanged; also speeds up the existing minimal-groups parent path.

```python
ht = ht.annotate_globals(**global_expr)
if entry_agg_group_membership:
    ht = ht.annotate_globals(
        **{
            f"_{ann}_targets": hl.struct(
                **dict(zip(("s_indices", "adj"), _per_target_indices_and_adj(targets)))
            )
            for ann, targets in entry_agg_group_membership.items()
        }
    )

def _agg_for(ann, f):
    if ann in entry_agg_group_membership:
        t = ht[f"_{ann}_targets"]
        return _agg_by_group(t.s_indices, t.adj, agg_func=f[1], ann_expr=ht[ann])
    ...
```

(`_per_target_indices_and_adj` needs to be defined before the `annotate_globals`, and the `_*_targets` globals dropped with `cols` at the end.)

tests/utils/test_annotations.py
```diff
@@ -2752,6 +2754,230 @@ def test_misaligned_sample_count_raises(self):
...
+        out, col_flags, row_flags = annotate_and_index_source_mt_for_sex_ploidy(
+            mt, mt.sex_karyotype
+        )
+        assert out.aggregate_cols(hl.agg.all(out._sex_ploidy_col == 1))
+        assert out.aggregate_rows(hl.agg.all(out._sex_ploidy_row == 2))
+        assert "xy" in col_flags and "in_non_par" in row_flags
+
+
+class TestFindStrataCells:
+    """Test the find_strata_cells function."""
```
ch-kr, line 2918: I asked claude if there were any missing tests that were important:

> A pinned parent with many cells. The production gnomAD path pins coverage_stats to {"group": "adj"}, which under cells is a parent of every adj cell. The only cells-mode pin tested is max_called -> {"group": "raw"} in [test_sparse_mt.py:966](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/Users/kchao/code/gnomad_methods/tests/utils/test_sparse_mt.py#L966), and the raw label collapses to one cell, so the multi-child parent path in agg_by_strata is never exercised under cells. Add a non-summable agg pinned to {"group": "adj"} or {"group": "adj", "gen_anc": "afr"} and compare to the full run. The hl.agg.hist parent test at [test_sparse_mt.py:703](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/Users/kchao/code/gnomad_methods/tests/utils/test_sparse_mt.py#L703) covers this under minimal groups only, and the IR-shape bug it guards is exactly the kind that could differ with hundreds of children.

gnomad/utils/annotations.py
```diff
@@ -694,53 +694,142 @@ def create_frequency_bins_expr(
...
+        **get_sex_ploidy_col_flags_expr(
+            karyotype_expr, xy_karyotype_str, xx_karyotype_str
+        )
+    ).cols()
+    row_ht = source_mt.annotate_rows(**get_sex_ploidy_row_flags_expr(locus_expr)).rows()
+
+    return col_ht[source_mt.col_key], row_ht[source_mt.row_key]
 
 
 def get_is_haploid_expr(
```
ch-kr, line 837: since this function was updated, should we add a test?

ch-kr (commented):
two comments about test, and one separate efficiency issue that claude flagged

gnomad/utils/annotations.py
mike-w-wilson: Added since this is already about optimizations.

tests/utils/test_annotations.py
mike-w-wilson: Added however "many" is only 8 here. We can increase it but I'm not sure its needed.

gnomad/utils/annotations.py
mike-w-wilson: Added!

gnomad/utils/annotations.py
ch-kr, line 3238:
```suggestion
       in 'group_membership'. If the groups were reduced (`freq_reduced=True`), a
        target can also be any group in `freq_meta_full`; it is aggregated over all
        samples in the leaves or cells that make it up. If not
```

ch-kr (approved):
one documentation nit but LGTM

